PR #1, opened 2018-11-18
minor changes to Vector class constructor

justbuchanan:
* Allow constructing with a list: `Vector([1, 2, 3])`
* change `ValueError` -> `TypeError`
* throw a `TypeError` if the argument isn't one of the recognized types. This helps catch mistakes earlier rather than later when the vector is used.

### DIFF
--- a/cadquery/occ_impl/geom.py
+++ b/cadquery/occ_impl/geom.py
@@ -28,7 +28,7 @@ class Vector(object):
         elif len(args) == 1:
             if isinstance(args[0], Vector):
                 fV = gp_Vec(args[0].wrapped.XYZ())
-            elif isinstance(args[0], tuple):
+            elif isinstance(args[0], tuple) or isinstance(args[0], list):
                 fV = gp_Vec(*args[0])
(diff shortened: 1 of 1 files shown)


cadquery/occ_impl/geom.py
```diff
@@ -28,7 +28,7 @@ def __init__(self, *args):
         elif len(args) == 1:
             if isinstance(args[0], Vector):
                 fV = gp_Vec(args[0].wrapped.XYZ())
-            elif isinstance(args[0], tuple):
+            elif isinstance(args[0], tuple) or isinstance(args[0], list):
```
42B:
```python
elif isinstance(args[0], (list, tuple)):  # ?
```

justbuchanan: Awesome, thanks!

42B: @justbuchanan 

No problem!

adam-urbanczyk:
Thanks @justbuchanan !

PR #2, opened 2018-11-18
fix Vector.__eq__()

justbuchanan:
I think the previous implementation was just doing a pointer comparison so two unique vectors with the same contents would not be considered equal.

### DIFF
--- a/cadquery/occ_impl/geom.py
+++ b/cadquery/occ_impl/geom.py
@@ -133,7 +133,7 @@ class Vector(object):
         return 'Vector: ' + str((self.x, self.y, self.z))
 
     def __eq__(self, other):
-        return self.x == other.x and self.y == other.y and self.z == other.z
+        return self.wrapped.IsEqual(other.wrapped, 0.00001, 0.00001)
     '''
     is not implemented in OCC
     def __ne__(self, other):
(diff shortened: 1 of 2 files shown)


adam-urbanczyk:
@justbuchanan  good catch. Did you consider using the IsEqual method of the underlying OCC object (gp_Vec)?

@Peque indeed. This is a fork of my repo, that is why. Anyhow, I added it to travis and appveyor.

justbuchanan:
@adam-urbanczyk thanks, I thought I checked for a builtin function, but somehow I missed `IsEquals()`. I switched to that and added a test case. The `IsEquals()` method has two required arguments for linear tolerance and angular tolerance. I put in arbitrary small values for those, but open to better suggestions.

adam-urbanczyk:
@justbuchanan  looks good to me. Maybe change the tolerance from 1e-5 to 1e-6? 

@dcowden  @jmwright  any thoughts? If not I will merge.

dcowden:
this seems fine to me.  i think hard-coding the tolerance is probably a mistake,though.  The implications of tolerance and equals run very deep, especially when you get into things like geometry fixing and stuff like that.

I think the ideal would be to test for equality, but then have a way to reference a single defined geometric tolerance, which can be changed one place. But i'm not sure the best way to accomplish that

adam-urbanczyk:
OK merging and will add an issue about the tolerance handling.

@justbuchanan  thanks for your contribution!

PR #4, opened 2018-11-19
Add test case that loads an invalid STEP file

justbuchanan:
I borrowed this test from cqparts. For some reason, attempting to load an invalid STEP file causes a segfault. I'm not sure yet if this is a bug in cadquery or in pythonOCC.

I'll do some more digging to try and find the root cause, but for now here's a test case to reproduce the issue.

### DIFF
--- a/cadquery/occ_impl/importers.py
+++ b/cadquery/occ_impl/importers.py
@@ -8,6 +8,7 @@ import urllib as urlreader
 import tempfile
 
 from OCC.STEPControl import STEPControl_Reader
+import OCC.IFSelect
 
 
 class ImportTypes:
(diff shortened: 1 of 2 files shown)


justbuchanan:
Looks like this is a problem in pythonOCC. Probably the same as this issue: https://github.com/tpaviot/pythonocc-core/issues/552.

justbuchanan:
I changed the import code to check the return value from `STEPControl_Reader.ReadFile()`, which fixes the issue. See the linked tpaviot/pythonocc-core#552 for discussion.

adam-urbanczyk:
Looks good! Thanks @justbuchanan .

PR #6, opened 2018-11-19
implement Vector operators __add__, __sub__, __mul__

justbuchanan:


adam-urbanczyk:
Looks good. One question though: do you have a specific use case for the Vector object in mind?
 
Also it would be nice to add some tests.

justbuchanan:
cqparts uses `Vector`s fairly extensively for moving between coordinate systems, among other things. Since several of the cadquery APIs accept and return `Vector` objects, it works out well to use them throughout. It would certainly be possible to use methods like `multiply` and `add` directly, but the operators are really convenient.

I added some test cases and implemented `__truediv__` to complete the basic four operators.

adam-urbanczyk:
Thanks for the contribution @justbuchanan

PR #14, opened 2018-11-26
Pass exportShape()'s 'tolerance' param to exportStl()

justbuchanan:
For reference, see the freecad_impl version: 

https://github.com/dcowden/cadquery/blob/1c86a979a08fa201283d89ad2848650dbee1009d/cadquery/freecad_impl/exporters.py#L87

### DIFF
--- a/cadquery/occ_impl/exporters.py
+++ b/cadquery/occ_impl/exporters.py
@@ -105,7 +105,7 @@ def exportShape(shape, exportType, fileLike, tolerance=0.1):
         if exportType == ExportTypes.STEP:
             shape.exportStep(outFileName)
         elif exportType == ExportTypes.STL:
-            shape.exportStl(outFileName)
+            shape.exportStl(outFileName, tolerance)
         else:
             raise ValueError("No idea how i got here")
 


adam-urbanczyk:
Looks good. I have one question - is the default tolerance of 0.1 good enough? Did you apply it in practice?

jmwright:
During the design process I usually care more about performance so that I can iterate faster. When I go to export things for printing, CNC, whatever I might then mess with the precision. However, there seems to be a minimum precision to keep models from doing weird things (at least with FreeCAD), and we want to definitely be above that level. So, I guess I would lean towards lower precision and better performance.

jmwright:
@justbuchanan No sorry, I can't remember any specific ones. I remember at some point there was an example, and maybe a test or two that were prone to tolerance issues. The example might have been either the [remote enclosure](https://github.com/dcowden/cadquery/blob/master/examples/FreeCAD/Ex027_Remote_Enclosure.py) or the [panel](https://github.com/dcowden/cadquery/blob/master/examples/FreeCAD/Ex030_Panel_with_Various_Holes_for_Connector_Installation.py), but I can't be certain. There may have been a selector bug mixed in there too.

adam-urbanczyk:
OK, I'll merge as-is then. Everyone agrees?

jmwright:
I think it's fine to merge.

PR #15, opened 2018-11-30
import all shapes from a STEP file, not just the first one

justbuchanan:
See the docs on `TransferRoot()` here: https://www.opencascade.com/doc/occt-6.9.1/refman/html/class_s_t_e_p_control___reader.html#a61c12e9a909a22f7bf52c478371001e1.

Not quite ready to merge yet, I'll add a test first.


justbuchanan:
I added a test case with an example STEP file borrowed from cqparts.

adam-urbanczyk:
@justbuchanan good catch! Is this ready for merging? Or do we need more tests?

adam-urbanczyk:
OK, merging. Thanks @justbuchanan

PR #16, opened 2018-12-01
implement Matrix.__getitem__() and Matrix(list) constructor

justbuchanan:
* Implement `Matrix.__getitem__()`, so individual values can be accessed with `matrix[row, column]` syntax. This is useful generally and is required for some cqparts testing of coordinate transforms.
* Matrix constructor accepts a list of values as input. The input can be all 16 values of the matrix or just the first 12. Since this is a transform matrix, the last row should always be `[0, 0, 0, 1]`.
* Matrix constructor throws a `TypeError` if an invalid input is given
* Test cases are added to test all of these things.

A couple notes:
* gp_Trsf value access is 1-indexed, but I used 0-indexing for the accessors to be consistent with everything python
* An alternative to accepting a flat list of values in the `Matrix` constructor would be to do what numpy does and accept a list of lists. Something like `[[a, b], [c, d]]` instead of `[a, b, c, d]`. I don't really have a preference, let me know your thoughts.

### DIFF
--- a/cadquery/occ_impl/geom.py
+++ b/cadquery/occ_impl/geom.py
@@ -161,6 +161,18 @@ class Matrix:
     """A 3d , 4x4 transformation matrix.
 
     Used to move geometry in space.
+
+    The provided "matrix" parameter may be None, a gp_Trsf, or a nested list of
+    values.
+
+    If given a nested list, it is expected to be of the form:
+
+        [[m11, m12, m13, m14],
+         [m21, m22, m23, m24],
+         [m31, m32, m33, m34]]
+
+    A fourth row may be given, but it is expected to be: [0.0, 0.0, 0.0, 1.0]
+    since this is a transform matrix.
     """
 
     def __init__(self, matrix=None):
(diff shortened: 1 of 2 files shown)


cadquery/occ_impl/geom.py
```diff
@@ -167,8 +167,23 @@ def __init__(self, matrix=None):
...
             self.wrapped = matrix
+        elif isinstance(matrix, list):
+            self.wrapped = gp_Trsf()
+            if len(matrix) == 12:
+                self.wrapped.SetValues(*matrix)
+            elif len(matrix) == 16:
+                # Only use first 12 values - the last 4 must be [0, 0, 0, 1].
+                last4 = matrix[12:]
+                if last4 != [0., 0., 0., 1.]:
+                    raise ValueError("Expected the last 4 values to be [0,0,0,1], but got: {}".format(last4))
```
fragmuffin: Technically this isn't always true.
https://en.wikipedia.org/wiki/Scaling_(geometry)#Using_homogeneous_coordinates

personally I'd leave this check out; if an end user is explicitly defining affine transform matrices, then validity is arguably scoped in their code, not ours.

Also, it's technically _just_ a 4x4 matrix, it's not necessarily exclusively for the purposes of a transform (although unlikely).
But I do think the assumption given 12 elements is fair enough.

I'll get off my soapbox now, what do you think?

justbuchanan: Thanks for the feedback!

You're definitely right that in the general case, the last row isn't required to be `[0,0,0,1]`. Because the `Matrix` class is backed by OCC's `gp_Trsf` though, it is restricted to transform matrices and not arbitrary 4x4s.

The `gp_Trsf` `SetValues()` method only lets you set the first 3 rows and assumes the last to be `[0,0,0,1]` because it stores a "normalized" matrix.

It would be possible to assume the last row given as input is a more general `[0,0,0,x]`, then normalize the first three rows depending on the value of `x`. Do you think this would be better than just asserting that it's the expected `[0,0,0,1]`?

See the `SetValues()` docs here: https://www.opencascade.com/doc/occt-6.9.1/refman/html/classgp___trsf.html#a236867197c4fc79c57b5e34cf643005f.

fragmuffin: > Because the Matrix class is backed by OCC's `gp_Trsf` though, it is restricted to transform matrices and not arbitrary 4x4s

🤦‍♂️  you're absolutely right.
If I just read the next line properly 🙄 

> It would be possible to assume the last row given as input is a more general `[0,0,0,x]`, then normalise the first three rows depending on the value of `x`.

No, that's scope-creep.
If somebody posts an issue with a legitimate use case for this in the future, _then_ it could be added... but for now, I'm doing a full 180°, your original implementation is good. 👍

fragmuffin:
@justbuchanan : also... 👍 tests! I love me some tests.

cadquery/occ_impl/geom.py
adam-urbanczyk: Following the OCC convention (so gp_Trsf) makes sense here. NB: this is a low level "*_impl" object. "Normal" users won't be in principle touching it.

adam-urbanczyk:
I use numpy a lot for my day job. Having that experience I find the nested list solution more intuitive. It is not a strong preference though. What do other users think?

fragmuffin:
I find the nested list more intuitive too.
But mine is not a strong preference either.

adam-urbanczyk:
Alright, merging the PR. Thanks @justbuchanan  for your hard work.

PR #65, opened 2018-12-23
Added missing functions to Workplane

adam-urbanczyk:


adam-urbanczyk:
New sweep option does not work with the current implementation of sweep. Check how FreeCAD does this currently:

https://github.com/FreeCAD/FreeCAD/blob/master/src/Mod/Part/App/PartFeatures.cpp#L413

dcowden:
wow, thats some pretty impenetrable code

dcowden:
@adam-urbanczyk pipeshell and left in the FreeCAD based CQ was flakey too, but i don't remember seeing this particular one.  We commonly saw failures to generate, and weird behaviors at the boundaries.

That said, i use Solidworks and Onshape frequently, and even they have a hard time getting things like this right.  I usually have to provide a guide curve to get shapes like i want in those packages. Based on that experience, it might be unreasonable to assume that the default behavior should 'just work' without extra information, in the form of either guide paths, or boundary constraints

adam-urbanczyk:
OK, I will stop investigating this for now then. BTW: OCC `BRepOffsetAPI_MakePipeShell` API provides a lot of options - it would be nice to have more understanding of those at some point in the future.

tests/TestCadQuery.py
```diff
@@ -1326,6 +1452,80 @@ def testExplicitClean(self):
...
+            .rect(1.5, 3.5, forConstruction=True).vertices().cskHole(0.125, 0.25, 82, depth=None)
+        self.saveModel(result)
+
+        # bottom plane
+        s = Workplane(Plane.bottom())
+        result = s.rect(2.0, 4.0).extrude(0.5).faces(">Z").workplane()\
+            .rect(1.5, 3.5, forConstruction=True).vertices().cskHole(0.125, 0.25, 82, depth=None)
+        self.saveModel(result)
+
+    def testIsIndide(self):
```
jmwright: I think this should be `testIsInside`

jmwright:
@adam-urbanczyk Is this ready for review/approval, or is there more you want to add?

adam-urbanczyk:
@jmwright ready for review

tests/TestCadQuery.py
adam-urbanczyk: Fixed in b59e277

cadquery/occ_impl/shapes.py
```diff
@@ -1364,22 +1364,49 @@ def sweep(cls, outerWire, innerWires, path, makeSolid=True, isFrenet=False):
...
+    def sweep_multi(cls, profiles, path, makeSolid=True, isFrenet=False):
+        """
+        Multi section sweep. Only single outer profile per section is allowed.
+
+        :param profiles: list of profiles
+        :param path: The wire to sweep the face resulting from the wires over
+        :return: a Solid object
+        """
+        if path.ShapeType() == 'Edge':
+            path = Wire.assembleEdges([path, ])
```
jmwright: Is there supposed to be another item in the list after `path`?

adam-urbanczyk: No, just a convention I use (analogous to 1-tuples (el,) ).  Should I change this?

jmwright: I think it's fine. I just wanted to confirm that it was intentional.

adam-urbanczyk: OK

adam-urbanczyk:
@dcowden any comments on the PR? If not I'd like to merge.

dcowden:
Looks great to me! Though, at this point I think you guys know cq better than I do!

PR #67, opened 2018-12-29
Remove import STEP form URL

adam-urbanczyk:
This code is not compatible with py3 and not covered by any tests. Any thoughts?

### DIFF
--- a/cadquery/occ_impl/importers.py
+++ b/cadquery/occ_impl/importers.py
@@ -4,7 +4,6 @@ from .shapes import Shape
 
 import sys
 import os
-import urllib as urlreader
 import tempfile
 
 from OCC.STEPControl import STEPControl_Reader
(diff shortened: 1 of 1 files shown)


jmwright:
I feel like @dcowden added this in at some point for testing or demo purposes, but I can't remember any details and may be mistaken.

adam-urbanczyk:
If it is needed I can refactor, but as-is it is not py3 compatible.

dcowden:
This code supported the original www.parametricparts.com site, where you could load models shared by others via url.  Its ok to dump it for now, its not hard to re-add later

adam-urbanczyk:
OK, I'll remove for now. If needed in the future I propose to use requests over urllib

dcowden:
Agreed!

Issue #70, opened 2018-12-31
Extrude / cut with draft angle


osterwood:
It would be wonderful if extrude and cut operations had a key-word argument for draft angle.  This would allow positive or negative rake on the profile of the operation.

I know this is a tough enhancement to implement -- but it would be very useful to design for some kinds of manufacturing processes.  PythonOCC-core has a few references to draft angle in the codebase, but use is unclear to me.

dcowden:
this might be an interesting opportunity to compare pythonOCC and pyOCCT .  There has been some talk of using pyOCCT instead of pythonOCC.   If pyOCCT proved to have clearer and/or better documented interfaces to the draft features, it might bode well for it vs pythonOCC.

osterwood:
I’ll have to play around with that.  Thanks.  

Any idea how it would / could work with a blind cavity (e.g. not a thru cut into an existing solid)?

adam-urbanczyk:
@osterwood  take a look at #71 .

For now only low level function in `occ_impl`. Do you have any thoughts on API for cq? E.g.

* OCC kernel does not support tapers (at least the function I am using - feel free to browse thorough sources of FreeCAD or SALOME to find something better) with multiple base wires
* Only normal (or -normal) direction is supported

Regarding your question - I am not sure if I get what you mean. Could you paste a sketch?

adam-urbanczyk:
@osterwood any luck with #71 ? Is it going in the right direction for your use case?

adam-urbanczyk:
@osterwood did you have some time to try out my proposal?

adam-urbanczyk:
I added `extrude` and `cutBlind` with taper in #71. Closing this issue

jmwright:
Thanks @adam-urbanczyk !

PR #71, opened 2019-01-02
Initial implementation of tapered extrusion

adam-urbanczyk:
Tapred `extrude` and `cutBlind` operations. Teaser pic: 
afbeelding

Related to #70

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -47,7 +47,7 @@ before_install:
 - export PATH="$HOME/miniconda/bin:$HOME/miniconda/lib:$PATH";
 - hash -r;
 - conda config --set always_yes yes --set changeps1 no;
-- conda create -y -q -n test_cq -c pythonocc -c oce -c conda-forge -c dlr-sc pythonocc-core=0.18
+- conda create -y -q -n test_cq -c pythonocc -c oce -c conda-forge -c dlr-sc pythonocc-core=0.18.1
   pyparsing mock;
 - source ~/miniconda/bin/activate test_cq;
 - python -c 'import OCC.gp as gp; print(gp.gp_Vec())'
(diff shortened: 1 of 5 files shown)


dcowden:
cool addition Adam!

adam-urbanczyk:
I think it is ready to be merged. Somehow `cutBlind` requires negative depths. I wonder if it would make sense to reverse this sign convention.

NB: the coverage misses are not real - I did not touch the red files.

.travis.yml
```diff
@@ -47,7 +47,7 @@ before_install:
 - export PATH="$HOME/miniconda/bin:$HOME/miniconda/lib:$PATH";
 - hash -r;
 - conda config --set always_yes yes --set changeps1 no;
-- conda create -y -q -n test_cq -c pythonocc -c oce -c conda-forge -c dlr-sc pythonocc-core=0.18
+- conda create -y -q -n test_cq -c pythonocc -c oce -c conda-forge -c dlr-sc pythonocc-core=0.18.1
```
jmwright: The .travis.yml file on the adam-urbanczyk-OCC-version-update branch pulls pythonocc-core 0.18.2. Should we standardize on a version, or does it matter?

jmwright:
> Somehow `cutBlind` requires negative depths.

I've run into this periodically and just invert the normal on the workplane. I think the normal is oriented to assume extrusion, and so a cut operation requires a negative depth unless the normal is inverted. At least I think that's what's happening.

> NB: the coverage misses are not real - I did not touch the red files.

I have annoyances like this with coverage in this project and others. There's probably a way to make coverage less sensitive, but I have never checked on that option.

dcowden:
It looks good to me.

.travis.yml
adam-urbanczyk: Yes that is on purpose. I wan't to keep 0.18.1 explicitly for now (that is what master used to be tested with). Once the issues with 0.18.2 (Mac) are solved we can switch to that version.

adam-urbanczyk:
OK, merging. Tagging @osterwood for info.

PR #74, opened 2019-01-03
BoundingBox tolerance default bugfix

fragmuffin:
The `tol` default for `cadquery.BoundBox._fromTopoDS` was part of the signature.
Which means changing the tolerance in the module with something like:

```python
@mock.patch('cadquery.occ_impl.geom.TOL', 1e-7)
```

had no effect... but should work now.

note: this is a bit of a band-aid for #7

### DIFF
--- a/cadquery/occ_impl/geom.py
+++ b/cadquery/occ_impl/geom.py
@@ -790,10 +790,11 @@ class BoundBox(object):
         return None
 
     @classmethod
-    def _fromTopoDS(cls, shape, tol=TOL, optimal=False):
+    def _fromTopoDS(cls, shape, tol=None, optimal=False):
         '''
         Constructs a bounding box from a TopoDS_Shape
         '''
+        tol = TOL if tol is None else tol  # tol = TOL (by default)
         bbox = Bnd_Box()
         bbox.SetGap(tol)
         if optimal:


adam-urbanczyk:
@fragmuffin I don't think it is instability - I think it is runtime of this line:
 ```python
mesh = BRepMesh_IncrementalMesh(shape, TOL, True)
```
You are essentially requesting to perform meshing with 1e-5 relative precision (not sure how it is precisely defined though).

May I know why do you need such precision of bounding box calculations? Maybe there are other applicable solutions implemented in OCC

P.S. In newer versions of OCC there is an `AddOptimal` method that (to my understanding) does not require meshing.

fragmuffin:
@adam-urbanczyk

> May I know why do you need such precision of bounding box calculations?

Of course!; I'm mainly using `BoundingBox` for:
* **glTF [accessors](https://github.com/KhronosGroup/glTF/blob/master/specification/2.0/README.md#accessors)** - each needs a range of values contained in the buffer.
* **auto simplification** - if a `cqparts.Part` is passed `_simple=True`, but no `make_simple` method is implemented, then it gets the output of `make()` and returns the bounding box of the complex shape.
* **validation** - a basic confirmation that geometry is where it's supposed to be

Most of the time these don't require much precision.
The worst use-case I've seen so far is the auto-simplification of a unit cube `(xmin, xmax) = (-0.52, 0.52)` because the error is compounded.

Validation: I think I've been spoilt with `~1e-6` precision from freecad implementations, it made verifying shapes by bounding boxes easy, so this PR has come from reducing the `cqparts` `feature/freecad-occ` branches test-case failures to zero: cqparts/cqparts#143

**Conclusion**
I think this PR is worth merging, changing the tolerance _does_ work, but at a risk.
However, this risk is clear to any python coder because the `cadquery.occ_impl.geom.TOL` value has to be explicitly changed.

> In newer versions of OCC there is an `AddOptimal` method that (to my understanding) does not require meshing.

I  think this would be worthwhile in future, but out of scope for this PR.

Thanks @adam-urbanczyk !

adam-urbanczyk:
Looks good - merging.

fragmuffin:
Thanks @adam-urbanczyk ! :trophy:

PR #88, opened 2019-01-30
Use pythonocc 0.18.1 explicitly

adam-urbanczyk:
This should fix failing builds


adam-urbanczyk:
OSX fails as expected. I'm going to merge this to move things forward.

Issue #90, opened 2019-02-08
B-spline path with specified tangent vectors at the ends


adam-urbanczyk:
Use `GeomAPI_Interpolate` i.s.o. `GeomAPI_PointsToBSpline` and allow the user to specify tangents. This will enable creation of more predictable curves.

Cf. https://www.opencascade.com/doc/occt-6.9.1/refman/html/class_geom_a_p_i___interpolate.html

This will also allow to construct closed paths.

adam-urbanczyk:
Fixed in #92

PR #92, opened 2019-02-08
Use GeomAPI_Interpolate for BSpline construction

adam-urbanczyk:
This allows to specify tangents and create closed bspline curves

Solves #90 

* Spline with/without constraints 
afbeelding

* Closed spline (converted to Face) 
afbeelding

### DIFF
--- a/cadquery/occ_impl/shapes.py
+++ b/cadquery/occ_impl/shapes.py
@@ -104,6 +104,8 @@ from OCC.Visualization import Tesselator
 
 from OCC.LocOpe import LocOpe_DPrism
 
+from OCC.BRepCheck import BRepCheck_Analyzer
+
 from math import pi, sqrt
 
 TOLERANCE = 1e-6
@@ -315,7 +317,7 @@ class Shape(object):
         return self.wrapped.IsEqual(other.wrapped)
 
     def isValid(self):  # seems to be not used in the codebase -- remove?
-        raise NotImplemented
+        return BRepCheck_Analyzer(self.wrapped).IsValid()
 
     def BoundingBox(self, tolerance=0.1):  # need to implement that in GEOM
         return BoundBox._fromTopoDS(self.wrapped)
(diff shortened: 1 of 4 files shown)


adam-urbanczyk:
NB: I also refactored `tangentAt` for arbitrary location (in [0,1] parameter space) on an edge. This is used in the tests for the spline functionality.

jmwright:
@adam-urbanczyk I won't get a chance to look at this until the morning (US time), but I'll review it first thing.

cadquery/occ_impl/shapes.py
```diff
@@ -729,18 +730,28 @@ def makeCircle(cls, radius, pnt=Vector(0, 0, 0), dir=Vector(0, 0, 1), angle1=360
...
         Interpolate a spline through the provided points.
         :param cls:
         :param listOfVector: a list of Vectors that represent the points
+        :param tangents: tuple of Vectors specifying start and finish tangent
+        :param periodic: creation of peridic curves
+        :param tol: tolerance of the algorithm (consult OCC documentation)
         :return: an Edge
         """
-        pnts = TColgp_Array1OfPnt(0, len(listOfVector) - 1)
+        pnts = TColgp_HArray1OfPnt(1, len(listOfVector))
```
jmwright: Odd that this array isn't 0 based, but it looks like they did it intentionally.

tests/TestCadQuery.py
```diff
@@ -362,6 +362,32 @@ def testRevolveCone(self):
...
+        # Spline path - just a smoke test
+        path = Workplane("XZ").spline(pts).val()
+
+        # Closed spline
+        path_closed = Workplane("XZ").spline(pts,periodic=True).val()
+        self.assertTrue(path_closed.IsClosed())
+        
+        # attempt to build a face
+        w = Wire.assembleEdges([path_closed,])
+        f = Face.makeFromWires(w)
```
jmwright: Do you want to add an assert here to make sure that the face was built correctly?

adam-urbanczyk: Yes, good catch! Actually the `isValid` method was not implemented. Now I check if the generated face is valid.

cadquery/occ_impl/shapes.py
adam-urbanczyk: Yes, it is weird. I could only get `GeomAPI_Interpolate` to work with 1-based arrays.

PR #106, opened 2019-02-21
Support for text rendering 

### LINKED ISSUE #48: Implement Text Generation
<a href="https://github.com/dcowden"></img></a> **Issue by [dcowden](https://github.com/dcowden)**
_Sunday Dec 11, 2016 at 21:05 GMT_
_Originally opened as https://github.com/dcowden/cadquery/issues/166_

----

Find a way to generate text.

The ideal implementation requires a baseline and a face/surface-- the text should follow the surface.

adam-urbanczyk:
This will resolve #48 

afbeelding

### DIFF
--- a/cadquery/cq.py
+++ b/cadquery/cq.py
@@ -2268,6 +2268,20 @@ class Workplane(CQ):
             r = baseSolid.fuse(obj)
 
         return self.newObject([r])
+      
+    def _cutFromBase(self, obj):
+        """
+        Cuts the provided object from the base solid, if one can be found.
+        :param obj:
+        :return: a new object that represents the result of combining the base object with obj,
+           or obj if one could not be found
+        """
+        baseSolid = self.findSolid(searchParents=True)
+        r = obj
+        if baseSolid is not None:
+            r = baseSolid.cut(obj)
+
+        return self.newObject([r])
 
     def combine(self, clean=True):
         """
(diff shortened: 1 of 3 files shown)


adam-urbanczyk:
@jmwright  @dcowden I think it is ready to be merged. Please take a look at the API:
```python
import cadquery as cq
result = cq.Workplane("XY" ).box(4, 4, 0.5).edges("|Z").fillet(0.125).faces('>Z')\
   .workplane().text('CQ 2.0',0.5,-.05,cut=True,font='Consolas')
```

dcowden:
@adam-urbanczyk you are on FIRE! Its great to see this actually happen.

I love the API as you have it, and id recommend merging this. 

However, i have some insight to share for future work. TL;DR, you _might_ consider implementing on a per-character basis, even though as the API is now, you do not have to. More about that:...

It just so happens that i developed a text plugin for onshape. As a part of that, i learned a lot about what users want from a CAD text package.

There's a list of features that are necessary to be 'perfect' at least based on the user base I was working with:
* bold/italic
* flip/mirror text top/bottom and left/right
* automatically fit the text to a base curve. this coudl be a straight line, but is commonly a curve. This becomes tricker than it sounds, because you need to compute the width of the font, and adjust the height to fit. This can be hard because most fonts dont make it easy to do that.
* justify the text relative to the top, bottom, or middle of the guide curve
* kerning. This was a big deal. Kerning is the ability to slightly change the spacing between each letter, to account for the weirdness of the fonts, combined with following a guide curve
* apply on curved surfaces.  This gets complex because you need to offset the base surface, and then use that offset to limit to the tops of each letter. Its also necessary to extrude each letter via a separate surface normal, if the surface is curved.

I mention all this because when faced with all of those requirements, I ended up implementing the text as a per-character function, because there's really no other way to do it.

You can see some examples here: http://store.parametricparts.com/store/p10/CURVED-TEXT-2.0.html

have a look specifically at the input dialog, which shows all the options I ended up with.

cadquery/cq.py
```diff
@@ -2770,6 +2784,45 @@ def clean(self):
             raise AttributeError(
                 "%s object doesn't support `clean()` method!" % obj.ShapeType())
         return self.newObject(cleanObjects)
+      
+    def text(self, txt, fontsize, distance, cut=True, combine=False, clean=True,
```
jmwright: Is the cut parameter combined in the function call here because you need to keep the text as-is without converting to edges/wires? I ask because I could also see making the text into wires and doing a separate, chained cut or extrude operation with it.

tests/TestCadQuery.py
```diff
@@ -1800,3 +1800,17 @@ def testClose(self):
...
+    def testText(self):
+      
+        box = Workplane("XY" ).box(4, 4, 0.5)
+        
+        obj1 = box.faces('>Z').workplane().text('CQ 2.0',0.5,-.05,cut=True)
+        
+        #combined object should have smaller volume
+        self.assertGreater(box.val().Volume(),obj1.val().Volume())
+        
+        obj2 = box.faces('>Z').workplane().text('CQ 2.0',0.5,.05,cut=False,combine=True)
```
jmwright: It's not really necessary, but you could check to make sure that you have the right number of faces as well. The Arial font should always produce a predictable number of extra faces when extruded.

cadquery/cq.py
adam-urbanczyk: OCC function returns a compound of faces. I could convert it to a compound of wires in principle, sort it and extrude again, but this feels like a detour. I also thought that the `cut` use case is quite common.

tests/TestCadQuery.py
adam-urbanczyk: Good idea. Done

adam-urbanczyk:
@dcowden  cool plugin! On OCC level it is possible to render on per-character basis. I'll leave the features you mentioned for another PR though. I think implementing everything would take quite some effort.

adam-urbanczyk:
Alright, thanks for all your feedback. Merging.

Issue #37, opened 2018-12-06
2d Spiral primative


dcowden:
<a href="https://github.com/fragmuffin"><img src="https://avatars0.githubusercontent.com/u/2626190?v=4" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [fragmuffin](https://github.com/fragmuffin)**
_Tuesday Sep 19, 2017 at 11:04 GMT_
_Originally opened as https://github.com/dcowden/cadquery/issues/196_

----

In the interest of #195, we also need a method of making a spiral primative

something like:

```python
import cadquery
from Helpers import show
part = cadquery.Workplane("XY") \
    .spiral(radius=1, growth=2, radius=108).close() \
    .extrude(3)
show(part)
```

looking into https://freecadweb.org/wiki/Part_Spiral/en now

dcowden:
<a href="https://github.com/fragmuffin"><img src="https://avatars0.githubusercontent.com/u/2626190?v=4" align="left" width="48" height="48" hspace="10"></img></a> **Comment by [fragmuffin](https://github.com/fragmuffin)**
_Wednesday Sep 20, 2017 at 12:59 GMT_

----

FreeCAD's spiral primitive implementation:
[`App::DocumentObjectExecReturn *Spiral::execute(void)`](https://github.com/FreeCAD/FreeCAD/blob/949c86d364f05eecb5e7c279d6b2bc20e76e65ec/src/Mod/Part/App/PrimitiveFeature.cpp#L907)

It makes a wire along a conical surface, then projects it onto a plane.
And there doesn't seem to be an existing python hook for it.

adam-urbanczyk:
This is currently possible with parametric curves, cf. #110

PR #110, opened 2019-02-23
Parametric curves support

### LINKED ISSUE #91: Construction of edges based on parametric equations
Handy for creation of mathematically defined shapes like airfoils, cycloidal gears , spirals

This would be a general solution for  #37

adam-urbanczyk:
This will resolve #91 . Also in principle #37 could be solved by this PR.

 Sneak preview of the functionality:
```python
import cadquery as cq
from math import sin, cos,pi,floor

def hypocycloid(t,r1,r2):
    return ((r1-r2)*cos(t)+r2*cos(r1/r2*t-t),(r1-r2)*sin(t)+r2*sin(-(r1/r2*t-t)))

def epicycloid(t,r1,r2):
    return ((r1+r2)*cos(t)-r2*cos(r1/r2*t+t),(r1+r2)*sin(t)-r2*sin(r1/r2*t+t))

def gear(t,r1=2,r2=1):
    if (-1)**(1+floor(t/2/pi*(r1/r2))) < 0:
        return epicycloid(t,r1,r2)
    else:
        return hypocycloid(t,r1,r2)

result = cq.Workplane('XY').parametricCurve(lambda t: gear(t*2*pi,6,1)).twistExtrude(15,90)
show_object(result)
```
afbeelding

### DIFF
--- a/cadquery/cq.py
+++ b/cadquery/cq.py
@@ -1375,15 +1375,15 @@ class Workplane(CQ):
 
         return self.newObject([rv])
       
-    def parametricCurve(self, func, N=400, start=0, end=1):
+    def parametricCurve(self, func, N=400, start=0, stop=1):
         """
         Create a spline interpolated through the provided points.
 
         :param func: function f(t) that will generate (x,y) pairs
         :type func: float --> (float,float)
         :param N: number of points for discretization
-        :param start: start value of the parameter t
-        :param end: end value of the parameter t
+        :param start: starting value of the parameter t
+        :param stop: final value of the parameter t
         :return: a Workplane object with the current point unchanged
 
         """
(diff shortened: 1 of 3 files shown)


adam-urbanczyk:
I think it is ready for review. I made quite some changes to the spline function so please take a good look.

I also added one example to the docs.

cadquery/cq.py
```diff
@@ -1312,14 +1312,16 @@ def move(self, xDist=0, yDist=0):
         return self.newObject([self.plane.toWorldCoords(newCenter)])
 
     def spline(self, listOfXYTuple, tangents=None, periodic=False,
-               forConstruction=False):
+               forConstruction=False, includeCurrent=True, makeWire=False):
```
jmwright: I'm thinking the default values of the two new parameters have the correct defaults to make sure spline works as it always has. If so, I think this change is fine.

adam-urbanczyk: Indeed. Spline tests pass without any modifications.

dcowden:
+1 for this, looks great and adds hotness too.

adam-urbanczyk:
@dcowden +1 for hotness : ). Merging.

Issue #39, opened 2018-12-06
Sweep Buggy for Sharp Paths (polylines)


dcowden:
<a href="https://github.com/shivams"><img src="https://avatars0.githubusercontent.com/u/6936383?v=4" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [shivams](https://github.com/shivams)**
_Saturday Jul 08, 2017 at 02:51 GMT_
_Originally opened as https://github.com/dcowden/cadquery/issues/186_

----

If the sweep path has sharp edges (e.g. sweep path is a poly-line), then sweep behaves in a buggy manner, and the object that is being swept doesn't turn correctly with the sweep path. 

In the following example, you can see that the "polyline" sweep is very thin after the "turn".

```
import cadquery as cq
from Helpers import show

pts = [(0,1), (0,20), (10,25)]

poly_path = cq.Workplane("YZ").polyline(pts)
spline_path = cq.Workplane("YZ").spline(pts)

sweep_poly = cq.Workplane("XY").circle(0.3).sweep(poly_path)
sweep_spline = cq.Workplane("XY").transformed(offset=(0, -3, )).circle(0.3).sweep(spline_path)

show(sweep_poly, (204, 100, 50, 0.2))
show(sweep_spline, (50, 204, 80, 0.2))
```

![sweep-bug](https://user-images.githubusercontent.com/6936383/27982063-658007ba-63b6-11e7-9b02-52b2df1dc5dd.jpeg)

dcowden:
<a href="https://github.com/jmwright"><img src="https://avatars0.githubusercontent.com/u/1015439?v=4" align="left" width="48" height="48" hspace="10"></img></a> **Comment by [jmwright](https://github.com/jmwright)**
_Saturday Jul 08, 2017 at 02:59 GMT_

----

Hi @shivams 

What version of FreeCAD are you running, and on which operating system?

dcowden:
<a href="https://github.com/shivams"><img src="https://avatars0.githubusercontent.com/u/6936383?v=4" align="left" width="48" height="48" hspace="10"></img></a> **Comment by [shivams](https://github.com/shivams)**
_Saturday Jul 08, 2017 at 05:24 GMT_

----

Hi @jmwright 

I'm running FreeCAD 0.16 on Ubuntu 16.04. 

Doing sweep manually using the "Parts" workbench works fine in FreeCAD, though.

dcowden:
[...]

PR #112, opened 2019-02-26
Pipe transition mode fix

### LINKED ISSUE #94: Expose more API of BRepOffsetAPI_MakePipeShell
Specifically allow the user to select transition modes:

afbeelding

Cf. https://opencascade.blogspot.com/2009/12/surface-modeling-part4.html

adam-urbanczyk:
This PR will allow different transition styles for sweeps (see below). I propose to use the leftmost style as the default one.
afbeelding

It will resolve #94 and #39

### DIFF
--- a/cadquery/cq.py
+++ b/cadquery/cq.py
@@ -2275,6 +2275,8 @@ class Workplane(CQ):
         :param boolean combine: True to combine the resulting solid with parent solids if found.
         :param boolean clean: call :py:meth:`clean` afterwards to have a clean shape
         :param transition:
+            handling of profile orientation at C1 path discontinuities.
+            Possible values are {'transformed','round', 'right'} (default: 'right').
         :return: a CQ object with the resulting solid selected.
         """
 
--- a/cadquery/cq.py
+++ b/cadquery/cq.py
@@ -2263,7 +2263,8 @@ class Workplane(CQ):
             newS = newS.clean()
         return newS
 
-    def sweep(self, path, sweepAlongWires=False, makeSolid=True, isFrenet=False, combine=True, clean=True):
+    def sweep(self, path, sweepAlongWires=False, makeSolid=True, isFrenet=False,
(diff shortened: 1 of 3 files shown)


adam-urbanczyk:
@dcowden @jmwright I think it is ready. Maybe additional example would be beneficial too?

dcowden:
the functionality looks great, and i would agree on your choice of default.
I think its necessary to somehow convey what the choices of transition style are, either in the comments, or by changing the api to accept a set of constants.

which kind of brings up a more general question i've always struggled with.  How leaky should we allow our interfaces to be? From a purity viewpoint, we'd want CQ interfaces to be their own, with the idea that it could be in front of any modelling kernel. from that viewpoint, the right decision is to use our own definition for the transition mode, and even to contemplate that perhaps other kernels might allow other options.

on the other hand, from a practical viewpoint, this is based on OCC, and always will be.  You could argue that ultimately, we'll never have other transition modes, becuase you're using MakePipeShell from OCC, and it is what it is. from that viewpoint, we should actually just link to the MakePipeShell documentation, which <presumably> already covers all the details. This makes CQ a leaky abstraction over OCC, but with the benefit that we dont have to re-document and describe many of the underlying operations.

adam-urbanczyk:
I wrote something in the doc strings.  It is not much I have to admit. Ultimately we'd need to extend it (as is the case for other options/methods). Regarding staying OCC agnostic, I'd prefer to be realistic. It is unlikel that it will ever change. I don't even think there exists any other opensource B-Rep CAD kernel at all.

BTW: I came to the conclusion that our doc string format is note very readable. I'll open an issue to discuss if we'd like to change to e.g. nump/scipy format.

adam-urbanczyk:
@dcowden do you think it is good for merging as-is?

adam-urbanczyk:
Merging to keep things going. @dcowden I will open a ticket regarding documentation.

PR #113, opened 2019-02-28
cutThruAll based on BRepFeat_MakeDPrism

### LINKED ISSUE #98: Start using BRepFeat_*  for some operations
`cutThruAll` would be a logical use case. 

More information in the docs: https://www.opencascade.com/doc/occt-6.9.1/overview/html/occt_user_guides__modeling_algos.html#occt_modalg_9

adam-urbanczyk:
This PR will resolve #98 . 

### Before:
```python
import cadquery as cq
result = cq.Workplane("front").box(3.0, 4.0, 0.25).pushPoints ( [ ( 0,0.75 ),(0, -0.75) ]) \
    .polygon(6, 1.0).cutThruAll()
show_object(result)
```
afbeelding

### After:
```python
import cadquery as cq
result = cq.Workplane("front").box(3.0, 4.0, 0.25).faces('>Z').pushPoints ( [ ( 0,0.75 ),(0, -0.75) ]) \
    .polygon(6, 1.0).cutThruAll()
show_object(result)
```
afbeelding

### DIFF
--- a/cadquery/cq.py
+++ b/cadquery/cq.py
@@ -432,6 +432,20 @@ class CQ(object):
         else:
             raise ValueError("Cannot End the chain-- no parents!")
 
+    def _findType(self, types, searchStack=True, searchParents=True):
+
+        if searchStack:
+            for s in self.objects:
+                if isinstance(s, types):
+                    return s
+
+        if searchParents and self.parent is not None:
+            return self.parent._findType(types,
+                                         searchStack=True,
+                                         searchParents=True)
+
+        return None
+
     def findSolid(self, searchStack=True, searchParents=True):
         """
         Finds the first solid object in the chain, searching from the current node
@@ -452,19 +466,21 @@ class CQ(object):
         if the plugin implements a unary operation, or if the operation will automatically merge its
         results with an object already on the stack.
         """
-        #notfound = ValueError("Cannot find a Valid Solid to Operate on!")
 
-        if searchStack:
-            for s in self.objects:
-                if isinstance(s, Solid):
-                    return s
-                elif isinstance(s, Compound):
-                    return s
+        return self._findType((Solid, Compound), searchStack, searchParents)
 
-        if searchParents and self.parent is not None:
-            return self.parent.findSolid(searchStack=True, searchParents=searchParents)
+    def findFace(self, searchStack=True, searchParents=True):
+        """
+        Finds the first face object in the chain, searching from the current node
+        backwards through parents until one is found.
 
-        return None
+        :param searchStack: should objects on the stack be searched first.
+        :param searchParents: should parents be searched?
+        :raises: ValueError if no solid is found in the current object or its parents,
+            and errorOnEmpty is True
+        """
+
+        return self._findType(Face, searchStack, searchParents)
 
     def _selectObjects(self, objType, selector=None):
         """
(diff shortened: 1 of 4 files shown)


adam-urbanczyk:
@jmwright  I used different function for `cutThruAll`. Now the OCC kernel does the work of determining how deep it has to go. Using `Feat` operations will also allow to perfrom operations to nearest face etc in the future. I also added taper angle to the `cutThruAll` operation.

I do have one dilemma. Underlying function `BRepFeat_MakeDPrism` requires a face. If a face is selected on the stack that is not a problem. If no faces were ever selected I am not sure what would be the proper solution. For now I added a fallback in the `Workplane` layer that finds a face parallel to the current plane and closest to its origin. That is quite a bit of magic. Maybe simply stopping and throwing an error would be better?

@dcowden adding you for info. If you have bandwidth to review, please do!

cadquery/cq.py
```diff
@@ -2501,11 +2516,32 @@ def cutThruAll(self, positive=False, clean=True):
...
+            sel = selectors.PerpendicularDirSelector(self.plane.zDir)
+            faces = sel.filter(solidRef.Faces())
+            #then select the closest
+            sel = selectors.NearestToPointSelector(self.plane.origin.toTuple())
+            faceRef = sel.filter(faces)[0]
+
+        rv = []
+        for solid in solidRef.Solids():
+            s = solid.dprism(faceRef, wires, thruAll=True, additive=False,
+                             taper=-taper)
```
jmwright: I'm assuming the - sign here is intentional and is because of the way the OCC API works, but I want to confirm.

cadquery/occ_impl/shapes.py
```diff
@@ -1451,9 +1453,41 @@ def sweep_multi(cls, profiles, path, makeSolid=True, isFrenet=False):
...
+
+        sorted_profiles = sortWiresByBuildOrder(profiles)
+        shape = self.wrapped
+        basis = basis.wrapped
+        for p in sorted_profiles:
+            face = Face.makeFromWires(p[0],p[1:])
+            feat = BRepFeat_MakeDPrism(shape,
+                                       face.wrapped,
+                                       basis,
+                                       #basis.normalAt().toDir(),
```
jmwright: Not sure if this comment was left in intentionally.

jmwright:
> If no faces were ever selected I am not sure what would be the proper solution. For now I added a fallback in the Workplane layer that finds a face parallel to the current plane and closest to its origin. That is quite a bit of magic.

If it's something that is most likely to be the user's fault, which I think is the case here, I'd probably go with an error. If it's a glitch with the kernel or the API that happens periodically, a warning along with performing the face finding magic might be justified.

cadquery/cq.py
adam-urbanczyk: Indeed, minus sign is intended.

cadquery/occ_impl/shapes.py
adam-urbanczyk: Yes, will remove that.

adam-urbanczyk:
I did think about the face issue a little bit longer. I think we should (at least for now) include the magic for the sake of API consistency. I can execute e.g. the following script without any face selection:
```python
result = cq.Workplane("front").box(3.0, 4.0, 4).pushPoints ( [ ( 0,0) ]) \
    .polygon(6, 1.0).cutBlind(1)
```
and get:
afbeelding

Therfore I reason that the same should be possible when using `cutThruAll`.

Would you agree @jmwright ?

jmwright:
@adam-urbanczyk I think that's fine. Most people just want it to work, and that's what your solution does. If there's too much magic and it causes problems, we can always walk it back later.

adam-urbanczyk:
Alright, merging.

Issue #124, opened 2019-05-01
Shell.makeShell error


rusc444:
I'm trying to build polygons in cadquery from vertices/faces definitions found here:

[http://dmccooey.com/polyhedra/](url)

A code example for the tetrahedron:

```
import cadquery as cq
import math

c0 = math.sqrt(2)/4
vertices = [[c0, -c0,  c0], [c0,  c0, -c0], [-c0,  c0,  c0], [-c0, -c0, -c0]]
faces = [[0, 1, 2], [1, 0, 3], [2, 3, 0], [3, 2, 1]]

l_faces = []
for aface in faces:
    aface.append(aface[0])
    l_lines = []
    for avertex in range(len(aface)-1):
        l_lines.append(cq.Edge.makeLine(cq.Vector(*vertices[aface[avertex]]),
                       cq.Vector(*vertices[aface[avertex+1]])))
    wire = cq.Wire.combine(l_lines)
    l_faces.append(cq.Face.makeFromWires(wire))
ashell = cq.Shell.makeShell(l_faces)
```

but I get an error in makeShell method:

> File "C:\Miniconda3\envs\cq\lib\site-packages\cadquery\occ_impl\shapes.py", line 1024, in makeShell
    shell_builder.Add(face.wrapped)
TypeError: TopoDS_Builder_Add expected 3 arguments, got 2

As it seems, the correct method call has to be 

`shell_builder.Add(shell_wrapped, face.wrapped)`

instead of

`shell_builder.Add(face.wrapped)`

This modified code works for me and the result seems to be a cadquery closed shell.

The next problem is that I don't see any way to convert this shell to a cadquery solid to work with. Is this possible ?.

Thanks

adam-urbanczyk:
Thanks for reporting this @rusc444 ! Since you already found a solution, would you consider submitting a PR?

Regarding shell to solid conversion: It should be possible with the `BRepBuilderAPI_MakeSolid` OCC class. If you feel like it, you could try implement a `makeSolid` method (similar to `makeShell`)

rusc444:
Thank you for the quick response.

I am not a professional programmer or a specialized CAD user. I just like to program in Python and use a 3d printer in hobby projects and cadquery is a awesome tool to combine the two interests. I do not have much knowledge of the procedures involving code development in GitHub and I think that is better if you manage the PR.

Regarding the shell to solid conversion, thanks for the suggestion. I'll try to do it using PythonOCC scripting.

[...]

Issue #107, opened 2019-02-22
Text alignment support


adam-urbanczyk:
Support left, center and right aligned text rendering.

Continuation of #48

adam-urbanczyk:
Solved by #128

PR #128, opened 2019-05-17
Better makeFromWires and related changes

### LINKED ISSUE #21: polyline.mirrorX and mirrorY Don't Work as Expected
<a href="https://github.com/jmwright"></img></a> **Issue by [jmwright](https://github.com/jmwright)**
_Tuesday Apr 07, 2015 at 18:55 GMT_
_Originally opened as https://github.com/dcowden/cadquery/issues/83_

----

Had what looks like maybe a bug in CQ or the FreeCAD library posted to the CQ module's issue tracker.

https://github.com/jmwright/cadquery-freecad-module/issues/51

This needs to be investigated to verify that it is a bug and not normal behavior that's non-intuitive. If it's a bug, we'll have to figure out how to address it.

adam-urbanczyk:
This PR will resolve #21, #124 and #107

Example of text alignment:
image

### DIFF
--- a/cadquery/cq.py
+++ b/cadquery/cq.py
@@ -2840,7 +2840,7 @@ class Workplane(CQ):
         return self.newObject(cleanObjects)
 
     def text(self, txt, fontsize, distance, cut=True, combine=False, clean=True,
-             font="Arial", kind='regular',align='center',valign='center'):
+             font="Arial", kind='regular',halign='center',valign='center'):
         """
         Create a 3D text
 
@@ -2853,8 +2853,8 @@ class Workplane(CQ):
         :param boolean clean: call :py:meth:`clean` afterwards to have a clean shape
         :param str font: fontname (default: Arial)
         :param str kind: font type (default: Normal)
-        :param str align:  horizontal alignment (default: center)
-        :param str valign: horizontal alignment (default: center)
+        :param str halign: horizontal alignment (default: center)
+        :param str valign: vertical alignment (default: center)
         :return: a CQ object with the resulting solid selected.
 
         extrude always *adds* material to a part.
@@ -2868,7 +2868,7 @@ class Workplane(CQ):
 
         """
         r = Compound.makeText(txt,fontsize,distance,font=font,kind=kind,
-                              align=align, valign=valign, position=self.plane)
+                              halign=halign, valign=valign, position=self.plane)
 
         if cut:
             newS = self._cutFromBase(r)
--- a/cadquery/cq.py
+++ b/cadquery/cq.py
@@ -2842,7 +2842,7 @@ class Workplane(CQ):
         return self.newObject(cleanObjects)
 
     def text(self, txt, fontsize, distance, cut=True, combine=False, clean=True,
-             font="Arial", kind='regular'):
+             font="Arial", kind='regular',align='center',valign='center'):
         """
         Create a 3D text
 
@@ -2855,6 +2855,8 @@ class Workplane(CQ):
         :param boolean clean: call :py:meth:`clean` afterwards to have a clean shape
         :param str font: fontname (default: Arial)
         :param str kind: font type (default: Normal)
+        :param str align:  horizontal alignment (default: center)
+        :param str valign: horizontal alignment (default: center)
         :return: a CQ object with the resulting solid selected.
 
         extrude always *adds* material to a part.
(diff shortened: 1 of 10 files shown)


adam-urbanczyk:
@jmwright could you take a look? I am also thinking about changing the interface of `sweep`, namely use `multipleSections` i.s.o. `sweepAlongWires` . I find the latter (i.e. current parameter name) very confusing.

jmwright:
@adam-urbanczyk Your changes look good. Thanks for doing this.

Regarding the sweepAlongWires change: Assuming the change you want to make is user-facing, I always try to think about whether the change I'm going to make will break users' existing scripts. The sweepAlongWires interface is a fairly recent addition if I remember right, so maybe you could get away with changing it. I'm not sure.

cadquery/occ_impl/shapes.py
```diff
@@ -1564,29 +1577,11 @@ def sortWiresByBuildOrder(wireList, result=[]):
...
-    for k, v in result.items():
-        tmp = [k, ]
-
-        iterator = TopTools_ListIteratorOfListOfShape(v)
-        while iterator.More():
-            tmp.append(Wire(iterator.Value()))
-            iterator.Next()
-
-        rv.append(tmp)
+    # make a Face, NB: this might return
```
jmwright: Was this supposed to say something like "might return null"?

adam-urbanczyk: Indeed, fixed in the last commits. Should read: `# make a Face, NB: this might return a compound of faces`

adam-urbanczyk:
@jmwright  could you take another look. I want to handle #124 and #107 in this PR as well.

NB: I deprecated the `sweepAlongWires` in favor of `multisection` kwarg. Old version will still work but generates a warning.

BTW: once this is merge I think it is time for 2.0 RC.

cadquery/cq.py
```diff
@@ -2871,6 +2855,8 @@ def text(self, txt, fontsize, distance, cut=True, combine=False, clean=True,
         :param boolean clean: call :py:meth:`clean` afterwards to have a clean shape
         :param str font: fontname (default: Arial)
         :param str kind: font type (default: Normal)
+        :param str align:  horizontal alignment (default: center)
+        :param str valign: horizontal alignment (default: center)
```
jmwright: `horizontal` should be `vertical` in this instance

cadquery/cq.py
```diff
@@ -2858,7 +2842,7 @@ def clean(self):
         return self.newObject(cleanObjects)
 
     def text(self, txt, fontsize, distance, cut=True, combine=False, clean=True,
-             font="Arial", kind='regular'):
+             font="Arial", kind='regular',align='center',valign='center'):
```
jmwright: I don't know that it matters that much, but you might consider changing `align` to `halign` so that it mirrors the `valign` parameter.

tests/TestCadQuery.py
```diff
@@ -1874,4 +1876,28 @@ def testParametricCurve(self):
...
+            faces.append(Face.makeFromWires(wire))
+        
+        shell = Shell.makeShell(faces)
+        solid = Solid.makeSolid(shell)
+        
+        self.assertTrue(shell.isValid())
+        self.assertTrue(solid.isValid())
+        
+        self.assertEqual(len(solid.Vertices()),4)
+        self.assertEqual(len(solid.Faces()),4)
```
jmwright: GitHub is complaining about no newline here.

jmwright:
There's also the [sweep example](https://github.com/CadQuery/cadquery/blob/master/examples/Ex023_Sweep.py). You might want to consider adding the `multiselection` change to it.

adam-urbanczyk:
Thanks for the feedback @jmwright  I will rework. Regarding examples there is one already: https://github.com/CadQuery/cadquery/blob/adam-urbanczyk-robust-make-face/examples/Ex024_Sweep_With_Multiple_Sections.py

cadquery/cq.py
adam-urbanczyk: Agreed

tests/TestCadQuery.py
adam-urbanczyk: I added a newline but not sure why that would cause troubles

adam-urbanczyk:
One step closer to 2.0RC

PR #132, opened 2019-05-22
Added option to center workplane on projected origin.

mgreminger:
Added 'ProjectedOrigin' option to centerOptions parameter of workplane method to place the center of the workplane at the position of the origin projected onto the plane. Also added test converge for all of the centerOption parameter choices.

### DIFF
--- a/cadquery/cq.py
+++ b/cadquery/cq.py
@@ -261,16 +261,19 @@ class CQ(object):
 
         return self.objects[0].wrapped
 
-    def workplane(self, offset=0.0, invert=False, centerOption='CenterOfMass'):
+    def workplane(self, offset=0.0, invert=False, centerOption='CenterOfMass',
+                  origin=(0, 0, 0)):
         """
         Creates a new 2-D workplane, located relative to the first face on the stack.
 
         :param offset:  offset for the work plane in the Z direction. Default
         :param invert:  invert the Z direction from that of the face.
-        :param centerOption: how local origin of workplane is determined
+        :param centerOption: how local origin of workplane is determined.
+        :param origin: origin for plane center, requires 'ProjectedOrigin' centerOption.
         :type offset: float or None=0.0
         :type invert: boolean or None=False
         :type centerOption: string or None='CenterOfMass'
+        :type origin: tuple
         :rtype: Workplane object ( which is a subclass of CQ )
 
         The first element on the stack must be a face, a set of
@@ -285,11 +288,10 @@ class CQ(object):
              face/faces, if a face/faces was selected. If a vertex was
              selected, the origin will be at the vertex, and located
              on the face. The centerOption paramter sets how the center is defined.
-             Options are 'CenterOfMass', 'CenterOfBoundBox', 'ProjectedOrigin', or
-             'ProjectedGlobalOrigin'. 'CenterOfMass' and 'CenterOfBoundBox' are 
-             in relation to the selected face/faces. 'ProjectedOrigin' uses the 
-             current origin that may have been updated by calls to transformed or
-             center. 'ProjectedGlobalOrigin' projects the global (0,0,0) origin.
+             Options are 'CenterOfMass', 'CenterOfBoundBox', or 'ProjectedOrigin'.
+             'CenterOfMass' and 'CenterOfBoundBox' are in relation to the selected
+             face/faces. 'ProjectedOrigin' uses the supplied origin, which defaults
+             to (0, 0, 0).
            * The Z direction will be normal to the plane of the face,computed
              at the center point.
            * The X direction will be parallel to the x-y plane. If the workplane is  parallel to
@@ -338,6 +340,9 @@ class CQ(object):
                 xd = Vector(1, 0, 0)
             return xd
 
+        if centerOption not in {'CenterOfMass', 'ProjectedOrigin', 'CenterOfBoundBox'}:
+            raise ValueError('Undefined centerOption value provided.')
+
         if len(self.objects) > 1:
             # are all objects 'PLANE'?
             if not all(o.geomType() in ('PLANE', 'CIRCLE') for o in self.objects):
@@ -348,12 +353,10 @@ class CQ(object):
             if not all(_isCoPlanar(self.objects[0], f) for f in self.objects[1:]):
                 raise ValueError("Selected faces must be co-planar.")
 
-            if centerOption in {'CenterOfMass', 'ProjectedOrigin', 'ProjectedGlobalOrigin'}:
+            if centerOption in {'CenterOfMass', 'ProjectedOrigin'}:
                 center = Shape.CombinedCenter(self.objects)
             elif centerOption == 'CenterOfBoundBox':
                 center = Shape.CombinedCenterOfBoundBox(self.objects)
-            else:
-                raise ValueError('Undefined value passed to centerOption')
 
             normal = self.objects[0].normalAt()
             xDir = _computeXdir(normal)
@@ -362,7 +365,7 @@ class CQ(object):
             obj = self.objects[0]
 
             if isinstance(obj, Face):
-                if centerOption in {'CenterOfMass', 'ProjectedOrigin', 'ProjectedGlobalOrigin'}:
+                if centerOption in {'CenterOfMass', 'ProjectedOrigin'}:
                     center = obj.Center()
                 elif centerOption == 'CenterOfBoundBox':
                     center = obj.CenterOfBoundBox()
@@ -370,7 +373,7 @@ class CQ(object):
                 xDir = _computeXdir(normal)
             else:
                 if hasattr(obj, 'Center'):
-                    if centerOption in {'CenterOfMass', 'ProjectedOrigin', 'ProjectedGlobalOrigin'}:
+                    if centerOption in {'CenterOfMass', 'ProjectedOrigin'}:
                         center = obj.Center()
                     elif centerOption == 'CenterOfBoundBox':
                         center = obj.CenterOfBoundBox()
@@ -382,9 +385,7 @@ class CQ(object):
 
         # update center to projected origin if desired
         if centerOption == 'ProjectedOrigin':
-            center = self.plane.origin.projectToPlane(Plane(center, xDir, normal))
-        elif centerOption == 'ProjectedGlobalOrigin':
-            center = Vector(0,0,0).projectToPlane(Plane(center, xDir, normal))
+            center = Vector(origin).projectToPlane(Plane(center, xDir, normal))
 
         # invert if requested
         if invert:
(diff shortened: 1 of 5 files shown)


adam-urbanczyk:
Thanks for your contribution @mgreminger . There was already interest in discussion of having such functionality so we really appreciate your work!

I will have some review comments. Mainly about moving the projection function to `cq.occ_impl.geom` , would you be interested in making small adjustments to your code?

mgreminger:
Thanks, @adam-urbanczyk, glad it's useful. Yes, I'm definitely willing to update the code based on your comments.

mgreminger:
@adam-urbanczyk, I see what you mean about putting the projection code in cq.occ_impl.geom, there's already a stub method for the projection for Vector in geom.py (see below). I'll go ahead a through the implementation in there. 

```
    def projectToPlane(self):
        raise NotImplementedError(
            "Have not needed this yet, but FreeCAD supports it!")
```

mgreminger:
Here's the [freecad implementation](https://github.com/FreeCAD/FreeCAD/blob/55db1feb2b9ed0dad3fda1c5fde63d6b28478423/src/Base/Vector3D.cpp#L215) of this method. Seems to be a vectorized version of the computation (with some C++ magic thrown in), which is a little cleaner. I'll try to implement it vectorized if I can figure out the equation, otherwise I'll use the implementation that I have worked out already.

mgreminger:
This should be ready for review now. I've moved the point-to-plane projection code to geom.py and it now uses a vectorized projection equation that matches what freecad uses for its implementation of the projectToPlane method. There is now full test coverage of the new code as well. Let me know if any further changes are required.

cadquery/occ_impl/geom.py
```diff
@@ -143,9 +143,31 @@ def distanceToPlane(self):
         raise NotImplementedError(
             "Have not needed this yet, but FreeCAD supports it!")
 
-    def projectToPlane(self):
-        raise NotImplementedError(
-            "Have not needed this yet, but FreeCAD supports it!")
+    def projectToPlane(self, *args):
```
adam-urbanczyk: It would be better to support only Plane input. The user can always construct a Plane object from two Vector objects if he needs to.

cadquery/occ_impl/geom.py
```diff
@@ -143,9 +143,31 @@ def distanceToPlane(self):
...
+            base = args[0]
+            normal = args[1]
+        elif len(args) == 1:
+            base = args[0].origin
+            normal = args[0].zDir
+        else:
+            raise TypeError("Expected 1 or 2 arguments consisting of 1 Plane object \
+or a base Vector and a normal Vector object.")
+
+        result = self-normal*(((self-base).dot(normal))/normal.Length**2)
```
adam-urbanczyk: I'd return here: 
```python
return self-normal*(((self-base).dot(normal))/normal.Length**2)
```
so no in-place modification will surprise our users.

cadquery/cq.py
```diff
@@ -371,6 +374,10 @@ def _computeXdir(normal):
                     raise ValueError(
                         "Needs a face or a vertex or point on a work plane")
 
+        # update center to projected origin if desired
+        if centerOption == 'ProjectedOrigin':
+            center = Vector(0, 0, 0).projectToPlane(center, normal)
```
adam-urbanczyk: I am a little bit confused - I thought that you intended to project current origin (i.e. `self.plane.origin`) onto the new plane. Here it seems that you are projecting a hardcoded point `Vector(0,0,0)` instead. Is this indeed the intention?

cadquery/occ_impl/geom.py
mgreminger: Agreed.

cadquery/occ_impl/geom.py
mgreminger: Agreed.

cadquery/cq.py
mgreminger: The intent is to project the global origin onto the workplane. The use case where I needed this is when I'm using an unusually shaped face to create a workplane (see picture below where I'm using the highlighted faces to create a new workplane). The center of mass or bounding box center of the selected faces are not intuitive to use as an origin for additional features. However, if I can project the origin (highlighted in red in the figure below) onto the workplane defined by the faces, I have a convenient center to reference for creating the pockets. The projection code is general and can project an arbitrary point to the surface. However, I thought a common use case would be to project the global origin onto the workplane to define its center. Often when I'm creating parts in SolidWorks, I'll reference dimensions off of the global origin in a similar fashion.

image

adam-urbanczyk: OK, got it. If you use `self.plane.origin` it will be more generic. The default value is `(0,0,0)` so it will support your use case but it will also allow users to to project other origin values specified with a `moveTo(x,y).workplane()` call.

mgreminger: I don't think this will work since the moveTo(x,y) commands will accumulate if we do this and the absolute origin will be lost without reversing all of the previous moveTo(x,y) calls in the stack. Changing Vector(0,0,0) to self.plane.origin results in the part shown below (the user space code is unmodified only the suggested change to the workplane() method from (0,0,0) to self.plane.origin was made). The first pocket comes out fine because it is after the first move. The other pockets end up off of the part since the move commands are accumulating and the absolute origin is lost.

image

If I understand your suggestion correctly, you can accomplish what you're suggesting by leaving the projection from Vector(0,0,0) in the workplane() method and using the moveTo after creating the workplane: `workplane(centerOption='ProjectedOrigin').moveTo(x,y)` This is the approach I'm using in the code to create the pockets in the part.

adam-urbanczyk: Weird, could you share the code used to generate your part? I'd like to investigate

mgreminger: In addition to `moveTo()` calls, I had used `transformed()` calls with the offset option, which changes `plane.origin` (I'm still learning how all this works). I did this because I wanted to rotate the workplane for certain features. You're right, if I had used only `moveTo()` calls, my code would probably have worked both ways (I did a simple test to confirm this). 

Would it be better to call my proposed option `workplane(centerOption='ProjectedGlobalOrigin')` rather than `workplane(centerOption='ProjectedOrigin')` and then have the `workplane(centerOption='ProjectedOrigin')` use `self.plane.origin` like you suggest? I want the ability to go back to the global origin even if I've used transformed() or center() calls.

mgreminger: I updated the pull request with the change to have both the `ProjectedOrigin` option that uses `self.plane.origin` and the `ProjectedGlobalOrigin` that uses `Vector(0,0,0)`. I think part of the confusion was me using the term origin when I really meant global origin. The behavior of these two options is only different if there has been a call to `transformed` (with the offset option) or to `center`, both of which update the current plane's origin (there may be other methods that update the plane origin that I'm not aware of). The examples below illustrate the difference.

I've also updated to pull request to incorporate the resolved review comments.

**Using moveTo with ProjectedOrigin**
```
import cadquery as cq

model = cq.Workplane('XY').rect(100,100).extrude(10)
model = model.faces('>Z').workplane(centerOption='ProjectedOrigin').moveTo(-40,-40).hole(5)
model = model.faces('>Z').workplane(centerOption='ProjectedOrigin').moveTo(40,40).hole(5)
```
image

**Using moveTo with ProjectedGlobalOrigin (no difference)**
```
import cadquery as cq

model = cq.Workplane('XY').rect(100,100).extrude(10)
model = model.faces('>Z').workplane(centerOption='ProjectedGlobalOrigin').moveTo(-40,-40).hole(5)
model = model.faces('>Z').workplane(centerOption='ProjectedGlobalOrigin').moveTo(40,40).hole(5)
```
image

**Using center with ProjectedOrigin (changes result)**
```
import cadquery as cq

model = cq.Workplane('XY').rect(100,100).extrude(10)
model = model.faces('>Z').workplane(centerOption='ProjectedOrigin').center(-40,-40).hole(5)
model = model.faces('>Z').workplane(centerOption='ProjectedOrigin').center(40,40).hole(5)
```
image

**Using center with ProjectedGlobalOrigin (same result as when using moveTo)**
```
import cadquery as cq

model = cq.Workplane('XY').rect(100,100).extrude(10)
model = model.faces('>Z').workplane(centerOption='ProjectedGlobalOrigin').center(-40,-40).hole(5)
model = model.faces('>Z').workplane(centerOption='ProjectedGlobalOrigin').center(40,40).hole(5)
```
image

mgreminger:
Back to one option with origin keyword argument as suggested. However, I'm using (0, 0, 0) as the default origin. This matches the pattern of the `mirror()` method of the same class so I feel it should be acceptable: https://github.com/CadQuery/cadquery/blob/10118e563add4e60f21c882ff2a842757ceaacad/cadquery/cq.py#L762

adam-urbanczyk:
I'd really propose to have the following interface
```python
    def workplane(self, offset=0.0, invert=False, centerOption='CenterOfMass',
                  origin=None):
```
And somewhere in the function body:
```python
if origin is None: origin = self.plane.origin
```
Indeed there are some keyword arguments here and there that default to (0,0,0) but the example you give is not a local operation. I'd argue that this one is - it depends on the state of the stack. I really think that it fits the current fluent api better if by default we project current origin.

Pinging @jmwright for second opinion.

mgreminger:
Implemented it as you suggested. Had to use the following to assign origin since the `plane` member of `self` doesn't always exist and 10 existing tests failed without limiting the assignment to the cases using the ProjectedOrigin option:
```python
if origin is None and centerOption == 'ProjectedOrigin': 
    origin = self.plane.origin
elif isinstance(origin, tuple):
    origin = Vector(origin)
```
Does raise issue that `self.plane` is not guaranteed to exist . The tests that created geometry as part of the constructor to CQ were the ones that failed. The following is an example where `self.plane` deosn't exist:
https://github.com/CadQuery/cadquery/blob/10118e563add4e60f21c882ff2a842757ceaacad/tests/TestCadQuery.py#L1147-L1154

mgreminger:
Cleaned it up a bit, put all of the 'ProjectedOrigin' code together and it only looks at `origin` if centerOption is set to 'ProjectedOrigin'
```python
        # update center to projected origin if desired
        if centerOption == 'ProjectedOrigin':
            if origin is None:
                origin = self.plane.origin
            elif isinstance(origin, tuple):
                origin = Vector(origin)
            center = origin.projectToPlane(Plane(center, xDir, normal))
```

jmwright:
@adam-urbanczyk I see that you pinged me on this, but it looks like you and @mgreminger already worked it out. From the very early days we've always tried to do what we thought would be intuitive for the user. We've failed plenty of times, but that's been a core guiding principle. I think that's still a good guiding principle, but it can be hard to predict what "intuitive" means sometimes.

adam-urbanczyk:
Thanks for all your hard work @mgreminger ! @jmwright I think this can be merged, do you agree?

jmwright:
@adam-urbanczyk Looks good, I agree. Thanks @mgreminger !

mgreminger:
@adam-urbanczyk and @jmwright,  you're welcome, glad to help. @adam-urbanczyk, thanks for the valuable feedback, it led to a final version that's much better than where I started.

PR #135, opened 2019-05-30
Update cutThruAll to remove unused positive argument (fixes #134)

### LINKED ISSUE #134: cutThruAll produces incorrect results for offset workplane
cutThruAll produces incorrect results for an offset workplane and the positive parameter has no effect. cutBlind produces the correct results for an offset workplane. See the examples below to reproduce the issue.

cutBlind produces the correct results for an offset workplane:

**Positive blind cut:** 
```
import cadquery as cq
test = cq.Workplane('XY').rect(100,100).extrude(100)
test = test.faces("<Z").workplane(offset=50,invert=True).rect(10,10).cutBlind(50)
```
image

**Negative blind cut:**
```
import cadquery as cq
test = cq.Workplane('XY').rect(100,100).extrude(100)
test = test.faces("<Z").workplane(offset=50,invert=True).rect(10,10).cutBlind(-50)
```
image

cutThruAll produces incorrect results for the same workplane:

**Positive thru-all cut:**
```
import cadquery as cq
test = cq.Workplane('XY').rect(100,100).extrude(100)
test = test.faces("<Z").workplane(offset=50,invert=True).rect(10,10).cutThruAll(positive=True)
```
image

**Negative thru-all cut:**
```
import cadquery as cq
test = cq.Workplane('XY').rect(100,100).extrude(100)
test = test.faces("<Z").workplane(offset=50,invert=True).rect(10,10).cutThruAll(positive=False)
```
image

This issue is most likely related to the face finding algorithm referenced here: https://github.com/CadQuery/cadquery/pull/113#issuecomment-469850786 . There are likely issues with this case since the model doesn't have a face adjacent to the offset workplane and ends up using the top or bottom face instead. The positive input argument is not referenced in the implementation of the method which explains why it has no effect.

mgreminger:
Removed positive argument for cutThruAll and clarified doc string to indicate that cut is in both directions normal to the workplane. Previous doc string indicated that cut was only on one side of plane. Fixes issue #134.

### DIFF
--- a/cadquery/cq.py
+++ b/cadquery/cq.py
@@ -2507,15 +2507,14 @@ class Workplane(CQ):
 
         return self.newObject([s])
 
-    def cutThruAll(self, positive=False, clean=True, taper=0):
+    def cutThruAll(self, clean=True, taper=0):
         """
         Use all un-extruded wires in the parent chain to create a prismatic cut from existing solid.
+        Cuts through all material in both normal directions of workplane.
 
         Similar to extrude, except that a solid in the parent chain is required to remove material
         from. cutThruAll always removes material from a part.
 
-        :param boolean positive: True to cut in the positive direction, false to cut in the
-            negative direction
         :param boolean clean: call :py:meth:`clean` afterwards to have a clean shape
         :raises: ValueError if there is no solid to subtract from in the chain
         :return: a CQ object with the resulting object selected


adam-urbanczyk:
Thanks for the cleanup @mgreminger !

Issue #139, opened 2019-06-05
Intermittent missing faces for face selectors


mgreminger:
**Description**
Sometimes a face or two is missing from a face selector call. It doesn't happen very often but it does lead to errors that seem to occur randomly. This sometimes leads to the test suite failing on code that will work the next time you run the test suite (see example [here](https://travis-ci.org/CadQuery/cadquery/jobs/536938386#L1384) for a test failure for code that passes most of the time). These errors occur on the tests that rely on returning the correct number of faces.

**How to Reproduce**
The code below reliably reproduces this issue by running one of the tests 1000 times. This code is based on the `testRectArray` test in `TestCadQuery.py`.

```python
import cadquery as cq

tally = {}
n = 1000

for i in range(n):
    NUMX = 3
    NUMY = 3
    s = cq.Workplane("XY").box(40, 40, 5, centered=(True, True, True)).faces(
           ">Z").workplane().rarray(8.0, 8.0, NUMX, NUMY, True).circle(2.0).extrude(2.0)

    count = tally.get(s.faces().size(),0) + 1
    tally[s.faces().size()] = count
    
for key, value in tally.items():
    print(f"{key} faces is returned in {value} out of {n} runs")
```
*The above produces the following output (the numbers will vary):*
```
24 faces is returned in 994 out of 1000 runs
23 faces is returned in 5 out of 1000 runs
22 faces is returned in 1 out of 1000 runs
```
**Likely Cause**
Occasionally the hash codes returned for the opencascade objects are repeating leading to some objects being left out when they are accumulated in the [_collectProperty method of CQ](https://github.com/CadQuery/cadquery/blob/9aaf110c59837e5b8915f6065950311c2a9ca478/cadquery/cq.py#L84-L110). This is occurring because the value for `HASH_CODE_MAX` is too small. It is defined here in shapes.py:
https://github.com/CadQuery/cadquery/blob/9aaf110c59837e5b8915f6065950311c2a9ca478/cadquery/occ_impl/shapes.py#L117
[...]

PR #140, opened 2019-06-05
Increase HASH_CODE_MAX

mgreminger:
Increase HASH_CODE_MAX to 2147483647 (C MAX_INT) to prevent hash collisions that result in occasional faces missing from face selector calls. Addresses issue #139.

### DIFF
--- a/cadquery/occ_impl/shapes.py
+++ b/cadquery/occ_impl/shapes.py
@@ -114,7 +114,7 @@ from functools import reduce
 
 TOLERANCE = 1e-6
 DEG2RAD = 2 * pi / 360.
-HASH_CODE_MAX = 2147483647  # MAX_INT, required by OCC.Core.HashCode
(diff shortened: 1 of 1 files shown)


cadquery/occ_impl/shapes.py
```diff
@@ -114,7 +114,7 @@
 
 TOLERANCE = 1e-6
 DEG2RAD = 2 * pi / 360.
-HASH_CODE_MAX = int(1e+6)  # required by OCC.Core.HashCode
+HASH_CODE_MAX = 2147483647  # MAX_INT, required by OCC.Core.HashCode
```
adam-urbanczyk: Could you maybe add: "max 32bit signed int" in the comment?

mgreminger: Yes, makes sense.

adam-urbanczyk:
Thanks, merging.

PR #143, opened 2019-06-08
Remove the combine arg for bool ops and get rid of side-effects

### LINKED ISSUE #133: Is the mode "combine=True" of function union correctly implemented?
I looked at `union` and the way `combine=True` works:
```
def union(...)
...
        if combine and solidRef is not None:
            r = solidRef.fuse(newS)
            solidRef.wrapped = newS.wrapped
...
```

Maybe I don't understand the mechanism, but shouldn't it be

```
        if combine and solidRef is not None:
            r = solidRef.fuse(newS)
            solidRef.wrapped = r.wrapped
```

so that `r` (the result) and `solidRef` (the original) point to the new object?

adam-urbanczyk:
This PR will resolve #133 

I removed the combine keyword argument from union, intersect and cut. This means that those functions never have side-effects, new cq stack object is always created.

### DIFF
--- a/cadquery/cq.py
+++ b/cadquery/cq.py
@@ -2378,13 +2378,11 @@ class Workplane(CQ):
 
         return self.newObject([s])
 
-    def union(self, toUnion=None, combine=True, clean=True):
+    def union(self, toUnion=None, clean=True):
         """
         Unions all of the items on the stack of toUnion with the current solid.
         If there is no current solid, the items in toUnion are unioned together.
-        if combine=True, the result and the original are updated to point to the new object
-        if combine=False, the result will be on the stack, but the original is unmodified
-
+        
         :param toUnion:
         :type toUnion: a solid object, or a CQ object having a solid,
         :param boolean clean: call :py:meth:`clean` afterwards to have a clean shape
@@ -2409,9 +2407,8 @@ class Workplane(CQ):
         # now combine with existing solid, if there is one
         # look for parents to cut from
         solidRef = self.findSolid(searchStack=True, searchParents=True)
-        if combine and solidRef is not None:
+        if solidRef is not None:
             r = solidRef.fuse(newS)
-            solidRef.wrapped = newS.wrapped
         else:
             r = newS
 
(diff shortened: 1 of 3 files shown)


adam-urbanczyk:
@bernhard-42 let me know if such a solution is more intuitive

bernhard-42:
I find it much more intuitive. I believe in "explicit is better than implicit" and hence the side effects were really confusing. I'd love to see this being merged into master.

Re combine for extrude, ... I'll continue in the original issue

bernhard-42:
@adam-urbanczyk @jmwright 
Just came across `split`. Should we remove the side effects too?

From
```python
        if keepTop and keepBottom:
            # Put both on the stack, leave original unchanged.
            return self.newObject([top, bottom])
        else:
            # Put the one we are keeping on the stack, and also update the
            # context solidto the one we kept.
            if keepTop:
                solid.wrapped = top.wrapped
                return self.newObject([top])
            else:
                solid.wrapped = bottom.wrapped
                return self.newObject([bottom])
```
To
```python
        if keepTop and keepBottom:
            # Put both on the stack, leave original unchanged.
            return self.newObject([top, bottom])
        else:
            # Put the one we are keeping on the stack, and also update the
            # context solidto the one we kept.
            if keepTop:
                return self.newObject([top])
            else:
                return self.newObject([bottom])
```

Here I don't see any need for the combination.

adam-urbanczyk:
Indeed `split` and `loft` had that too. Removed in the latest commit.

bernhard-42:
great, thanks!

adam-urbanczyk:
@jmwright I think this PR is ready. Could you take a look?

jmwright:
@adam-urbanczyk I think the code looks good. A couple of thoughts/questions.

1. Will these changes break any of the examples in the documentation?
2. We're probably going to end up with a few support requests on the Google Group, with users wondering why their CQ 1.0 scripts won't run in CQ 2.0 anymore. I think it's unavoidable and necessary to make breaking changes like this, we'll just need to help users through the transition.

adam-urbanczyk:
Good point, I'll check the examples.

adam-urbanczyk:
Examples run fine now.

jmwright:
Great, thanks. I don't see any reason this PR can't be merged now.

adam-urbanczyk:
Alright, merging. Thanks for the feedback @jmwright .

PR #146, opened 2019-06-10
Use a different font for testing

adam-urbanczyk:
This is a fix for build issues on azure


adam-urbanczyk:
@dcowden @jmwright I am trying to configure building conda packages on azure. So far it works on Win and Mac but it seems that our default font is not available on their Ubuntu image. I need to merge this to sovle this issue.

PR #167, opened 2019-07-29
Fixed bug where tolerance parameter of BoundingBox had no effect

mgreminger:
The tolerance parameter of the BoundingBox method of shape had no effect. Fixed this by passing the tolerance to the _fromTopoDS call. Changed the tolerance default value from 0.1 to None so that the global TOL is used by default.  This allows the user to set the global TOL value as outlined in #74.  The CenterofBoundbox method incorrectly passed a shape to the BoundingBox method, which is the position for the tolerance paramter. This has been fixed.  The _fromTopoDS method hard coded the global variable TOL in it's call to BRepMesh_IncrementalMesh. This has been updated to use the user supplied tolerance if one has been provided. Added test coverage for the tolerance parameter of the BoundingBox method.

### DIFF
--- a/cadquery/occ_impl/geom.py
+++ b/cadquery/occ_impl/geom.py
@@ -827,7 +827,7 @@ class BoundBox(object):
             raise NotImplementedError
             # brepbndlib_AddOptimal(shape, bbox) #this is 'exact' but expensive - not yet wrapped by PythonOCC
         else:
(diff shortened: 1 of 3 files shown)


jmwright:
@mgreminger Thanks for the contribution.

@adam-urbanczyk I'll merge this if you're ok with it.

adam-urbanczyk:
Looks good @jmwright , thanks for the fix @mgreminger !

Issue #169, opened 2019-08-07
CadQuery2 Dockerfile


roipoussiere:
I made a Dockerfile for CadQuery2:

- [docker-hub](https://hub.docker.com/r/roipoussiere/cadquery2/dockerfile)
- [source ](https://github.com/roipoussiere/cadquery/blob/master/docker/Dockerfile)

It is based on [nvidia's cudagl image](https://hub.docker.com/r/nvidia/cudagl) in order to use OpenGL in Docker.

For now, CadQuery2 is installed from sources and not with Conda, because I experienced [some issues](https://github.com/CadQuery/cadquery/issues/168) with this last.

I'm interested to switch to a Conda-based installation, because the compilation process is very long and reaches [Docker hub limitations](https://forums.docker.com/t/automated-build-resource-restrictions/1413/2), so I need to build the image locally then upload it to docker hub directly. The drawback is the dockerfile page on docker hub does not have many information (no Dockerfile, no description, etc.).

Usage:

```
echo "import cadquery; print(cadquery.Workplane('XY').box(1,2,3).toSvg())" > cadquery_script.py
cat cadquery_script.py | docker run --rm -i roipoussiere/cadquery2 > 3d_model.svg
```

You can read more on the [dedicated readme](https://github.com/roipoussiere/cadquery/blob/master/docker/readme.md).

I'm writing this issue before submitting a pull request in order to get some feedback about this.

In particular, I note that there is no CLI script in CadQuery2, like in [CadQuery vanilla](https://github.com/dcowden/cadquery/blob/master/cq_cmd.py). Should I create it, based on this one, or is there a specific reason to not include it?

There are many improvements to do, put it works enough for a proof of concept.

Things I would like to do before an eventual merge:
- get your feedback;
- Conda-based installation;
- make the container [CQGI-compliant](https://cadquery.readthedocs.io/en/latest/cqgi.html);
- allow input and output volumes to be mounted.

jmwright:
@roipoussiere Thanks for all your work on this. I'm not aware of any reason not to include the CLI. I expect we just forgot it in the transition.

roipoussiere:
> I'm not aware of any reason not to include the CLI. I expect we just forgot it in the transition.

Ok, I will include it!

Since there is now a Github organization for all CadQuery-related projects, it makes sense for me to do the same thing on docker hub.

[...]

PR #176, opened 2019-08-13
CadQuery dockerfile

roipoussiere:
This is a work in progress: cadquery docker image is partially functional, this should not be merged for now.

- related issue: #169
- [docker hub image](https://cloud.docker.com/u/cadquery/repository/docker/cadquery/cadquery)

I finally managed to build the image on docker-hub, by splitting images: cadquery image is built from [pythonocc-core image](https://github.com/roipoussiere/pythonocc-core) (also on [docker hub](https://cloud.docker.com/u/cadquery/repository/docker/cadquery/pythonocc-core)), which itself is built from [oce image](https://github.com/roipoussiere/oce) (also on [docker hub](https://cloud.docker.com/u/cadquery/repository/docker/cadquery/oce)). The image build is now much master and respects docker hub's build limitations (max 2h to build an image).

That's allow us to use docker hub automated builds, so every new commit in master triggers a build on the docker hub, in order to update the docker image as well, which makes maintenance easier.

Each update on the oce image triggers the pythonocc-core image, and each pythonocc-core image update triggers the cadquery image update.

So I'm going to create an issue on CadQuery's pythonocc-core as well, in order to add its Dockerfile.

I have also forked oce to add its Dockerfile, but this one still take too long time (for now) to be built on the docker hub, so the images must be built locally then pushed to the docker hub (without autobuild).

So pythonocc-core's auto-build listens to this oce fork and not to oce vanilla. That means that this fork must be maintained, unless the oce team is interested for a pull request that adds this Dockerfile.

By the meantime, it can be a good idea to add this oce fork to the cadquery organization, what do you think?

What to do next?

- [ ] make the container [CQGI-compliant](https://cadquery.readthedocs.io/en/latest/cqgi.html);
- [ ] allow input and output volumes to be mounted;
- [ ] merge;
- [ ] link docker-hub builds to cadquery org's repositories;
- [ ] update documentation.


jmwright:
> By the meantime, it can be a good idea to add this oce fork to the cadquery organization, what do you think?

I'm ok with it, but I'd be interested to hear what @adam-urbanczyk has to say about it.

roipoussiere:
> I really like the idea of a command line flow that allows running a script I side the container in one line. The trick to look for requirements.txt seems good.
>
> Once you are mounting directories, you might consider looking for /data/requirements. txt rather than in root. That would allow the use case in which I want to have a script that reads other files from the filesystrm and or writes output to the file system.
>
> As an example, suppose my script generates lots of step files as output. This is no longer good to stream to stdout. But if the script is run with /data as the working directory, I can throw requirements.txt there, and write output files there too.
[dcowden](https://github.com/CadQuery/cadquery/issues/177#issuecomment-521982578)

What about this:

```bash
docker run --rm --user $(id -u):$(id -g) -v <input volume>:/input:ro -v <output volume>:/output cadquery/cadquery [options]
```

Where:
- `--rm` just removes the container after the command is executed;
- `--user $(id -u):$(id -g)`: give user's rights to the container, so the generated files will be writable and removable by the host when the container will be removed;
- `-v <input volume>:/input:ro` mounts a folder containing user's cadquery scripts, in read-only mode in order to be sure that the container will not accidentally mess up its content. If this folder contains a `requirements.txt` file, dependencies will be installed;
- `-v <output volume>:/output ` mounts an other user folder (empty or not), where files will be generated;
- `[options]` are cadquery.sh script options:
    - `--input <input>` (or `-i`), where `<input>` is a pattern used to execute files relative to the *input* directory, like `main.py`, `{script1.py,script2.py}` or `*.py`, or `-` to  execute stdin as python script;
    - `--output <file_name>` (or `-o`), set output file name, which can also be `-` to outputs to stdout (only the last model if there are more than one call to `show_model()`). If not given, use the input file name (or just `out` if we `stdin` is used). If there are more than one model to generate, <file_name> will act as a name suffix;
    - `--format <format>` (or `-f`): define the file format to produce, where `<format>` can be one of `svg`, `stl` or `step`. If not given, try to guess from the output file name extension, or use `svg` as default`;
    - `--debug <file_name>` (or `-d`): path where `debug()` results will be saved (stdout by default).

Examples:

1. Read from stdin, write to stdout as svg, with `print`:
```bash
echo "print(cq.Workplane('XY').box(1,2,3).toSvg())" | docker run --rm cadquery/cadquery -i - -o - > my_model.svg
```

2. Read from stdin, write to stdout as stl, with `show_model()`:
```bash
echo "show_model(cq.Workplane('XY').box(1,2,3))" | docker run --rm cadquery/cadquery -i - -f stl -o - > my_model.stl
```

3. Read from a file, write to stdout as stl, with dependencies:
```bash
echo "show_model(cq.Workplane('XY').box(1,2,3))" > input/my_model.py
echo "asynico" > input/requirements.py
docker run --rm --user -v $(pwd)/input:/input:ro cadquery/cadquery -i my_model.py -o - > file.stl
```

4. Read from a file, write to a file, generating a file named `my_model.svg` (default extension):
```bash
echo "show_model(cq.Workplane('XY').box(1,2,3))" > input/my_model.py
docker run --rm --user $(id -u):$(id -g) -v $(pwd)/input:/input:ro -v $(pwd)/output:/output cadquery/cadquery -i my_model.py
```

5. Read from a file, write to a file, generating a file named `my_script.stl`:
```bash
echo "show_model(cq.Workplane('XY').box(1,2,3))" > input/my_model.py
docker run --rm --user $(id -u):$(id -g) -v $(pwd)/input:/input:ro -v $(pwd)/output:/output cadquery/cadquery -i my_model.py -f stl
```

6. Same thing (but use -o instead of -f to name the file), and also outputs debug entries in `debug.txt` :
```bash
echo "box=cq.Workplane('XY').box(1,2,3)\nshow_model(box)\ndebug(box, {'color': 'red'} )" > input/my_model.py
docker run --rm --user $(id -u):$(id -g) -v $(pwd)/input:/input:ro -v $(pwd)/output:/output cadquery/cadquery -i my_script.py -o my_model.stl -d debug.txt
```

7. Read from many files, write to a single file:
```bash
echo "def add(a, b)\n    return a+b" > input/utils.py
echo "from utils import add\nshow_model(cq.Workplane('XY').box(1,2,add(1+2)))" > input/my_model.py
docker run --rm --user -v $(pwd)/input:/input:ro cadquery/cadquery -i my_model.py -f stl -o - > my_model.stl
```

8. Read from many files, write to many files, generating `my_model1.stl` and `my_model2.stl`:
```bash
echo "show_model(cq.Workplane('XY').box(1,2,3))" > input/my_model1.py
echo "show_model(cq.Workplane('XY').box(2,3,4))" > input/my_model2.py
docker run --rm --user $(id -u):$(id -g) -v $(pwd)/input:/input:ro  -v $(pwd)/output:/output cadquery/cadquery -i *.py -f stl
```

9. Read from many files, write to many files, generating `my_model1_out.stl` and `my_model2_out.stl`:
```bash
echo "show_model(cq.Workplane('XY').box(1,2,3))" > input/my_model1.py
echo "show_model(cq.Workplane('XY').box(2,3,4))" > input/my_model2.py
docker run --rm --user $(id -u):$(id -g) -v $(pwd)/input:/input:ro  -v $(pwd)/output:/output cadquery/cadquery -i *.py -f stl -o out.stl
```

dcowden:
@roipoussiere the input, output, and format arguments you proposed are in line with what the old command line did. 

That said, when I saw your simpler proposal, all-in-all i liked yours more. Simple but flexible i think wins here since after all we're talking about someone who's certainly already a developer.  

If you are developing your own script already, you are probably perfectly willing to write the little bit of code to choose output format, filenames, and all that. And the extra flexibility is huge. For example, what if you generate both step and stl? or svg and step?

Similarly, having both input and output mapping is slightly more complex and slightly more constraining, in comparison with simply mounting a folder and letting the script do whatever it wants within that folder.  

To better support users with a complex script, it might make sense to support two ways of working:

(1) , mount /data, which optionally has requirements.txt, if present, load it.

(2) seperately, allow --input <value> .  if <value> is 'stdin', read stdin as a script. otehrwise, assume it is a filename to execute that's located in /data

roipoussiere:
> That said, when I saw your simpler proposal, all-in-all i liked yours more.

Which syntax are you talking about, the one suggested here with all these options, or [the initial one](https://github.com/CadQuery/cadquery/issues/169#issue-477928881)?

In this last case, how do you generate .stl or .step from your script? I though that step was supposed to be executed by the cqgi.

Dustyposa:
Oh, that's wonderful.I never thinking about make docker images like a shell.It's so convenient.
Has a way can don't use line cmd to run this image?
And I think can improve some tag to reduce the image size. ie. don't install CQ-editor.
If docker line cmd will cost less time than run code in container. I will fall in love with it!!!:)

dcowden:
> Which syntax are you talking about, the one suggested here with all these options, or the initial one?
The initial one. it struck me as pretty simple.

> In this last case, how do you generate .stl or .step from your script? I though that step was supposed to be executed by the cqgi.
Ah! that's a very good point!  I didn't have this consideration in mind, but I"m glad you did!
Yes, there is the case where we're running a cqgi-compliant script, and then separately there's the case where the developer can't live with those constraints, and is totally 'off the reservation'.  

The cqgi flow is best when things are simple, but there are cases where that interface is simply too constraining.  As one example, the KiCAD guys that are building thousands of files from one back run. Though I suppose we could argue that people doing that are probably not going to be trying to use a pre-packaged docker file!

So maybe you're indeed right that the best move is to have the docker file only able to execute cqgi-complaint scripts. in that case, then the output format would be important. though its still possible for a cqgi-compliant script to yield more than one object

dcowden:
@Dustyposa regarding image size, i think there will eventually need to be several flavors.  There's a huge difference between mini-conda and the full anaconda. 

Some uses may prefer downloading a large image once, but then know they have all the usual libraries at their disposal, while others may favor a very light image with manual installation of dependencies.  It's certainly reasonable to package a different image for use to run CQ-editor ( as a gui app) vs command line builds

> Has a way can don't use line cmd to run this image?

The most straightforward approach I can come up with would be to have a caqdquery:cq-editor image tag, which is designed run interactively, while exporting the Xdisplay so that it can display a GUI. 

Another strategy that's more work but substantially cooler would be to have the cadquery image be able to expose a REST endpoint that can allow building models, coupled with an embedded webserver that serves up an electron webapp for local use.  This would have the added benefit that we could use this image to allow people to run their own cloud cadquery servers with minimal effort

Dustyposa:
@Dustyposa The another way I'm doing. I use Starlette to expose a RSET endpoint. receive cadquery code, then response the stdout.
 The cadquery must depend on conda? I use miniconda run `conda install -c cadquery cadquery=2`, this image only have 2.33G.But I can't `import cadquery`, `traceback Miss visual_...`
 If can build by python. The image size will be smaller.

roipoussiere:
For your information, the docker images i have created compile oce, pythonocc-core and and cadquery from sources. There is no [mini|ana]Conda inside, I don't how to make them lighter.

Also, about cq-editor, I tried to create a flatpak, but I still get some errors (see related PR, if you can give some help it will be useful). I don't know how to create a docker image that interacts with a x-server, do you? Maybe I can also try to create a snap package by the meantime.

I like the idea of a web server with a rest api, I would love to work on it.

Also, is there a way to export to stl or step *from* the script? In this case it will be easier to build the docker image.

dcowden:
@roipoussiere 

> Also, is there a way to export to stl or step from the script? In this case it will be easier to build the docker image.

If I understand this question correctly, yes.  all code paths for exporting objects end up calling https://github.com/CadQuery/cadquery/blob/master/cadquery/occ_impl/exporters.py#L49

cqgi eventually calls this method to render the objects.

I do not understand how that lightens the dockerfile size, but I'm sure you know what you're talking about there.

dcowden:
@roipoussiere what is the reason that you started the cadquery image stack from 

nvidia/cudagl:10.1-base 

I've never heard of that one.  I'm guessing that it is suitable due to video drivers of some kind.

Could we produce a set of images that start with stretch: slim or similar, which would be command line only but very small?  I'll bet if you compile from source, and start with stretch slim, you could get under 500MB easily.

roipoussiere:
I started with nvidia image to be able to use OpenGL but maybe there is a better solution.

roipoussiere:
> If I understand this question correctly, yes. all code paths for exporting objects end up calling https://github.com/CadQuery/cadquery/blob/master/cadquery/occ_impl/exporters.py#L49

Ok! Why don't implement shape.export(...) then ? In this way users could easily use it in the docker image.

dcowden:
perhaps something along these lines:

```
FROM debian:slim-stretch
RUN apt-get update && apt-get install libgl1-mesa-dev libglu1-mesa-dev, freeglut3-dev

```

Dustyposa:
@roipoussiere compiled after， the source is deleted， and relative compiled can be remove .
About the rest api. Now I can get the stl file content by stdout. The step I hadn't a try.

Dustyposa:
I make a server to interpreter cadquery code. But every request will import cadquery, a little slow. Has any way to speed up import process?

Shimwell:
Hi all, I've been looking for a cadquery dockerfile and had a quick go at making one and found all the threads very useful. 

```
FROM ubuntu:18.04

# Updating Ubuntu packages
RUN apt-get update && yes|apt-get upgrade

# Adding wget and bzip2
RUN apt-get install -y wget bzip2

# Anaconda installing
RUN wget https://repo.continuum.io/archive/Anaconda3-2020.02-Linux-x86_64.sh

RUN bash Anaconda3-2020.02-Linux-x86_64.sh -b

RUN rm Anaconda3-2020.02-Linux-x86_64.sh

# Set path to conda
ENV PATH /root/anaconda3/bin:$PATH

# Configuring access to Jupyter
RUN mkdir /opt/notebooks
RUN jupyter notebook --generate-config --allow-root
RUN echo "c.NotebookApp.password = u'sha1:6a3f528eec40:6e896b6e4828f525a6e20e5411cd1c8075d68619'" >> /root/.jupyter/jupyter_notebook_config.py

# # Jupyter listens port: 8888
EXPOSE 8888

RUN conda install -c conda-forge -c cadquery cadquery=2

RUN apt-get update
RUN apt-get install -y libgl1-mesa-dev 
RUN apt-get install -y libglu1-mesa-dev
RUN apt-get install -y freeglut3-dev
```

Naturally those last 4 commands can be combined if you prefer.

roipoussiere:
By the way I'm working on a [flatpak](flatpak.org) for cadquery/cq-editor. The installation process will be fast and simple. I will create a pull-request when ready. :)

jmwright:
@roipoussiere Sounds great, thanks for working on that.

@Shimwell Thanks for posting your sample dockerfile.

roipoussiere:
> By the way I'm working on a flatpak for cadquery/cq-editor. The installation process will be fast and simple. I will create a pull-request when ready. :)

And that also means that I will probably not work on this PR since it does not really make sense for me because a flatpak will be easier to use than a Dockerfile.

So what should I do? Close the PR or un-draft it in order to let someone else merge it?

Shimwell:
I think you will always have some people who want different installation options, pip, conda, flatpack, apt-get, docker etc. Naturally it is hard to satisfy and maintain every option. Docker is quite handy for cloud deployment, k8 and Google cloud run bit people can also make there own docker images that use the flatpak

jmwright:
I needed to create a CadQuery Dockerfile for something else, so I extracted the relevant parts and put it [here](https://github.com/jmwright/cadquery-docker). The Dockerfile uses the latest master of cadquery with Miniconda3 because CQ development has been moving quickly and many of us are always wanting the latest features and fixes. The image ends up being about 1.4 GB in size. 

To use the Dockerfile, clone the repo, cd into it, and run
```
docker build -t cadquery ./
```
After that you should be able to run something like
```
docker run [image ID] python -c "import cadquery"
```

I typically only use Docker indirectly, so I'll need some feedback from the community. I have a few questions for everyone.

1. The CadQuery 1.x [Dockerfile](https://github.com/dcowden/cadquery/blob/master/Dockerfile#L54) made use of a [Python](https://github.com/dcowden/cadquery/blob/master/cq_cmd.py)/[shell](https://github.com/dcowden/cadquery/blob/master/cq_cmd.sh) script pair so that you could use the image like a command line utility that took in CQ script text and output an exported format (STL, STEP, etc). Should this be carried forward into this new Docker setup, or does it not fit the typical use case?
2. Should this Dockerfile include skeleton comments for mounting a host directory?
3. Where should this Dockerfile and the associated config/artifacts live? I'm inclined for them to stay in a separate repo since Docker is not needed for CQ to operate and would clutter up the CQ repo a bit. I can transfer this repo to the CadQuery org if/when it's deemed useful.

@dcowden @adam-urbanczyk

Issue #158, opened 2019-06-25
Implement rounded slot feature


adam-urbanczyk:
![obraz](https://user-images.githubusercontent.com/13981538/60075055-b7d02880-9724-11e9-9d90-bd40418fcc8a.png)

bweissinger:
What kind of slot commands should we implement? Commonly used slots seem to be:

- 	Center to Center
- 	Center Point
- 	Three Point Arc Slot
- 	Center Point Arc Slot

A 'Slot Along Spline' type may be useful as well.

Also, should these features be implemented as 2d sketch objects or as 3d operations? i.e. should they behave like circle() or hole()?

adam-urbanczyk:
I'd propose to have `center`, `length` and `dir` as inputs to follow current conventions. My preference would go to a 2D sketch.

bweissinger:
Agreed on both counts. I think it would be an excellent addition. 

I think the ability to use existing sketch geometry to slot along would cover any advanced uses of slots, and be a good future addition. I imagine that would be closely related to offsetting sketch geometry, which would also be a good feature to have.

dcowden:
@adam-urbanczyk @bsilvereagle I know this is off topic somewhat, but I've been desperate for many years ( since about the first month of CQ development actually ) to integrate parametric sketches.  I was stuck for a long time because to do that you need a constraint solver.    CQ does a good job considering design intent when going from 2d to 3d, but when constructing sketches you can't convey intent as easily as you can when you draw in a CAD tool.  IE: I want a slot, and i want it _centered_ with respect to these two sides.

adam-urbanczyk:
Solved by #186

PR #186, opened 2019-09-04
Add rounded slot feature

bweissinger:
Adds basic functionality for creating slots. It uses the current point as the center point for the slot. Length is the overall end to end length of the slot, with direction set by angle in degrees. Addresses #158

### DIFF
--- a/cadquery/cq.py
+++ b/cadquery/cq.py
@@ -1346,7 +1346,7 @@ class Workplane(CQ):
         newCenter = p + Vector(xDist, yDist, 0)
         return self.newObject([self.plane.toWorldCoords(newCenter)])
 
-    def slot(self, length, diameter, angle):
+    def slot(self, length, diameter, angle=0):
         """
         Creates a rounded slot for each point on the stack.
 
--- a/cadquery/cq.py
+++ b/cadquery/cq.py
@@ -1370,12 +1370,12 @@ class Workplane(CQ):
             radius = diameter/2
             centerPoint = self.plane.toWorldCoords(pnt)
 
-            p1 = centerPoint.add(Vector((-length/2) + radius, diameter/2))
-            p2 = p1.add(Vector(length - diameter, 0))
-            p3 = p1.add(Vector(length - diameter, -diameter))
-            p4 = p1.add(Vector(0, -diameter))
-            arc1 = p2.add(Vector(radius, -radius))
-            arc2 = p4.add(Vector(-radius, radius))
+            p1 = centerPoint + Vector((-length/2) + radius, diameter/2)
+            p2 = p1 + Vector(length - diameter, 0)
+            p3 = p1 + Vector(length - diameter, -diameter)
+            p4 = p1 + Vector(0, -diameter)
+            arc1 = p2 + Vector(radius, -radius)
+            arc2 = p4 + Vector(-radius, radius)
 
             edges=[(Edge.makeLine(p1,p2))]
             edges.append(Edge.makeThreePointArc(p2, arc1, p3))
@@ -1384,7 +1384,7 @@ class Workplane(CQ):
 
             slot = Wire.assembleEdges(edges)
 
-            return slot.rotate(centerPoint, centerPoint.add(Vector(0,0,1)), angle)
+            return slot.rotate(centerPoint, centerPoint + Vector(0,0,1), angle)
 
         return self.eachpoint(_makeslot, True)
 
--- a/cadquery/cq.py
+++ b/cadquery/cq.py
@@ -1346,7 +1346,7 @@ class Workplane(CQ):
         newCenter = p + Vector(xDist, yDist, 0)
         return self.newObject([self.plane.toWorldCoords(newCenter)])
 
-    def slot(self, length, diameter, angle=0):
+    def slot2D(self, length, diameter, angle=0):
         """
         Creates a rounded slot for each point on the stack.
 
--- a/cadquery/cq.py
+++ b/cadquery/cq.py
@@ -1368,9 +1368,8 @@ class Workplane(CQ):
             """
 
             radius = diameter/2
-            centerPoint = self.plane.toWorldCoords(pnt)
 
-            p1 = centerPoint + Vector((-length/2) + radius, diameter/2)
+            p1 = pnt + Vector((-length/2) + radius, diameter/2)
             p2 = p1 + Vector(length - diameter, 0)
             p3 = p1 + Vector(length - diameter, -diameter)
             p4 = p1 + Vector(0, -diameter)
@@ -1384,7 +1383,7 @@ class Workplane(CQ):
 
             slot = Wire.assembleEdges(edges)
 
-            return slot.rotate(centerPoint, centerPoint + Vector(0,0,1), angle)
+            return slot.rotate(pnt, pnt + Vector(0,0,1), angle)
 
         return self.eachpoint(_makeslot, True)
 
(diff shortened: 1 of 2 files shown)


cadquery/cq.py
```diff
@@ -1346,6 +1346,48 @@ def move(self, xDist=0, yDist=0):
...
+            """
+            Inner function that is used to create a slot for each point/object on the workplane
+            :param pnt: The center point for the slot
+            :return: A CQ object representing a slot
+            """
+
+            radius = diameter/2
+            centerPoint = self.plane.toWorldCoords(pnt)
+
+            p1 = centerPoint.add(Vector((-length/2) + radius, diameter/2))
```
adam-urbanczyk: Why not using `+` operator?

cadquery/cq.py
```diff
@@ -1346,6 +1346,48 @@ def move(self, xDist=0, yDist=0):
...
+            arc2 = p4.add(Vector(-radius, radius))
+
+            edges=[(Edge.makeLine(p1,p2))]
+            edges.append(Edge.makeThreePointArc(p2, arc1, p3))
+            edges.append(Edge.makeLine(p3, p4))
+            edges.append(Edge.makeThreePointArc(p4, arc2, p1))
+
+            slot = Wire.assembleEdges(edges)
+
+            return slot.rotate(centerPoint, centerPoint.add(Vector(0,0,1)), angle)
```
adam-urbanczyk: Is `centerPoint.add(Vector(0,0,1))` needed? I thought `Vector(0,0,1)` would be enough (though did not check it).

adam-urbanczyk:
Thanks a lot @bweissinger ! Some minor comments in the review.

cadquery/cq.py
bweissinger: As in 
```
p1 = centerPoint + Vector((-length/2) + radius, diameter/2)
```
Honestly, I saw .add() used elsewhere so I just assumed `__add__` wasn't defined. I think changing to the `+` operator would be a good improvement in readability.

cadquery/cq.py
bweissinger: I tried `Vector(0,0,1)` first, also assuming that was all that was needed. However, that only works with the slot centered at `(0, 0, 0)`.

tests/TestCadQuery.py
```diff
@@ -2059,3 +2059,17 @@ def testFindSolid(self):
         s = r.findSolid()
         self.assertEqual(len(s.Solids()),2)
         self.assertTrue(isinstance(s,Compound))
+
+    def testSlot(self):
+
+        decimal_places = 9
+
+        #Ensure it produces a solid with the correct volume
+        result = Workplane("XY").slot(4,1,0).extrude(1)
```
jmwright: Might be worth adding [cutBlind](https://cadquery.readthedocs.io/en/latest/classreference.html?highlight=cutblind#cadquery.Workplane.cutBlind) and/or [cutThruAll](https://cadquery.readthedocs.io/en/latest/classreference.html?highlight=cutthruall#cadquery.Workplane.cutThruAll) tests since I would expect that's normally how slots would be used, and on rare occasions extrude suceeds when a cut doesn't.

bweissinger: That is a good point, and cutBlind() actually does reveal a bug.
```
result = cq.Workplane("XY").box(5,5,1).faces(">Z").workplane() \
             .slot(4,1,0).cutBlind(-0.5)
```
The above code does not produce the expected result. For some reason, the slot is offset by 0.5 above the surface of the box. Further testing shows that it is always offset above the top of the box by half the box height.

bweissinger: Ah, it is caused by the cast of pnt to world coordinates.

bweissinger:
@adam-urbanczyk @jmwright Added changes to address noted issues.

jmwright:
@bweissinger Looks good. Thanks for implementing this.

cadquery/cq.py
adam-urbanczyk: Indeed, it is implemented a start and end and not start and dir.

cadquery/cq.py
adam-urbanczyk: It was added quite recently.

adam-urbanczyk:
Thanks for you work @bweissinger . Before we merge it, I think we need to still discuss the naming. I realized that `slot` could imply 3D feature. Do you know how other CAD packages name such a 2D thing?

cadquery/cq.py
```diff
@@ -1346,6 +1346,48 @@ def move(self, xDist=0, yDist=0):
         newCenter = p + Vector(xDist, yDist, 0)
         return self.newObject([self.plane.toWorldCoords(newCenter)])
 
+    def slot(self, length, diameter, angle):
```
adam-urbanczyk: One last technical proposal: could you default angle to zero? I think it will be natural for the users and will allow for nicer syntax when we implement something like `polarArray`

`def slot(self, length, diameter, angle=0):`

bweissinger:
> Thanks for you work @bweissinger . Before we merge it, I think we need to still discuss the naming. I realized that `slot` could imply 3D feature. Do you know how other CAD packages name such a 2D thing?

Good catch. 

As far as actual function names, the only package I have any idea on would be Freecad. Looking at the Freecad console, the slot sketch command seems to be named `Sketcher_CreateSlot`.

In my experience, slot _usually_ refers to a 2D sketch feature. If a CAD package does use 3D slots, they don't seem to have a 2D slot features. A notable exception is Solid Works. It has both, but the 3D feature is actually located under `Hole Wizard`. I think it just depends on the function/industry the software is oriented towards. For instance, Catia has (as far as I know) only a 3D slot function, and it actually uses a path and a profile to cut along that path. 

Unfortunately, the above programs all have a GUI to provide context. They could have both 2D and 3D, and name them both `slot` without much issue. However, 2D sketch slots tend to be named after the way they are defined:
  - `Center to Center Slot`
  - `Center Point Slot`
  - `Arc Slot`
  - ...

With everything above in mind, a couple of possibilities includes:
 - `sketchSlot`
 - `centerPointSlot`

cadquery/cq.py
bweissinger: I like that. That would be a good addition.

adam-urbanczyk:
@jmwright any thoughts on the name?

jmwright:
@adam-urbanczyk I'm not sure. `sketchSlot` would be a break from how we name 2D operations, but it's probably the most accurate, and I think this is a little bit of a different case than we've had before with 2D operations. If it's named `sketchSlot` or something like `slot2D`, the user could still call some sort of 3D `*Slot` operation later that creates the 3D version, cut from the selected face.

Hatatister:
@jmwright I think in AutoCad there is a similar feature called "RoundedRect", but where there is one more degree of freedom namely the height of the rect. Our "slot" would be a special case of a rounded rect, where H = 2 * R. Maybe be it makes sense to generalize our slot to a rounded  rect, which can be parameterized by width, height, edge_radius and where height defaults to 2 times egde_radius when not given by the user.

adam-urbanczyk:
I looked at the newest FreeCAD and in the sketcher UI I can see a `Create Slot` tool. One day we might want to expose the OCC 3D slot feature so I'm in favor of renaming this to `slot2D`.

@Hatatister you can create a rounded rectangle by using `rect`, extruding it and filleting the edges. This is not possible for a slot (due to OCC kernel limitations).

bweissinger:
Slot2D sounds good to me as well.

adam-urbanczyk:
I think it is ready to be merged. Anything to add @jmwright @dcowden ?

dcowden:
@adam-urbanczyk looks great to me!

jmwright:
+1

adam-urbanczyk:
Thanks for your contribution @bweissinger !

PR #200, opened 2019-10-15
Spline example fix

adam-urbanczyk:
This will fix one broken example in the documentation


adam-urbanczyk:
Simple doc fix, so merging as-is.

PR #215, opened 2019-11-21
Documentation corrections, fixed #213

### LINKED ISSUE #213: Minor documentation issue
In the ["Quick Start" documentation](https://cadquery.readthedocs.io/en/latest/quickstart.html), I've noticed a very minor mistake.  It's not a big deal, but it confused me for a while when I was trying to work out how CQ worked.

Under ["More Holes"](https://cadquery.readthedocs.io/en/latest/quickstart.html#more-holes) it says "The centers of these holes should be 12mm from the edges of the block." but the code draws the centres of the holes 6 mm from the edge.  As described in the explanation:

> Line 11 draws a rectangle 12mm smaller than the overall length and width of the block, which we will use to locate the corner holes. We’ll use the vertices ( corners ) of this rectangle to locate the holes. The rectangle’s center is at the center of the workplane, which in this case coincides with the center of the bearing hole.

**If the rectangle is 12 mm smaller than the length, then the holes are 6 mm from the edges**, as the "padding" is split between both edges.

Oh, and in [this section](https://cadquery.readthedocs.io/en/latest/designprinciples.html#principle-2-capture-design-intent), "affect" should be "effect".

Renha:


jmwright:
@Renha Thanks for the contribution.

@abudden Thanks for pointing these errors out.

PR #236, opened 2019-12-12
Use pytest to run the test suite

Peque:
While creating the test for https://github.com/CadQuery/cadquery/pull/235, I missed a couple of things about `pytest`:

- I had to add extra indentation (tests needed to be inside a class, so extra "boilerplate")
- Output of the test suite was not very pretty

But specially, I had to write:

```python
self.assertAlmostEqual(rotated[0], normal[0])
self.assertAlmostEqual(rotated[1], normal[1])
self.assertAlmostEqual(rotated[2], normal[2])
```

Instead of (`from pytest import approx`):

```python
assert rotated == approx(normal)
```

The latter is definitely shorter, clearer and more Pythonic. :blush:

Also, I think nowadays `pytest` is the *de facto* tool. Or that is my impression.

Hope you like the shift.

PR #237, opened 2019-12-13
Update assembledEdges in shapes.py

bragostin:
```BRepBuilderAPI_MakeWire::Add(const TopTools_ListOfShape & L)``` offers the option to accept a list of shapes directly as argument:
```Adds the edges of <L> to the current wire. The edges are not to be consecutive. But they are to be all connected geometrically or topologically. If some of them are not connected the Status give DisconnectedWire but the "Maker" is Done() and you can get the partial result. (ie connected to the first edgeof the list <L>)```
Following this I modified the ```assembledEdges``` in ```shapes.py``` to be able to provide a list of unordered edges to ```BRepBuilderAPI_MakeWire```. This way, when the list of edges is generated by another function, there is not need to make them consecutive.

### CODE REVIEW

**Comment by jmwright** [comment]
@bragostin @adam-urbanczyk What would you guys think about printing a warning when the status is DisconnectedWire? Is it practical to do that?

> If some of them are not connected the Status give DisconnectedWire but the "Maker" is Done()

Users often have trouble when they're using a set of points and a joint isn't connected or the wire isn't closed. It would be nice to have some indication that they need to look at their points again unless they meant for things to be disconnected.

**Comment by jmwright** [comment]
I'm not saying it needs to happen in this PR, but looking ahead, implementing warnings like that would be nice for users.

**Comment by jmwright** [comment]
I'm not saying it needs to happen in this PR, but looking ahead, implementing warnings like that would be nice for users.

**Comment by bragostin** [comment]
@jmwright @adam-urbanczyk would a simple ```print('BRepBuilderAPI_MakeWire() Status = ', wire_builder.IsDone(), wire_builder.Error())``` be enough?
As I understand the documentation, if the wires are not connected it will not throw an exception but return a partial result.

**Comment by jmwright** [comment]
@bragostin That sounds good to me. It's just something to keep the user from having to guess what's going on if things don't work as expected.

**Comment by adam-urbanczyk** [comment]
@bragostin looks good but I'd propose to use `warnings.warn` i.s.o. `print`

**Comment by bragostin** [comment]
@jmwright @adam-urbanczyk  I am just discovering the module ```warnings.warn```. Would something like 
```python
if not wire_builder.IsDone():
    w1 = 'BRepBuilderAPI_MakeWire::IsDone(): returns the construction status. BRepBuilderAPI_WireDone if the wire is built, or another value of the BRepBuilderAPI_WireError enumeration indicating why the construction failed = '+ str(wire_builder.IsDone())
    w2 = 'BRepBuilderAPI_MakeWire::Error(): returns true if this algorithm contains a valid wire. IsDone returns false if: there are no edges in the wire, or the last edge which you tried to add was not connectable = ' + str(wire_builder.Error())
    warnings.warn(w1)
    warnings.warn(w2)
```
be ok?

jmwright:
@bragostin I think that looks fine.

bragostin:
I am not familiar with using github yet and I am obviously doing something wrong here, any idea what it is?

adam-urbanczyk:
What do you mean @bragostin ?

bragostin:
Well, 2 checks were not successful:
FAILED: codecov/patch — 60% of diff hit (target 94.86%)
FAILED: codecov/project — 94.76% (-0.11%) compared to acf363d
PASSED: continuous-integration/appveyor/pr — AppVeyor build succeeded
PASSED: continuous-integration/travis-ci/pr — The Travis CI build passed

adam-urbanczyk:
Yes, indeed: codecov is complaining about test coverage decrease. If you feel like you could add a test that triggers the newly implemented warning.

bragostin:
@adam-urbanczyk I just did add a test that triggers the warning and the 4 checks are ok now.

bragostin:
I have improved the test ```TestAssembleEdgesWarnings.py``` so that it returns only a warning and no error.

jmwright:
@adam-urbanczyk Do you want me to go ahead and merge this?

bragostin:
Sorry I did a mess, I forgot to create a new branch for ```BRepOffsetAPI_MakeFilling```, I put things back in order.

cadquery/occ_impl/shapes.py
```diff
@@ -809,12 +810,20 @@ def assembleEdges(cls, listOfEdges):
         """
             Attempts to build a wire that consists of the edges in the provided list
             :param cls:
-            :param listOfEdges: a list of Edge objects
+            :param listOfEdges: a list of Edge objects. The edges are not to be consecutive.
```
jmwright: Does "The edges are not to be consecutive" here mean that the call will fail if they are consecutive, or is it just saying that they don't have to be consecutive?

bragostin: @jmwright 3 out of the 5 examples I made for ```BRepOffsetAPI_MakeFilling``` #253 have non consecutive edges, so I guess it does not matter.

bragostin: Do you want me to add one non consecutive edges case to the tests?

jmwright: Yes, please. Based on all the bugs we've fought over the years with edge ordering I think it's worth it.

bragostin: @jmwright I have added 3 more tests in TestsAssembleEdges.py with consecutive and non consecutive edges. I modified runtests.py to check that it is running, but now I have one conflict reported. Maybe I should not have included this test in runtests.py?

jmwright:
@bragostin The reason you're getting a conflict is because runtest.py was deleted recently as part of the upgrade to pytest. https://github.com/CadQuery/cadquery/pull/236

bragostin:
@jmwright In the test_cadquery.py file I have added a ```test_assembleEdges``` function with 4 tests including consecutive and non consecutive edges. All checks have passed.

jmwright:
Thanks. Sorry you didn't get a heads-up about the test changes. Changes have been happening quickly and it's hard to keep track.

cadquery/occ_impl/shapes.py
```diff
@@ -809,12 +810,20 @@ def assembleEdges(cls, listOfEdges):
...
         """
+        
         wire_builder = BRepBuilderAPI_MakeWire()
-        for edge in listOfEdges:
-            wire_builder.Add(edge.wrapped)
+        edges_list = TopTools_ListOfShape()
+        for e in listOfEdges:
+            edges_list.Append(e.wrapped) 
+        wire_builder.Add(edges_list)
+        if wire_builder.Error():
```
adam-urbanczyk: Shouldn't this actually be:
```python
if not wire_builder.IsDone():
```
given that `IsDone` returns a bool and `Error()` returns an enum? I'm basing my question on the OCC docs: https://www.opencascade.com/doc/occt-6.9.0/refman/html/class_b_rep_builder_a_p_i___make_wire.html#a4e60ad05c4fd315e7b8304b8687b9fa3 but I'm not sure if I get why there are two similar methods.

adam-urbanczyk:
@bragostin I think it looks good. Just take a look a the single comment I had.

cadquery/occ_impl/shapes.py
bragostin: Good point! 
The original motivation to have these warnings in the new ```assembleEdges``` was this comment in the OCC docs:
```If some of them are not connected the Status give DisconnectedWire but the "Maker" is Done() and you can get the partial result.```
The warning is supposed to inform the user if the wire is done while he does not get the expected results. That's why ```Error()``` is used rather than ```IsDone()```. Actually I could not come with any example where ```IsDone()``` returns false...
However you are correct that ```if wire_builder.Error():``` is not adequate as ```Error()``` returns a number. I found that the returned BRepBuilderAPI_MakeWire::Error() values are
    BRepBuilderAPI_WireDone = 0
    BRepBuilderAPI_EmptyWire = 1
    BRepBuilderAPI_DisconnectedWire = 2
    BRepBuilderAPI_NonManifoldWire = 3
Therefore I have now replaced ```if wire_builder.Error():``` with ```if wire_builder.Error()!=0:``` and modified the first test to pass an open wire to ```assembleEdges``` so that ```IsDone``` is still true but ```Error``` returns 2, so that the warning functionality is tested. I have also added the signification of the 4 values in the docstring.
What do you think?

jmwright: Looks good to me.

jmwright:
@bragostin It looks like you addressed the comment from @adam-urbanczyk . Do you feel this PR is ready to merge? I looked it over again and I can't think of anything else that we're missing.

bragostin:
@jmwright I have been using this new ```assembleEdges``` function for 2 weeks now because it is a pre-requisite for the ```BRepOffsetAPI_MakeFilling implementation``` and I have not noticed any issue so far, and older models are still running correctly. So from my side it looks ok.

jmwright:
@bragostin Thanks for all your work on this. Looks like it's time to merge it.

When I get a chance I'll take a look at some of the other edge assembly related issues here in GitHub to see if this fixes them.

adam-urbanczyk:
Thanks for the contribution @bragostin !

PR #253, opened 2019-12-21
Implementation of BRepOffsetAPI_MakeFilling

bragostin:
This is an implementation of the OCC function ```BRepOffsetAPI_MakeFilling``` into CadQuery in order to produce a plate surface with a thickness (optional), enclosed by edge points, polylines or splines, and going through interpolation points (optional) .


bragostin:
Something strange. I just added asserts in the test, and now get this error:
```
================================== FAILURES ===================================
________________________ TestCadQuery.test_interpPlate ________________________
self = <tests.test_cadquery.TestCadQuery testMethod=test_interpPlate>
    def test_interpPlate(self):
    
        decimal_places = 9
    
        # example from PythonOCC core_geometry_geomplate.py, use of thickness = 0 returns 2D surface.
        thickness = 0
        edge_points = [[0.,0.,0.], [0.,10.,0.], [0.,10.,10.], [0.,0.,10.]]
        surface_points = [[5.,5.,5.]]
        plate_0 = Workplane('XY').interpPlate(edge_points, surface_points, thickness)
        self.assertTrue(plate_0.val().isValid())
>       self.assertAlmostEqual(plate_0.val().Area(), 141.218823892, decimal_places)
E       AssertionError: 141.23402758675547 != 141.218823892 within 9 places
tests\test_cadquery.py:2198: AssertionError
============================== warnings summary ===============================
cadquery\selectors.py:198
```
However locally all the tests pass. Any idea? Should I just reduce the number of decimal places?

jmwright:
It seems odd to me that the `Area` method is returning a mass from the OCC properties.

https://github.com/CadQuery/cadquery/blob/3cd466a693ce1d7f084e5cd6ddee256fc28d020a/cadquery/occ_impl/shapes.py#L470

@bragostin Have you figured out the area of that plate any other way to make sure it's correct, or have you relied on CQ to give you the correct value? I'm wondering if the `Area` call is even giving back the correct area at this point.

bragostin:
@jmwright so, this is what I get now:
- Area() from local CQ under linux = 141.218823892
- Area() from CI under Window and Mac = 141.234027586
- Area() from CI under Linux = 141.218823892 --> same as local
- Volume() from local CQ under linux = 141.218823892 --> same as Area()
- Volume() with FreeCAD = 174.174510736 --> totally different

The second test in ```test_assembleEdges``` also fails, but after the 4th decimal only.

jmwright:
It looks like the `Mass` defaults to the area if no density is set or only surfaces are selected.

https://www.opencascade.com/doc/occt-7.4.0/refman/html/class_g_prop___g_props.html#a1796c83e8b5c53b959e41c55e87112a5

I tested the Area with one face selected of a 1x1x1 cube and got 0.9999999999999998

jmwright:
I can imagine there's always going to be some fuzziness with more complex surfaces like you're working with here. Maybe we're really just seeing the rounding error stack-up between platforms. It's a very small error percentage, so I guess my vote would be to reduce the number of decimal places for the assert. I'll be curious what @adam-urbanczyk thinks though.

bragostin:
The 5 tests pass only if I compare the volumes with 0 decimal places.
The OCC docs of ``BRepGProp::VolumeProperties`` state:

``
 Updates <VProps> with the shape <S>, that contains its pricipal properties. The volume properties of all the FORWARD and REVERSED faces in <S> are computed. If OnlyClosed is True then computed faces must belong to closed Shells. Adaptive 2D Gauss integration is used. Parameter Eps sets maximal relative error of computed mass (volume) for each face. Error is calculated as Abs((M(i+1)-M(i))/M(i+1)), M(i+1) and M(i) are values for two successive steps of adaptive integration. Method returns estimation of relative error reached for whole shape. WARNING: if Eps > 0.001 algorithm performs non-adaptive integration.
``

The default value of ``Eps`` seems to be 0.001, so the best accuracy one can hope for should be 0.1%, and 0.1/140 = 0.07%, so I guess 0 decimal places sounds about right.

cadquery/occ_impl/shapes.py
```diff
@@ -1150,6 +1199,67 @@ class Solid(Shape, Mixin3D):
     a single solid
     """
     
+    @classmethod
+    def interpPlate(cls, surf_edges, surf_pts, thickness, Degree=3, NbPtsOnCur=15, NbIter=2, Anisotropie=False, Tol2d=0.00001, Tol3d=0.0001, TolAng=0.01, TolCurv=0.1, MaxDeg=8, MaxSegments=9):
```
adam-urbanczyk: typo: Aniosotropie -> Anisotropy. BTW: why are the arg names capitalized?

cadquery/occ_impl/shapes.py
```diff
@@ -1150,6 +1199,67 @@ class Solid(Shape, Mixin3D):
...
+            w = wire_builder.Wire()
+        else: # Closed wires provided
+            w = surf_edges
+        edges = [i for i in TopologyExplorer(w).edges()]
+        
+        # MAKE SURFACE
+        continuity = GeomAbs_C0 # Fixed, changing to anything else crashes.
+        face = Face.makeNSidedSurface(edges, pts_array, continuity, Degree, NbPtsOnCur, NbIter, Anisotropie, Tol2d, Tol3d, TolAng, TolCurv, MaxDeg, MaxSegments) 
+        
+        if abs(thickness)>0: # Thicken surface, abs() because negative values are allowed (sets direction of thickening)
```
adam-urbanczyk: Is this really needed? Maybe an error (even from OCC) would be better?

cadquery/occ_impl/shapes.py
```diff
@@ -969,6 +977,47 @@ def innerWires(self):
         
         return [w for w in self.Wires() if not w.isSame(outer)]
 
+    @classmethod
+    def makeNSidedSurface(cls, edges, points, continuity=GeomAbs_C0, Degree=3, NbPtsOnCur=15, NbIter=2, Anisotropie=False, Tol2d=0.00001, Tol3d=0.0001, TolAng=0.01, TolCurv=0.1, MaxDeg=8, MaxSegments=9):
```
adam-urbanczyk: Same remark a below - maybe good to check the argument naming convention wrt rest of the CQ codebase

cadquery/cq.py
```diff
@@ -2728,6 +2728,66 @@ def _sweep(self, path, multisection=False, makeSolid=True, isFrenet=False,
 
         return Compound.makeCompound(toFuse)
 
+    def interpPlate(self, surf_edges, surf_pts=[], thickness=0, combine=True, clean=True, Degree=3, NbPtsOnCur=15, NbIter=2, Anisotropie=False, Tol2d=0.00001, Tol3d=0.0001, TolAng=0.01, TolCurv=0.1, MaxDeg=8, MaxSegments=9):
+        """
+        Returns a plate surface that is 'thickness' thick, enclosed by 'surf_edge_pts' points,  and going through 'surf_pts' points.
+        
+        :param surf_edges
+        :type 1 surf_edges: list of [x,y,z] float ordered coordinates
```
adam-urbanczyk: I'm not sure I I understand the intent here

cadquery/cq.py
```diff
@@ -2728,6 +2728,66 @@ def _sweep(self, path, multisection=False, makeSolid=True, isFrenet=False,
...
+        
+        # If thickness is 0, only a 2D surface will be returned.
+        if thickness==0:
+            combine=False
+        
+        # If a list of wires is provided, make a closed wire
+        if not isinstance(surf_edges, list): 
+            surf_edges = [o.vals()[0] for o in surf_edges.all()]
+            surf_edges = Wire.assembleEdges(surf_edges)
+            surf_edges = surf_edges.wrapped
```
adam-urbanczyk: I think it would be more elegant to do this on the interpPlate (i.e. occ_impl) level

cadquery/cq.py
```diff
@@ -2728,6 +2728,66 @@ def _sweep(self, path, multisection=False, makeSolid=True, isFrenet=False,
...
+        def _makeplate(pnt):
+            return Solid.interpPlate(surf_edges, surf_pts, thickness, Degree, NbPtsOnCur, NbIter, Anisotropie, Tol2d, Tol3d, TolAng, TolCurv, MaxDeg, MaxSegments)
+
+        plates = self.eachpoint(_makeplate, True)
+
+        # if combination is not desired, just return the created boxes
+        if not combine:
+            return plates
+        else:
+            # combine everything --> CRASHES, probably OCC 6 CAD kernel issue, better use pushpoints through each() instead
```
adam-urbanczyk: Does this also crash with clean=False?

examples/Ex101_InterpPlate.py
```diff
@@ -0,0 +1,85 @@
+import numpy as np
+from numpy import sin, cos, pi, sqrt
```
adam-urbanczyk: Is numpy really needed?

tests/test_cadquery.py
```diff
@@ -2176,3 +2182,74 @@ def test_assembleEdges(self):
...
+        
+        decimal_places = 0
+        
+        # example from PythonOCC core_geometry_geomplate.py, use of thickness = 0 returns 2D surface.
+        thickness = 0
+        edge_points = [[0.,0.,0.], [0.,10.,0.], [0.,10.,10.], [0.,0.,10.]]
+        surface_points = [[5.,5.,5.]]
+        plate_0 = Workplane('XY').interpPlate(edge_points, surface_points, thickness)
+        self.assertTrue(plate_0.val().isValid())
+        self.assertAlmostEqual(plate_0.val().Area(), 141.218823892, decimal_places)
```
adam-urbanczyk: Out of curiosity: how did you calculate the "correct" area?

adam-urbanczyk:
@bragostin  your reasoning sounds correct to me. Thanks for all the hard  work, this a really nice addition to CQ. Otherwise I had few comments and questions.

tests/test_cadquery.py
dumblob: A comment in the source code would probably be enough to explain how to calculate it manually.

cadquery/occ_impl/shapes.py
bragostin: The original OCC function args are like this:
``BRepOffsetAPI_MakeFilling (const Standard_Integer Degree=3, const Standard_Integer NbPtsOnCur=15, const Standard_Integer NbIter=2, const Standard_Boolean Anisotropie=Standard_False, const Standard_Real Tol2d=0.00001, const Standard_Real Tol3d=0.0001, const Standard_Real TolAng=0.01, const Standard_Real TolCurv=0.1, const Standard_Integer MaxDeg=8, const Standard_Integer MaxSegments=9)``
All capitalized and using ``Anisotropie`` instead of ``Anisotropy``

I have changed it in ``cq.py`` to lowercase and anisotropy:
``def interpPlate(self, surf_edges, surf_pts=[], thickness=0, combine=True, clean=True, degree=3, nbPtsOnCur=15, nbIter=2, anisotropy=False, tol2d=0.00001, tol3d=0.0001, tolAng=0.01, tolCurv=0.1, maxDeg=8, maxSegments=9):``

In ``shapes.py`` I left the original OCC argument naming convention because ultimately ``BRepOffsetAPI_MakeFilling`` needs to be called with OCC convention. Is it OK like this?

cadquery/occ_impl/shapes.py
bragostin: Here the idea is that ``BRepOffset_MakeOffset()`` accepts as offset negative and positive values, it just changes the direction of offsetting. The ``abs()`` is just here to make sure that the thickening of the surface is performed in both cases when thickness is different than 0. 
I modified the docstring ``:type thickness: float (may be negative or positive depending on thicknening direction)``
So in my opinion there would be no need to trigger an error. Is that right?

cadquery/occ_impl/shapes.py
bragostin: Done also in ``shape.py`` as explained in above comment https://github.com/CadQuery/cadquery/pull/253#discussion_r363340912

cadquery/cq.py
bragostin: Done. Yes it is cleaner and actually saved 3 lines of code.

examples/Ex101_InterpPlate.py
bragostin: call to numpy removed.

tests/test_cadquery.py
bragostin: The numerical values are the result of ``plate_0.val().Area()`` on my own machine, under Linux.
I don't know if it actually makes sense to use this as a test or if I should come with something else. Any thoughts on this?

cadquery/cq.py
bragostin:
```python
plate_3 = cq.Workplane('XY').pushPoints(pts).interpPlate(edge_wire, surface_points, thickness, combine=True, clean=False, degree=2, nbPtsOnCur=20, nbIter=2, anisotropy=False, tol2d=0.00001, tol3d=0.0001, tolAng=0.01, tolCurv=0.1, maxDeg=8, maxSegments=9)
```
just runs forever (still waiting for the result), while
```python
def face(pos):
    return be.rotate((0,0,0),(0,0,1), 30).translate(pos).val()
plate_3 = cq.Workplane('XY').pushPoints(pts).each(face)
```
runs instantly.

cadquery/cq.py
bragostin: Here the idea is that a simple list of points can still be supplied (more convenient sometimes, for example Schwarz D surfaces have 6 straight borders in 6 different planes; in such case the boundary can be defined simply by providing 6 points).
So the user has the choice of defining the surface boundaries as a list of wires or a list of points that will be polygonized.

cadquery/cq.py
jmwright: You can try reducing `tol3d` to something like 0.001 or even 0.1. Tolerance is a common culprit for me when the kernel hangs while doing an operation.

tests/test_cadquery.py
jmwright: What else would you use other than area? The nice thing about area is that it's a single metric that tells you a lot about whether complex geometry was created properly or not.

cadquery/occ_impl/shapes.py
adam-urbanczyk: +1

tests/test_cadquery.py
adam-urbanczyk: I have nothing against area - I was just curious about the reference value.

adam-urbanczyk: @bragostin maybe indeed add a comment that it was computed numerically and that's it.

cadquery/cq.py
adam-urbanczyk: OK, clear.

cadquery/cq.py
bragostin: @jmwright tried that, has no effect. I suspect the CAD kernel it does not manage to achieve a proper union of such complex shapes when CQ calls ``self.union(plates, clean=clean)`` when ``combine=True``

tests/test_cadquery.py
bragostin: OK, I added this comment in the docstring of the test: ``Tests the interpPlate() functionnalites.      Numerical values of Areas and Volumes were obtained with the Area() and Volume() functions on a Linux machine under Debian 10 with python 3.7.``

Fine tuning the number of decimal places for each test now yields this:
```python
self.assertAlmostEqual(plate_0.val().Area(), 141.218823892, 1)
self.assertAlmostEqual(plate_1.val().Volume(), 26.124970206, 3)
self.assertAlmostEqual(plate_2.val().Volume(), 10.956054314, 0)
self.assertAlmostEqual(plate_3.val().Volume(), 0.459661032, 1)
self.assertAlmostEqual(plate_4.val().Volume(), 7.760559490, 3)
```

bragostin:
I would like to propose a further improvement. I think it is not a good idea to have the thickening of the surface as an option passed to the ``interpPlate`` function. The reason is that if the user wants to use ``pushpoints`` to pattern the plate after thickening, then the edges will not match correctly (the thickening creates distortions at the surface edges.)

Rather I propose to create a new ``thichen`` method to the ``Shell`` class like this:
```python
from OCC.Core.BRepOffset import BRepOffset_MakeOffset, BRepOffset_Skin
from OCC.Core.GeomAbs import GeomAbs_Intersection
def _thicken(self, thickness):
    solid = BRepOffset_MakeOffset()
    solid.Initialize(self.wrapped, thickness, 1.e-5, BRepOffset_Skin, False, False, GeomAbs_Intersection, True) #The last True is important to make solid
    solid.MakeOffsetShape()
    return cq.Shape.cast(solid.Shape())
cq.Shell.thicken = _thicken
```
This way one can use ``pushpoints`` to sew 2D surfaces created by ``interpPlate`` this way:
```python
fin = cq.Workplane('XY').interpPlate(edge_points, surface_points)
def af(pos):
    return fin.translate(pos)
finS = cq.Workplane('XY').pushPoints(xyz).each(af)
finS = [o.vals()[0] for o in finS.all()]
finS = cq.Shell.makeShell(finS)
```
and after that thicken the new sewn surface like this:
```python
finS = finS.thicken(thickness)
```
This way there are no discontinuities in the thickened patterned plate. What do you think?

dumblob:
@bragostin if your proposal gets in, then the example should definitely go to `examples/` as "thicken".

bragostin:
Actually this would be cleaner I think:
```python
def _thicken(self, thickness):
    solid = BRepOffset_MakeOffset()
    solid.Initialize(self.val().wrapped, thickness, 1.e-5, BRepOffset_Skin, False, False, GeomAbs_Intersection, True) #The last True is important to make solid
    solid.MakeOffsetShape()
    return cq.Shape.cast(solid.Shape())
cq.Workplane.thicken = _thicken
```
Single surface thickening:
```python
fin = cq.Workplane('XY').interpPlate(edge_points, surface_points)
fin = fin.thicken(thickness)
```
Sewn surfaces thickening:
```python
def af(pos):
    return fin.translate(pos)
fin = cq.Workplane('XY').pushPoints(xyz).each(af)
fin = [o.vals()[0] for o in fin.all()]
fin = cq.Shell.makeShell(fin)
fin = cq.Workplane('XY').add(fin).shells()
fin = fin.thicken(thickness)
```

cadquery/cq.py
adam-urbanczyk: @bragostin does it always crash (i.e. even for simple shapes)? Is so it would be good to disable this branch and emit a warning. Afte moving to OCCT7.4 we could revisit this.

bragostin: @adam-urbanczyk actually I realized that pushpoints is not properly managed in the ``interpPlate()`` function, I would need to work on it, maybe the crash comes from that.
Also the question of thickening inside the function or outside as a workplane method has to be decided I think.
So this PR is anyway not ready yet I would say and I agree that it is probably better to wait after moving to OCCT7.4 to proceed with it.

adam-urbanczyk:
@bragostin if you want to add thicken to this PR, please put all OCP specific code (i.e. `BRepOffset_MakeOffset` to occ_impl/shapes.py). With that being said, you could keep this PR small and make another PR with offsetting. It would be  even fine to just implement the `occ_impl` layer here and move the fluent discussion to yet another PR. This is pretty complex so small steps are more managable.

Regarding `thicken` you could just add a `thickness=0` keyword param. That would be rather self-explanatory as far as I'm concerned.

bragostin:
@adam-urbanczyk good idea, I will make a separate PR for ``thicken`` since it should work for any 2D surface actually, not only those from ``interpPlate``

Regarding the ``pushpoints`` issue, it works fine if ``combine=False``. So for the time being I have set the default value at False in the ``interpPlate`` function. I have added a short info about that in the docstring also. Is that ok?

As you suggested I have also set ``thickness=0`` in the ``interpPlate`` function.

bragostin:
From my side I would say that this is ready to be merged. I think I addressed all the issues,or is there something missing?

jmwright:
@adam-urbanczyk Any objections to me merging this now?

adam-urbanczyk:
This is quite big PR (thanks for the work @bragostin !) so give me a day or two to take a final look @jmwright .

adam-urbanczyk:
Thanks again for this PR @bragostin !

bragostin:
@adam-urbanczyk, my pleasure, thank you for taking the time to check the code!

Issue #261, opened 2019-12-29
Implement ellipse shape


adam-urbanczyk:
cf. `GC_MakeArcOfEllipse`

Related to #260

bernhard-42:
I also missed ellipses in one project and I am happy to give it a try. 
Some first test are promising, however, I have an issue that the ellipse center is slightly translated if major_radius != minor-radius (which actually is the point about ellipses ...) I could further look into it if you are interested

bernhard-42:
Found a possible reason for the shifted center for ellipses with `major_radius > minor_radius`:

```python
from OCC.Core.gp import gp_Ax1, gp_Ax2, gp_Elips, gp_Pnt, gp_Dir
from OCC.Core.GC import GC_MakeArcOfCircle, GC_MakeArcOfEllipse
from OCC.Core.BRepBuilderAPI import BRepBuilderAPI_MakeEdge
from OCC.Core.GProp import GProp_GProps
from OCC.Core.BRepGProp import brepgprop_LinearProperties, brepgprop_SurfaceProperties

pnt = gp_Pnt(0.0, 0.0, 0.0)
dir = gp_Dir(0.0, 0.0, 1.0)

ellipse_gp = gp_Elips(gp_Ax2(pnt, dir), 4000, 40)
ellipse = BRepBuilderAPI_MakeEdge(ellipse_gp).Edge()

l_gp = ellipse_gp.Location()
print("ellipse_gp: %9.6f %9.6f %9.6f" % (l_gp.X(), l_gp.Y(), l_gp.Z()))

Properties = GProp_GProps()
brepgprop_SurfaceProperties(ellipse, Properties)
com = Properties.CentreOfMass()
print("ellipse (surface): %9.6f %9.6f %9.6f" % (com.X(), com.Y(), com.Z()))

Properties = GProp_GProps()
brepgprop_LinearProperties(ellipse, Properties)
com = Properties.CentreOfMass()
print("ellipse (linear): %9.6f %9.6f %9.6f" % (com.X(), com.Y(), com.Z()))

from cadquery.occ_impl.geom import BoundBox
bb = BoundBox._fromTopoDS(ellipse, tol=1e-8)
print("BoundingBox: %9.6f %9.6f %9.6f" % bb.center.toTuple())
```

results in 

```python
ellipse_gp:  0.000000  0.000000  0.000000
ellipse (surface):  0.000000  0.000000  0.000000
ellipse (linear): -81.295066  0.000000  0.000000
[...]

Issue #256, opened 2019-12-24
docs give examples of tuples where they should be Vectors


marcus7070:
https://github.com/CadQuery/cadquery/blob/e69b2f83bde43ad6a0159a72bc058165e684e72e/cadquery/selectors.py#L287-L338

The docs suggest `DirectionMinMaxSelector((0, 0, 1), True)`, but this code:
```python
from cadquery import *
test = Workplane("XY").box(1, 1, 1).faces(DirectionMinMaxSelector((0, 0, 1), True))
```
fails with `AttributeError: 'tuple' object has no attribute 'wrapped'`. 

Changing the tuple to a Vector works:
```python
test = Workplane("XY").box(1, 1, 1).faces(DirectionMinMaxSelector(Vector(0, 0, 1), True))
```

jmwright:
Many of the methods will implicitly convert a tuple to a Vector so that either can be passed by the caller. Either we missed that in DirectionMinMaxSelector, or it was dropped in the transition to PythonOCC.

jmwright:
Here's an example: https://github.com/CadQuery/cadquery/blob/f9d9ae06406aac2feea4a2a94569de76d473a8d0/cadquery/occ_impl/shapes.py#L330

marcus7070:
Haha @jmwright, you're too quick, I was in the middle of writing the following when you commented.

Rather than change the docs I would suggest we consistently handle tuples or Vectors. But what is the consistent CadQuery way of doing this?

We check type for tuple here:
https://github.com/CadQuery/cadquery/blob/e69b2f83bde43ad6a0159a72bc058165e684e72e/cadquery/occ_impl/shapes.py#L481-L501

Request tuple in docs and init a Vector here:
https://github.com/CadQuery/cadquery/blob/e69b2f83bde43ad6a0159a72bc058165e684e72e/cadquery/selectors.py#L62-L94

But the Vector class handles being initiated with a `Vector`, so in the above both a `Vector` or a tuple would work.

Personally I want to check for the attribute the method needs. eg. `DirectionMinMaxSelector` needs to `dot` the vector, so change the init to
```python
def __init__(self, vector, directionMax=True, tolerance=0.0001): 
         self.vector = vector if hasattr(vector, 'dot') else Vector(vector)
```

marcus7070:
Also found
[...]

PR #265, opened 2020-01-03
Ellipse

bernhard-42:
I have implemented ellipses following the implementation of circles, however, having some more flexibility. Compared to circle the following parameters are avaiable:

- `x_radius`, `y_radius`: While OCC needs `major_radius > minor_radius`, the implementation allows for arbitrary `x` and `y` radiuses. If `x_radius < y_radius`, the implementation swaps both and rotates the resulting ellipse by `+90°`. This was done to simplify building arbitrary ellipses in cadquery
- `angle1` (default=`360`), angle2 (default=`360`): allow to return an arc between `angle1`and `angle2`. This actually just makes `GC_MakeArcOfEllipse`available
- `closed` (default=`False`): If an arc, return a closed wire if `True`
- `rotation_angle` (default=`0.0`): rotate the resulting ellipse or arc around its center if `rotation_angle != 0.0`

### DIFF
--- a/cadquery/cq.py
+++ b/cadquery/cq.py
@@ -1585,13 +1585,12 @@ class Workplane(CQ):
 
 
         # Start building the ellipse with the current point as center
-        # to support strating as the result of center(x,y) or moveTo(x,y) do not use local coordinates
         center = self._findFromPoint(useLocalCoords=False)
-        e = Edge.makeEllipse(x_radius, y_radius, center, Vector(0, 0, 1), angle1, angle2, sense==1)
-
+        e = Edge.makeEllipse(x_radius, y_radius, center, self.plane.zDir, self.plane.xDir, angle1, angle2, sense==1)
+        
         # Rotate if necessary
         if rotation_angle != 0.0:
-            e = e.rotate(center, center.add(Vector(0,0,1)), rotation_angle)
+            e = e.rotate(center, center.add(self.plane.zDir), rotation_angle)
 
         # Move the start point of the ellipse onto the last current point
         if startAtCurrent:
@@ -2076,7 +2075,7 @@ class Workplane(CQ):
         :type forConstruction: true if the wires are for reference, false if they are creating
             part geometry
         :return: a new CQ object with the created wires on the stack
-        
+
         *NOTE* Due to a bug in opencascade (https://tracker.dev.opencascade.org/view.php?id=31290)
         the center of mass (equals center for next shape) is shifted. To create concentric ellipses
         use Workplane("XY")
@@ -2084,7 +2083,7 @@ class Workplane(CQ):
             .center(0, 0).ellipse(50, 5)
         """
         def makeEllipseWire(obj):
-            elip = Wire.makeEllipse(x_radius, y_radius, obj, Vector(0, 0, 1), rotation_angle=rotation_angle)
+            elip = Wire.makeEllipse(x_radius, y_radius, obj, Vector(0, 0, 1), Vector(1, 0, 0), rotation_angle=rotation_angle)
             elip.forConstruction = forConstruction
             return elip
 
--- a/cadquery/cq.py
+++ b/cadquery/cq.py
@@ -1567,8 +1567,18 @@ class Workplane(CQ):
 
         return self.spline(allPoints, includeCurrent=False, makeWire=True)
 
-    def ellipseArc(self, x_radius, y_radius, angle1=360, angle2=360, rotation_angle=0.0, sense=1,
-                forConstruction=False, startAtCurrent=True, makeWire=False):
+    def ellipseArc(
+        self,
+        x_radius,
+        y_radius,
+        angle1=360,
+        angle2=360,
+        rotation_angle=0.0,
+        sense=1,
+        forConstruction=False,
+        startAtCurrent=True,
+        makeWire=False,
+    ):
         """Draw an elliptical arc with x and y radiuses either with start point at current point or
         or current point being the center of the arc
 
@@ -1583,11 +1593,19 @@ class Workplane(CQ):
         :param makeWire: convert the resulting arc edge to a wire
         """
 
-
         # Start building the ellipse with the current point as center
         center = self._findFromPoint(useLocalCoords=False)
-        e = Edge.makeEllipse(x_radius, y_radius, center, self.plane.zDir, self.plane.xDir, angle1, angle2, sense==1)
-        
+        e = Edge.makeEllipse(
+            x_radius,
+            y_radius,
+            center,
+            self.plane.zDir,
+            self.plane.xDir,
+            angle1,
+            angle2,
+            sense == 1,
+        )
+
         # Rotate if necessary
         if rotation_angle != 0.0:
             e = e.rotate(center, center.add(self.plane.zDir), rotation_angle)
@@ -2082,8 +2100,16 @@ class Workplane(CQ):
             .center(10, 20).ellipse(100,10)
             .center(0, 0).ellipse(50, 5)
         """
+
         def makeEllipseWire(obj):
-            elip = Wire.makeEllipse(x_radius, y_radius, obj, Vector(0, 0, 1), Vector(1, 0, 0), rotation_angle=rotation_angle)
+            elip = Wire.makeEllipse(
+                x_radius,
+                y_radius,
+                obj,
+                Vector(0, 0, 1),
+                Vector(1, 0, 0),
+                rotation_angle=rotation_angle,
+            )
             elip.forConstruction = forConstruction
             return elip
 
--- a/cadquery/cq.py
+++ b/cadquery/cq.py
@@ -2020,6 +2020,41 @@ class Workplane(CQ):
 
         return self.eachpoint(makeCircleWire, useLocalCoordinates=True)
 
+    # ellipse from current point
+    def ellipse(self, x_radius, y_radius, angle1=360, angle2=360, rotation_angle=0.0,
+                closed=False, forConstruction=False):
+        """
+        Make an ellipse for each item on the stack.
+        :param x_radius: x radius of the ellipse (x-axis of plane the ellipse should lie in)
+        :type x_radius: float > 0
+        :param y_radius: y radius of the ellipse (y-axis of plane the ellipse should lie in)
+        :type y_radius: float > 0
+        :param forConstruction: should the new wires be reference geometry only?
+        :type forConstruction: true if the wires are for reference, false if they are creating
+            part geometry
+        :param rotation_angle: angle to rotate the ellipse (0 = no rotation = default)
+        :type rotation_angle: float
+        :param angle1: start angle of arc
+        :type angle1: float
+        :param angle2: end angle of arc
+        :type angle2: float
+        :param rotation_angle: angle to rotate the created ellipse / arc
+        :type rotation_angle: float
+        :return: a new CQ object with the created wires on the stack
+        
+        *NOTE* Due to a bug in opencascade (https://tracker.dev.opencascade.org/view.php?id=31290)
+        the center of mass (equals center for next shape) is shifted. To create concentric ellipses
+        use Workplane("XY")
+            .center(10, 20).ellipse(100,10)
+            .center(0, 0).ellipse(50, 5)
+        """
+        def makeEllipseWire(obj):
+            elip = Wire.makeEllipse(x_radius, y_radius, obj, Vector(0, 0, 1), angle1, angle2, rotation_angle, closed)
+            elip.forConstruction = forConstruction
+            return elip
+
+        return self.eachpoint(makeEllipseWire, useLocalCoordinates=True)
+
     def polygon(self, nSides, diameter, forConstruction=False):
         """
         Creates a polygon inscribed in a circle of the specified diameter for each point on
--- a/cadquery/cq.py
+++ b/cadquery/cq.py
@@ -1567,6 +1567,48 @@ class Workplane(CQ):
 
         return self.spline(allPoints, includeCurrent=False, makeWire=True)
 
+    def ellipseArc(self, x_radius, y_radius, angle1=360, angle2=360, rotation_angle=0.0, sense=1,
+                forConstruction=False, startAtCurrent=True, makeWire=False):
+        """Draw an elliptical arc with x and y radiuses either with start point at current point or
+        or current point being the center of the arc
+
+        :param x_radius: x radius of the ellipse (along the x-axis of plane the ellipse should lie in)
+        :param y_radius: y radius of the ellipse (along the y-axis of plane the ellipse should lie in)
+        :param angle1: start angle of arc
+        :param angle2: end angle of arc (angle2 == angle1 return closed ellipse = default)
+        :param rotation_angle: angle to rotate the created ellipse / arc
+        :param sense: clockwise (-1) or counter clockwise (1)
+        :param startAtCurrent: True: start point of arc is moved to current point; False: center of
+            arc is on current point
+        :param makeWire: convert the resulting arc edge to a wire
+        """
+
+
+        # Start building the ellipse with the current point as center
+        # to support strating as the result of center(x,y) or moveTo(x,y) do not use local coordinates
+        center = self._findFromPoint(useLocalCoords=False)
+        e = Edge.makeEllipse(x_radius, y_radius, center, Vector(0, 0, 1), angle1, angle2, sense==1)
+
+        # Rotate if necessary
+        if rotation_angle != 0.0:
+            e = e.rotate(center, center.add(Vector(0,0,1)), rotation_angle)
+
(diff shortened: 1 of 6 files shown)


bernhard-42:
Adresses issue https://github.com/CadQuery/cadquery/issues/261

jmwright:
@bernhard-42 Thanks for putting the work into this. Is this ready for review?

bernhard-42:
@jmwright yes, it is ready for review. 

The only unresolved issue I know of is the OCCT bug about center of mass of an ellipse - which I cannot do anything. Impact: concentric ellipses need a `center(0,0)`, see doc string of ellipse method

cadquery/occ_impl/shapes.py
```diff
@@ -730,6 +730,47 @@ def makeCircle(cls, radius, pnt=Vector(0, 0, 0), dir=Vector(0, 0, 1), angle1=360
...
+        :param x_radius: x radius of the ellipse (along the x-axis of plane the ellipse should lie in)
+        :param y_radius: y radius of the ellipse (along the y-axis of plane the ellipse should lie in)
+        :param pnt: vector representing the center of the ellipse
+        :param dir: vector representing the direction of the plane the ellipse should lie in
+        :param angle1: start angle of arc
+        :param angle2: end angle of arc (angle2 == angle1 return closed ellipse = default)
+        :return: an Edge
+        """
+
+        pnt = Vector(pnt).toPnt()
```
jmwright: Could you make it so that pnt and dir could be tuples here as well? There's a discussion about how we're not handling Vector vs tuple consistently here: https://github.com/CadQuery/cadquery/issues/256

jmwright: Hold on, this isn't end-user facing at the top level of the abstraction, so it probably doesn't matter whether they Vector or tuple.

bernhard-42: The signature of `makeEllipse` is a copy of `makeCircle` and afaik both `cq.Vector(tuple)` and `cq.Vector(vector)` return a `Vector`. So `Vector(pnt)` and `Vector(dir)` will return `Vector` types even when a tuple is passed as a parameter

cadquery/cq.py
```diff
@@ -1916,6 +1916,41 @@ def makeCircleWire(obj):
 
         return self.eachpoint(makeCircleWire, useLocalCoordinates=True)
 
+    # ellipse from current point
+    def ellipse(self, x_radius, y_radius, angle1=360, angle2=360, rotation_angle=0.0,
```
adam-urbanczyk: Wouldn't angle1=0  be more logical?

cadquery/cq.py
```diff
@@ -1916,6 +1916,41 @@ def makeCircleWire(obj):
...
+        :type rotation_angle: float
+        :param angle1: start angle of arc
+        :type angle1: float
+        :param angle2: end angle of arc
+        :type angle2: float
+        :param rotation_angle: angle to rotate the created ellipse / arc
+        :type rotation_angle: float
+        :return: a new CQ object with the created wires on the stack
+        
+        *NOTE* Due to a bug in opencascade (https://tracker.dev.opencascade.org/view.php?id=31290)
```
adam-urbanczyk: Is this true? I thought that ellipse is added to `pendingWires` which are not used to calculate center.

cadquery/occ_impl/shapes.py
```diff
@@ -730,6 +730,47 @@ def makeCircle(cls, radius, pnt=Vector(0, 0, 0), dir=Vector(0, 0, 1), angle1=360
...
+        
+        if y_radius > x_radius:
+            # swap x and y radius and rotate by 90° afterwards to create an ellipse with x_radius < y_radius
+            correction_angle = 90.0 * DEG2RAD
+            ellipse_gp = gp_Elips(ax2, y_radius, x_radius).Rotated(ax1, correction_angle)
+        else:
+            correction_angle = 0.0
+            ellipse_gp = gp_Elips(ax2, x_radius, y_radius)
+        
+        if angle1 == angle2:  # full ellipse case
```
adam-urbanczyk: Do you really need this if else? Wouldn't GC_MakeArcOfEllipse suffice?

cadquery/occ_impl/shapes.py
```diff
@@ -846,6 +887,34 @@ def makeCircle(cls, radius, center, normal):
         w = cls.assembleEdges([circle_edge])
         return w
 
+    @classmethod
+    def makeEllipse(cls, x_radius, y_radius, center, normal, angle1=360.0, angle2=360.0,
```
adam-urbanczyk: What is the rationale for having two makeEllipse functions?

cadquery/occ_impl/shapes.py
bernhard-42: makeCircle is built the same way, and I wanted to be as similar with my ellipse implementation as possible.

cadquery/occ_impl/shapes.py
bernhard-42: This is just like makeCircle, it exists in class Edge and Wire.
I didn't fully understand, but again, wanted to keep the ellipse implementation as similar to the circle implementation as possible

cadquery/cq.py
bernhard-42: absolutely, however, makeCircle (and hence makeEllipse) uses 360 and I decided to keep that up to the top level. And from the later logic perspective, angle1 needs to be equal angle2 to get a full circle.

cadquery/cq.py
bernhard-42:
```python
v = (cq.Workplane("XY")
     .center(10, 20).ellipse(100,10)
     .ellipse(50, 5).vertices()).objects[0]
v.X, v.Y
```
results in `(58.34650359492122, 20.0)`

While 
```python
v = (cq.Workplane("XY")
     .center(10, 20).ellipse(100,10)
     .center(0,0).ellipse(50, 5).vertices()).objects[0]
v.X, v.Y
```

gets the correct result `(60.0, 20.0)`. This is btw. a test case I've added.

If you look at the `eachpoint` code
https://github.com/CadQuery/cadquery/blob/74573fc3bb032321c50a9e817dbf806c2a286d53/cadquery/cq.py#L1825-L1834
it actually uses `Center` and `circle` and `ellipse`  use `eachpoint`.

cadquery/cq.py
bernhard-42: ... or full ellipse

cadquery/occ_impl/shapes.py
bernhard-42: see also https://www.opencascade.com/doc/occt-6.9.0/refman/html/class_g_c___make_arc_of_ellipse.html 
`GC_MakeArcOfEllipse` constructor requires a `gp_Elips`

cadquery/occ_impl/shapes.py
adam-urbanczyk: Indeed, you are right.

cadquery/cq.py
adam-urbanczyk: I see that circle does not expose those options at all - @jmwright what do you think?

cadquery/cq.py
adam-urbanczyk: May bad, the wire is added to pendingWires and objects

cadquery/cq.py
bernhard-42: I just looked at `Workplane.spline` which follows a different approach than `Workplane.circle`:
It does only have `Edge.makeSpline` and then applies the following logic
https://github.com/CadQuery/cadquery/blob/74573fc3bb032321c50a9e817dbf806c2a286d53/cadquery/cq.py#L1441-L1450

This allows splines to be used when building 2D paths.

If we expose `angle1` and `angle2` with `ellipse` shouldn't we change it like `spline` to allow ellipse arcs being used in building 2D paths? That was my original intention with exposing both angles, but I missed the point that `circle` doesn't add itself to pendingWires/pendingEdges (and hence my `ellipse` code doesn't either.

jmwright: I wouldn't assume that the code in `circle` is the correct way to do it. I don't think that code has been touched since the early days of CQ, and may not represent our current understanding of best practices.

bernhard-42:
The more I look into the cadquery source code the more I understand from the overall approach.

After analysing `Workplane.spline` and e.g. `Workplane.sagittaArc` I would keep `Workplane.ellipse` as is, but additionally add `Workplane.ellipseArc` as a 2d construction function:

```python
def ellipseArc(self, x_radius, y_radius, angle1=360, angle2=360, rotation_angle=0.0, sense=1,
               forConstruction=False, startAtCurrent=True, makeWire=False):

    # start building the ellipse with the current point as center ...
    center = self._findFromPoint(useLocalCoords=True)
    e = Edge.makeEllipse(x_radius, y_radius, center, Vector(0, 0, 1), angle1, angle2, sense==1)

    # rotate if necessary ...
    if rotation_angle != 0.0:
        e = e.rotate(center, center.add(Vector(0,0,1)), rotation_angle)
    
    # and move the start point of the ellipse onto the last current point
    if startAtCurrent:
        startPoint = e.startPoint()
        e = e.translate(center.sub(startPoint))

    if makeWire:
        rv = Wire.assembleEdges([e])
        if not forConstruction:
            self._addPendingWire(rv)
    else:
        rv = e
        if not forConstruction:
            self._addPendingEdge(e)
    print(rv)
    return self.newObject([rv])
```

This would be analogous to circle:

- `Workplane.ellipse` adds a wire object like `Workplane.circle` (but additionally exposes the angles)
- `Workplane.allipseArc` works for 2d construction like `Workplane.sagittaArc`

Note, I had to add a sense parameter to control whether the arc goes clockwise or not.
This would now allow to write 

```python
x = (cq.Workplane("XY")
     .lineTo(0, 1)
     .ellipseArc(5, 4, -10, 190, 45, sense=-1)
)

ep = x.val().endPoint()

x = (x
     .hLineTo(ep.x + 2)
     .ellipseArc(5, 4, -10, 190, -45, sense=-1)
     .vLineTo(0)
     .mirrorX()
)
```

image

Would that be an enhancement you would like to add to cadquery?

jmwright:
`Workplane.sagittaArc` is a relatively new addition to CQ (within probably the last year), just FYI.

I think the approach you're suggesting makes sense. If you want to put the work in, I'm all for having the increased functionality.

adam-urbanczyk:
@bernhard-42 I also think that keeping `ellipse` similar to `circle` is the best approach.

bernhard-42:
I have built `ellipse` back to resemble what `circle` does. I only kept `rotation_angle` as a convenience parameter (which would not makes sense for `circle`)

I then added a `ellipseArc` method that allows to use arcs from ellipses in 2d construction. It supports a `sense` parameter for clockwise or counter clockwise arcs. 

As a side note, since no point is included in the construction, I had to find the last point with `localCoords=False` to get the correct result for starting with `center()` or eg. `moveTo`. I wanted in both cases the center of the new ellipse be either the around the new center or around the point moved to. Is this the correct way?

cadquery/cq.py
```diff
@@ -1468,6 +1468,48 @@ def parametricCurve(self, func, N=400, start=0, stop=1):
...
+        :param startAtCurrent: True: start point of arc is moved to current point; False: center of
+            arc is on current point
+        :param makeWire: convert the resulting arc edge to a wire
+        """
+
+
+        # Start building the ellipse with the current point as center
+        # to support strating as the result of center(x,y) or moveTo(x,y) do not use local coordinates
+        center = self._findFromPoint(useLocalCoords=False)
+        e = Edge.makeEllipse(x_radius, y_radius, center, Vector(0, 0, 1), angle1, angle2, sense==1)
```
adam-urbanczyk: if center is in global coords the Vector(0,0,1) is NOK I think. You might want to try: `self.plane.zDir`

cadquery/cq.py
```diff
@@ -1468,6 +1468,48 @@ def parametricCurve(self, func, N=400, start=0, stop=1):
...
+
+        # Rotate if necessary
+        if rotation_angle != 0.0:
+            e = e.rotate(center, center.add(Vector(0,0,1)), rotation_angle)
```
adam-urbanczyk: Same potential issue here

cadquery/cq.py
```diff
@@ -1468,6 +1468,48 @@ def parametricCurve(self, func, N=400, start=0, stop=1):
 
         return self.spline(allPoints,includeCurrent=False,makeWire=True)
 
+    def ellipseArc(self, x_radius, y_radius, angle1=360, angle2=360, rotation_angle=0.0, sense=1,
+                forConstruction=False, startAtCurrent=True, makeWire=False):
```
adam-urbanczyk: Open question: do we have a usecase for `startAtCurrent=False`?

cadquery/cq.py
bernhard-42: yes, it is wrong. I will change and add a test case for a transformed plane

cadquery/cq.py
bernhard-42: see above

cadquery/cq.py
bernhard-42: Good question. I found a similar construct with `spline` (`includeCurrent`) where I would also not see an immediate use case. So having seen this pattern in cq, I decided to add the same approach to `ellipseArc` and let others find out use cases.
But I am also happy to remove if someone has arguments against it.

bernhard-42:
I tried `self.plane.zDir` however, got slightly lost:

For this example:

```python
def sample(angle):
    return(
        cq.Workplane("XY")
          .transformed(offset=(20, 5, 10), rotate=(angle, 0, 0))
          .rect(10,10)
          .workplane()
          .ellipseArc(20, 10, 90, 180, sense=-1, startAtCurrent=False))
```

calling `sample(45)` results in

image

while calling `sample(46)` results in 

image

Why does changing the rotation angle from 45° to 46° change the orientation in which OCC creates ellipse arcs?
It will be in the same orientation from 46° until 135° 

image

and then flip again 136°

image

bernhard-42:
I've opened an issue in pythonocc https://github.com/tpaviot/pythonocc-core/issues/773

bernhard-42:
I went for the `gp_Ax2(pnt, zDir, xDir)` approach as mentioned in https://github.com/tpaviot/pythonocc-core/issues/773 to ensure the orientation of ellipse arcs. If I now plot ellipseArcs onto a sphere (`Rx` and `Ry`rotate `(0,0,20)` around the x-axis or y-axis):

```python
ellipses1 = [cq.Workplane("XY").transformed(offset=Rx(i, (0,0,20)), rotate=(i,0,0)).ellipseArc(3, 5, 0, 180, startAtCurrent=False).close().extrude(1) for i in range(0, 360, 30)]
ellipses2 = [cq.Workplane("XY").transformed(offset=Ry(i, (0,0,20)), rotate=(0,i,0)).ellipseArc(3, 5, 0, 180, startAtCurrent=False).close().extrude(1) for i in range(0, 360, 30)]
```

I get arcs of ellipses nicely aligned with the sphere surface:
image

jmwright:
@bernhard-42 @adam-urbanczyk Where are we at with this pull request? Is pythonocc-core issue #773 a blocker for merging this?

cadquery/occ_impl/shapes.py
```diff
@@ -743,6 +744,62 @@ def makeCircle(
...
+        y_radius,
+        pnt=Vector(0, 0, 0),
+        dir=Vector(0, 0, 1),
+        xdir=Vector(1, 0, 0),
+        angle1=360.0,
+        angle2=360.0,
+        sense=1,
+    ):
+        """
+        Interpolate a spline through the provided points.
```
```suggestion
        Makes an Ellipse centered at the provided point, having normal in the provided direction
```

cadquery/occ_impl/shapes.py
```diff
@@ -861,6 +918,46 @@ def makeCircle(cls, radius, center, normal):
...
+        """
+            Makes an Ellipse centered at the provided point, having normal in the provided direction
+            :param x_radius: floating point major radius of the ellipse (x-axis), must be > 0
+            :param y_radius: floating point minor radius of the ellipse (y-axis), must be > 0
+            :param center: vector representing the center of the circle
+            :param normal: vector representing the direction of the plane the circle should lie in
+            :param angle1: start angle of arc
+            :param angle2: end angle of arc
+            :param rotation_angle: angle to rotate the created ellipse / arc
+            :return:
```
```suggestion
            :return: Wire
```

adam-urbanczyk:
@bernhard-42 @jmwright two minor comments on the docstrings. Otherwise I think it can be merged. Using `zDir` and `xDir` sounds OK to me.

bernhard-42:
@adam-urbanczyk Thought committing your proposals would be sufficient, however there is now a conflict. Shall I rebase my branch and push or can you resolve that?

adam-urbanczyk:
I will resolve the conflict.

adam-urbanczyk:
Alright, merging. Thanks for the contribution @bernhard-42 !

PR #270, opened 2020-01-23
removed CQ.firstPoint from CQ.__init__

marcus7070:
`CQ.__init__` creates an attribute called `firstPoint`. I can't find any references to the `CQ` attribute `firstPoint`, only the modelling context's `firstPoint`. I'm guessing it's just an artifact from before the modelling context was implemented and should be removed.


jmwright:
@marcus7070 Thanks for this. I believe you are correct that nothing is ever done with `Workplane.firstPoint` and so it serves no purpose.

PR #308, opened 2020-04-03
Update azure pipelines to use MacOS 10.15

adam-urbanczyk:
Will fix failing azure pipelines


adam-urbanczyk:
Trivial fix, so going ahead with merging.

Issue #202, opened 2019-10-23
Using 2 times pushPoints() offsets positions


bragostin:
Running the code below produces the results shown in the attached image: there is an offset of 0.6 mm between the holes through the shell and the holes through the cylinders while they should be coincident. In some other cases it goes up to 1.6 mm. Could this be coming from some accuracy setting somewhere?

```python
import numpy as np
import cadquery as cq

y_sh = np.array([14.5,15,15,14.5,29.5,62.3,70,70,119,119,125,125,129.8,131.8,132,132]) - 140./2.
z_sh = np.array([22.5,39.5,89.5,107.5,15.9,122.5,22.5,107.5,22.5,107.5,39.5,89.5,140.,15.9,108,126]) - 145./2.
dsh = np.array([6.5,6.5,6.5,6.5,5.5,5.5,6.5,6.5,6.5,6.5,6.5,6.5,5.5,5.5,5.5,5.5])

cq_object = cq.Workplane('XY').box(5, 140, 145).cut(cq.Workplane("XY" ).box(4, 138, 143))

y_pl = np.linspace(-64, 64, 13)
z_pl = np.linspace(-66.5, 66.5, 13)
y_pl, z_pl = np.meshgrid(y_pl, z_pl)
yz_pl = [(y,z) for y,z in zip(y_pl.flatten(), z_pl.flatten())]
pl = cq.Workplane('YZ').workplane(offset=-2).polygon(4, 10).extrude(4)
def s_pillar(pos):
    return pl.translate(pos).val()
pillars = cq.Workplane('YZ').pushPoints(yz_pl).each(s_pillar)

for dh in np.unique(dsh):
    ysh = y_sh[dsh==dh]
    zsh = z_sh[dsh==dh]
    yzsh = [(y,z) for y,z in zip(ysh.flatten(), zsh.flatten())]
    cq_object = cq_object.faces('<X').workplane().pushPoints(yzsh).polygon(12, dh).cutBlind(-5)
    pillars = pillars.faces('<X').workplane().pushPoints(yzsh).hole(dh)
    
show_object(cq_object)
show_object(pillars)
```
![holes_pos](https://user-images.githubusercontent.com/26216511/67380960-6aa57b80-f58b-11e9-86a8-1abb4e18ac5e.png)

adam-urbanczyk:
@bragostin could you try the following in your loop?
```python
pillars = pillars.faces('<X').workplane(centerOption='ProjectedOrigin')
```
[...]

PR #313, opened 2020-04-07
Warn about upcoming change of the centerOption

adam-urbanczyk:
`centerOption` default value will change from "CenterOfMass" to "ProjectedOirigin"

Related to #202

### DIFF
--- a/cadquery/cq.py
+++ b/cadquery/cq.py
@@ -33,6 +33,8 @@ from . import (
     exporters,
 )
 
+from .utils import deprecate_kwarg
+
 
 class CQContext(object):
     """
@@ -305,6 +307,7 @@ class CQ(object):
 
         return self.objects[0].wrapped
 
+    @deprecate_kwarg("centerOption", "ProjectedOrigin")
     def workplane(
         self, offset=0.0, invert=False, centerOption="CenterOfMass", origin=None
     ):
--- a/cadquery/utils.py
+++ b/cadquery/utils.py
@@ -1,0 +1,26 @@
+from functools import wraps
+from inspect import signature
+from warnings import warn
+
+
+class deprecate_kwarg:
+    def __init__(self, name, new_value):
+
+        self.name = name
+        self.new_value = new_value
+
+    def __call__(self, f):
+        @wraps(f)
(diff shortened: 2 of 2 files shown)


jmwright:
I think the code looks good. I do have a few questions.

1. What's the next release that is mentioned in the warning? Does that include RC releases, or just the first stable release?
2. I only get the warning on the first call of `workplane`. I'm assuming this is normal behavior, but I wanted to confirm that, and to make sure that's the intended behavior. I'm fine with it if it is.
3. I get the following output:
```
>>> import cadquery as cq
>>> result = cq.Workplane("XY").workplane().box(1, 1, 1)
/home/jwright/Downloads/repos/cadquery/cadquery/utils.py:21: FutureWarning: Default walue of centerOption will change in next relase to ProjectedOrigin
  FutureWarning,
```
Why does the last `FutureWarning` have a trailing comma? Is there supposed to be more output after that?

adam-urbanczyk:
> I think the code looks good. I do have a few questions.
> 
>     1. What's the next release that is mentioned in the warning? Does that include RC releases, or just the first stable release?

I'd actually propose to release 2.0 soon, and change this in 2.1.

>     2. I only get the warning on the first call of `workplane`. I'm assuming this is normal behavior, but I wanted to confirm that, and to make sure that's the intended behavior. I'm fine with it if it is.

This is your local configuration I think - I cannot reproduce it.

>     3. I get the following output:
> 
> 
> ```
> >>> import cadquery as cq
> >>> result = cq.Workplane("XY").workplane().box(1, 1, 1)
> /home/jwright/Downloads/repos/cadquery/cadquery/utils.py:21: FutureWarning: Default walue of centerOption will change in next relase to ProjectedOrigin
>   FutureWarning,
> ```
> 
> Why does the last `FutureWarning` have a trailing comma? Is there supposed to be more output after that?

I'm afraid this is the default (not so nice) formatting of the warning message showing the line of code where the warnining is invoked. I'd propose to leave as-is.

jmwright:
Sounds good. Do you want to merge, or should I?

adam-urbanczyk:
I'll merge it.

Issue #266, opened 2020-01-03
largestDimension() does not return the largest dimension


greyltc:
Why is it the diagonal of the bounding box times five?
https://github.com/CadQuery/cadquery/blob/74573fc3bb032321c50a9e817dbf806c2a286d53/cadquery/cq.py#L2020

jmwright:
I think that code has been around since the first commit of CadQuery 1.x, so I'm not sure why it's that way. The [test](https://github.com/CadQuery/cadquery/blob/3cd466a693ce1d7f084e5cd6ddee256fc28d020a/tests/test_cadquery.py#L1052) seems to account for it being that way.

@dcowden Do you remember why it's that way? Was it a work-around for something in FreeCAD?

adam-urbanczyk:
@jmwright  I think it was related to how cuts used to be make (i.e. with CSG ).

adam-urbanczyk:
Do you want to use `largestDimension` for something @greyltc ? Otherwise I'd like to close the issue.

greyltc:
Yeah. I wanted to use it to find the largest dimension of an object :wink: 
I ended up writing my own that's pretty similar. It returns an accurate value (barring some corner cases that I didn't bother to address) and lets the user decide to measure along X, Y or Z:

https://github.com/greyltc/mechanical/blob/master/toolbox/utilities.py#L144-L161

It's not a drop in replacement, but it gets the job done for me.

I'm not really sure why this should be closed though, besides it being annoyoing to have issues open :wink:. The function as it is today doesn't do what its namesake says it does so I'd imagine it could cause someone else problems down the road if it's left like this.

jmwright:
I think at one time I tried using `longestDistance` to set camera position, but ended up using the bounding box, and combining the extents of the bounding boxes of all the objects in the scene when there was more than one.

https://github.com/jmwright/cadquery-gui/blob/master/assets/js/viewer.js#L157

My preference would be to leave this open and determine an example case that shows how it should work. If we can determine the expected behavior I'll put together a PR when I get a chance.

greyltc:
We could mesh the objects and calculate the distance between all possible pairs of verticies and return the largest value of that list...

adam-urbanczyk:
@greyltc this sounds rather slow. There is OBB in opencascade https://www.opencascade.com/doc/occt-7.4.0/refman/html/class_b_rep_bnd_lib.html#aed61fe542529ccd72e7e8fdad61a0bbe we could use it.
[...]

PR #163, opened 2019-07-10
Fix findSolid if multiple solids present on the stack

adam-urbanczyk:
This PR will resolve #159

PR #317, opened 2020-04-11
Making largestDimension work as expected

jmwright:
**WIP - do not merge.**

Intended to address #266 but I've run into a couple of issues.

1. Several tests fail now because `largestDimension` is used for automatically determining hole depth. I don't think the original method was doing this correctly, and I need to get this sorted out before this PR is viable.
Ex: https://github.com/CadQuery/cadquery/blob/6ecb0fc96f2379fbfd8de7f5b935dd880be99047/cadquery/cq.py#L2347

I think that's a reason that only the bounding box of the first solid was checked. You wouldn't want the hole depth to be based on all the solids in a compound (or would you?). In any case, I suspect the original implementation was kind of a hack to get hole depth to work properly. I need to think about this a bit to figure out how to refactor the code for hole depth.

2. I'm not getting the enclosing sphere diameter that I expected. For a unit cube (1x1x1), I would expect the diagonal (which is the diameter of the enclosing sphere?) to be sqrt(1^2 + 1^2) or 1.414, but my method using the Vector length yields about 1.76. I'm probably missing something simple in vector length vs diagonal/hypotenuse using two sides of the square.

Thoughts and guidance welcome.

@adam-urbanczyk @greyltc

### DIFF
--- a/cadquery/cq.py
+++ b/cadquery/cq.py
@@ -2279,13 +2279,51 @@ class Workplane(CQ):
         how long or wide a feature must be to make sure to cut through all of the material
         :return: A value representing the largest dimension of the first solid on the stack
         """
+        xmin = ymin = zmin = 999999
+        xmax = ymax = zmax = -999999
+
+        # Get all the solids contained within this CQ object
+        solids = self.solids().vals()
+
+        # Protect against this being called on something like a blank workplane
+        if len(solids) == 0: return -1
+
+        # Find the combined outer bounds of all the solids
+        for solid in solids:
+            # Mins
+            if solid.BoundingBox().xmin < xmin:
+                xmin = solid.BoundingBox().xmin
+            if solid.BoundingBox().ymin < ymin:
+                ymin = solid.BoundingBox().ymin
+            if solid.BoundingBox().zmin < zmin:
+                zmin = solid.BoundingBox().zmin
+
+            # Maxes
+            if solid.BoundingBox().xmax > xmax:
+                xmax = solid.BoundingBox().xmax
+            if solid.BoundingBox().ymax > ymax:
+                ymax = solid.BoundingBox().ymax
+            if solid.BoundingBox().zmax > zmax:
+                zmax = solid.BoundingBox().zmax
+
+        xLength = (xmax - xmin)
+        yLength = (ymax - ymin)
+        zLength = (zmax - zmin)
+
+        centroid = Vector(xLength, yLength, zLength)
+
+        # Calculate the sphere size of the outer bounds of all solids
+        sphereSize = centroid.Length
+
+        return sphereSize
+
         # TODO: this implementation is naive and returns the dims of the first solid... most of
         # TODO: the time this works. but a stronger implementation would be to search all solids.
-        s = self.findSolid()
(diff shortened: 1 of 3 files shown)


jmwright:
Nevermind on issue 2. My brain was stuck thinking in terms of 2d vectors, not 3d. 1.76 is correct for a 3d diagonal of a 3d unit cube.

jmwright:
Reworked the code a little bit which now gives a slightly larger answer. `BoundingBox().DiagonalLength`, which was used before, consistently gives a value on the low side.

It seems that this method was just created to serve thru cuts by returning a distance that would be sure to be through the object. I think that's where the multiplication by 5 came in. It was was a hard coded overkill value to make sure it always returned a value that was long enough. That's fine when all it does is thru cuts. However, when you try to use this function to aid in setting something like camera distance, it gives the wrong answer.

It now gives a correct (and intuitive) answer, while not breaking the tests for thru cuts.

jmwright:
@adam-urbanczyk It's a small-ish change, but could potentially impact thru cuts. Probably worth a quick check before I merge, even though I know you're busy with OCP.

cadquery/cq.py
```diff
@@ -2279,14 +2279,32 @@ def largestDimension(self):
...
+        zmax = compound.BoundingBox().zmax
+
+        # Find the length of each axis
+        xLength = xmax - xmin
+        yLength = ymax - ymin
+        zLength = zmax - zmin
+
+        # Calculate the sphere size of the outer bounds of all solids from the lengths along each axis
+        centroid = Vector(xLength, yLength, zLength)
+        sphereSize = centroid.Length
```
adam-urbanczyk, line 2304: I might be missing something but aren't you calculating the diagonal length by hand here? Wouldn't this work?
```
return compound.BoundingBox().DiagonalLength
```

jmwright: That's the way it was before, but it gives a slightly different answer that's smaller. My concern is if I use that the cut thru operations will start coming up short in some edge cases. However, I can use the built in diagonal length and then switch back to doing it by hand if users start encountering issues.

adam-urbanczyk: Is it a large difference?

jmwright:
It was between 0.5% and 1% in most of my testing, which wasn't terribly extensive.

cadquery/cq.py
jmwright: It was between 0.5% and 1% in most of my testing, which wasn't terribly extensive.

adam-urbanczyk: i think it is this line: https://github.com/CadQuery/cadquery/blob/b81319afd3a9586a61819da4666e9934fbdb1fac/cadquery/occ_impl/geom.py#L858

TOL=1e-2

jmwright: The PR has been updated to use DiagonalLength again. Essentially the only change is that the multiplication by 5 was removed. When that function was originally written, I don't think `findSolid` returned more than the first solid on the stack either, and that's since been fixed in #163

adam-urbanczyk:
Thanks @jmwright !

PR #321, opened 2020-04-14
Extrude with both=True fix

### LINKED ISSUE #320: Extrude with both=True produces two solids i.s.o. one
The code below produces two solids i.s.o. one:

```python

import cadquery as cq

points = [(20, 20), (20, 30), (30, 30), (30, 20)]

solid = (cq.Workplane("YZ")
         .polyline(points)
         .close()
         .extrude(distance=10., both=True))
```

It looks like `combineWithBase` does not fuse if there is no parent solid. I think we should change this behavior.

adam-urbanczyk:
Will resolve #320

### DIFF
--- a/cadquery/cq.py
+++ b/cadquery/cq.py
@@ -2675,6 +2675,8 @@ class Workplane(CQ):
         r = obj
         if baseSolid is not None:
             r = baseSolid.fuse(obj)
+        elif isinstance(obj, Compound):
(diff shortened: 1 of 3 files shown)


adam-urbanczyk:
It is ready @jmwright , the only question is if it should be merged before or after 2.0 .

adam-urbanczyk:
Alright, merging.

PR #378, opened 2020-06-09
MyPy in CI [WIP]

adam-urbanczyk:
Just running mypy over the codebase. There are no annotations yet.

EDIT:
I added some annotations in geom.


adam-urbanczyk:
@jmwright  I added the OCP stubs here: https://github.com/CadQuery/OCP-stubs . It was necessary to make this thing work. I did manage to find one bug in the geom.py code that way, I think this will be very helpful in the long run. Do you want to take another look, or should I merge as is?

jmwright:
I'm good with merging.

adam-urbanczyk:
OK, if this round passes (black was complaining) I'll merge

Issue #247, opened 2019-12-20
Provide type annotations


adam-urbanczyk:
This will allow to use tools like Mypy. We could try to generate the stubs automatically.

Issue #118, opened 2019-03-22
Polar array support with changing orientation


adam-urbanczyk:
Currently there is no easy way to achieve the right result:

![afbeelding](https://user-images.githubusercontent.com/13981538/54803179-bd722a00-4c6e-11e9-9785-cc180291e00a.png)

Wheras it is trival to get to the left one:
```python
r1 = cq.Workplane("XY" )\
    .box(10, 10, 1).faces(">Z")\
    .polarArray(3.5,0,360,6).rect(1.6,1.6).cutThruAll()
```

I think the right scenario occurs quite frequently, so it would make sense to support it.

Related to:
https://github.com/CadQuery/CQ-editor/issues/31

adam-urbanczyk:
I was thinking about how to implement it this and came up with the following design

1. extend `geom.Vector` so that it allows to attach a transform object to itself
2. refactor `polarArray` to create `Vectors` with proper rotation transforms attached
3. refactor `eachpoint` to apply the attached transform if it exists

Any thoughts?

fapbrandt:
Hi. Why not in 2.0RC2?

jmwright:
@fapbrandt 2.0RC2 is more about finalizing a stable version of CadQuery that runs closer to the OCC CAD kernel than FreeCAD allowed us to. We accept PRs for new features like this for the 2.0 release candidates, but the core development team probably won't have time to get it done before version 2.1.

fapbrandt:
Thanks for answer, jmwright. I personally will check cq.py to see what I can do in my own code. See you.

jmwright:
@fapbrandt You can probably implement this for use in your own scripts using plugins.

https://cadquery.readthedocs.io/en/latest/extending.html#cadqueryexample-plugins

PR #380, opened 2020-06-16
Location implementation and CQ typing

adam-urbanczyk:
This is related to #118  and #247. I think it will be a solid basis for further extending of CQ.

- [x]  Wrap `TopLoc_Location`
- [x]  Annotate cq.py
- [x]  Rework `eachpoint` and related methods
- [x] Update docs
- [x] Do not add the origin of workplane to `objects`, handle in `eachpoint` and `val`. This makes using `add` more intuitive.

### DIFF
--- a/cadquery/cq.py
+++ b/cadquery/cq.py
@@ -19,9 +19,14 @@
 
 import math
 from copy import copy
+from typing import overload, Sequence, Union, Tuple, Type, Optional, Any, Iterable, Callable, List
+from typing_extensions import Literal
+
 from . import (
     Vector,
     Plane,
+    Matrix,
+    Location,
     Shape,
     Edge,
     Wire,
@@ -33,8 +38,11 @@ from . import (
     exporters,
 )
 
-from .utils import deprecate_kwarg
+from .utils import deprecate_kwarg, deprecate
 
+CQObject = Union[Vector,Location,Shape]
+Selector = selectors.Selector
+VectorLike = Union[Tuple[float,float,float],Vector]
 
 class CQContext(object):
     """
@@ -43,6 +51,12 @@ class CQContext(object):
         All objects in the same CQ chain share a reference to this same object instance
         which allows for shared state when needed,
     """
+
+    pendingWires : List[Wire]
+    pendingEdges : List[Edge]
+    firstPoint : Optional[Vector]
+    tolerance : float
+
 
     def __init__(self):
         self.pendingWires = (
@@ -63,8 +77,13 @@ class CQ(object):
     Examples include feature selection, feature creation, 2d drawing
     using work planes, and 3d operations like fillets, shells, and splitting
     """
+    
+    objects : List[CQObject]
+    ctx : CQContext
+    parent : Optional['CQ']
+    _tag : Optional[str]
 
-    def __init__(self, obj):
+    def __init__(self, obj : Optional[CQObject]) -> None:
         """
         Construct a new CadQuery (CQ) object that wraps a CAD primitive.
 
@@ -79,7 +98,7 @@ class CQ(object):
         if obj:  # guarded because sometimes None for internal use
             self.objects.append(obj)
 
-    def newObject(self, objlist):
+    def newObject(self, objlist : Sequence[CQObject]) -> 'CQ':
         """
         Make a new CQ object.
 
@@ -98,7 +117,7 @@ class CQ(object):
         r.objects = list(objlist)
         return r
 
-    def tag(self, name):
+    def tag(self, name : str) -> 'CQ':
         """
         Tags the current CQ object for later reference.
 
@@ -109,7 +128,7 @@ class CQ(object):
         self._tag = name
         return self
 
-    def _collectProperty(self, propName):
+    def _collectProperty(self, propName : str) -> Sequence[CQObject]:
         """
         Collects all of the values for propName,
         for all items on the stack.
@@ -141,7 +160,7 @@ class CQ(object):
 
         return list(all.values())
 
-    def split(self, keepTop=False, keepBottom=False):
+    def split(self, keepTop: bool=False, keepBottom: bool=False) -> 'CQ':
         """
             Splits a solid on the stack into two parts, optionally keeping the separate parts.
 
@@ -182,8 +201,9 @@ class CQ(object):
                 return self.newObject([top])
             else:
                 return self.newObject([bottom])
-
-    def combineSolids(self, otherCQToCombine=None):
+    
+    @deprecate()
+    def combineSolids(self, otherCQToCombine: Optional['CQ']=None) -> 'CQ':
         """
         !!!DEPRECATED!!! use union()
         Combines all solids on the current stack, and any context object, together
@@ -198,7 +218,6 @@ class CQ(object):
         combined and returned on the stack of the new object.
         """
         # loop through current stack objects, and combine them
-        # TODO: combine other types of objects as well, like edges and wires
         toCombine = self.solids().vals()
 
         if otherCQToCombine:
@@ -221,7 +240,7 @@ class CQ(object):
 
         return self.newObject([s])
 
-    def all(self):
+    def all(self) -> Sequence['CQ']:
         """
         Return a list of all CQ objects on the stack.
 
@@ -233,13 +252,13 @@ class CQ(object):
         """
         return [self.newObject([o]) for o in self.objects]
 
-    def size(self):
+    def size(self) -> int:
         """
          Return the number of objects currently on the stack
         """
         return len(self.objects)
 
-    def vals(self):
+    def vals(self) -> Sequence[CQObject]:
         """
         get the values in the current list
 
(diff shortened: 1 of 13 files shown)


adam-urbanczyk:
@jmwright  @dcowden  I did more or less finish annotating `cq.py` and it turns out that the code is not self-consistent anymore. `CQ` and `Workplane` classes are intermixed - sometimes `CQ` is assuming it is actually a `Workplane`. Although it is fixable, fixing it would break backward-compatibility. I think it would be best to just merge `CQ` and `Workplane` into one class (`Workplane`) and provide an alias 'CQ` for backward compatibility. Do you agree with such an approach?

The other thing I changed (for better I hope) is the `eachpoint` and related methods. They are now using the new `Location` class (which can represent rotations as well as translations). It has implications for people writing their own plugins (maybe it'd impact you @michaelgale ). 

All in all I think it is worth it, although slightly breaking.

jmwright:
> ...and it turns out that the code is not self-consistent anymore. CQ and Workplane classes are intermixed...

@adam-urbanczyk When you say "anymore", are you saying that the process of annotation caused this problem?

> Although it is fixable, fixing it would break backward-compatibility. I think it would be best to just merge CQ and Workplane into one class (Workplane) and provide an alias 'CQ` for backward compatibility. Do you agree with such an approach?

Changing something so fundamental in the core of CQ makes me nervous. Are you saying the CQ alias prevent will prevent breakage of all existing scripts, or just some of them?

adam-urbanczyk:
@jmwright I meant that somewhere along the way we ended up in the situation where methods of CQ call methods of Workplane.  Probably at the beginning there was a nice design, but it was not enforced and a somewhat entangled situation materialized. We can either fix it (move methods around and break backward compatibility) or merge the two classes into one (which I did in this PR). Annotating+mypy check **_revealed_** the issue. I start to appreciate mypy more and more.

To be honest merging of `CQ` and `Workplane` did not feel fundamental at all - I deleted a couple of lines and merged the two `__init__` methods. All tests just passed.  I expect that all scripts will just work.

What did change is the way plugins need to be written - they need to accept `Location` instead of `Vector`.  I could still add some logic inside `eachpoint` but I'd prefer to keep things clean. Take a look what changed in the tests.

adam-urbanczyk:
To be more specific, here is an example of the entanglement I was referring to above: 
https://github.com/CadQuery/cadquery/blob/f522899ae928fbd12169bebc2ed89cd5324a9ba5/cadquery/cq.py#L169
`CQ.split` calls `self.rect` which is not defined in the  class `CQ`.

jmwright:
I see now. It doesn't sound like as much of a breaking change as I expected. It would be nice to hear from someone who uses the plugin system about those changes.

adam-urbanczyk:
Absolutely @jmwright , I'm curious though if such a person exists.

An alternative approach would be to add another method, say `eachloc`, and deprecate `eachpoint` while keeping the old implementation. I'm very much open for discussion but I also think that we need to start moving forward if there are good reasons to do so. Current approach for representing locations (translation based) is extremely limiting and does not facilitate solving of #118 in a nice way. I'd like to work on sketches and (simple) assemblies in the near future and having `TopLoc_Location` will be extremely handy in that context.

jmwright:
> I'm curious though if such a person exists.

Maybe not. I know that we have pointed some people to plugins over the years, but I don't know how many are using it. I do know that a couple of spots in our codebase were designed to test the plugin functionality.

https://github.com/CadQuery/cadquery/blob/f9fe7b1b842bce3a590ec8e6a1becdc2f04bd054/tests/test_cadquery.py#L118

If the tests like that work, we're probably ok.

michaelgale:
The changes to `eachPoint` etc. and the introduction of the `Location` class do not appear to impact my code built around CQ.  I am curious about the `Location` class as a wrapper for `TopLoc_Location`--is this intended for use internally within the CQ API or does it have some benefits for user code as well?  More broadly, the `geom.py` module has some good utility functions that I accidentally implemented myself since I only discovered its functionality later.  Often I find that I need to do some quick (and often trivial and repetitive) computation on things like points and rectangles--it would be nice if these convenience geometry functions could be baked into `geom.py`.  The number of times I type `x + w/2, x - w/2` etc. or simply want to query coordinate min, max, mid, etc. is so numerous.  It would be nice to quickly compute a point rotated by an offset angle or the bounds of a rectangle shifted by some offset (or rotated) etc.  Just a thought.

adam-urbanczyk:
@michaelgale I'd like to start using `Location` (`TopLoc_Location` wrapper) to represent location in future developments of CQ. It is used in the kernel to represent [relative] position (see e.g. `Locate`, `Located`, `Move`, `Moved` methods of `TopoDD_Shape`), so I think it is the right approach. First natural application would be to use it in assemblies. I'd like to start with implementing a container that maps well to the STEP exporter in OCCT and then add a constraint solver.

adam-urbanczyk:
I think I'm happy with this PR. We can let it sink in for a couple of days, maybe some user of the plugin system will still show up here.

I'd still like to annotate the `Shapes.py` file as well. But I'll start another PR for that.

jmwright:
+1 for merging on the 25th if there are no issues brought up by the community.

adam-urbanczyk:
There seem to be no complaints, so I'd propose to move forward with this.

jmwright:
@adam-urbanczyk I see there's now a merge conflict.

adam-urbanczyk:
Resolved.

cadquery/cq.py
```diff
@@ -2787,27 +2983,30 @@ def cut(self, toCut, clean=True):
...
+    ) -> "Workplane":
         """
         Intersects the provided solid from the current solid.
         
         :param toIntersect: object to intersect
         :type toIntersect: a solid object, or a CQ object having a solid,
         :param boolean clean: call :py:meth:`clean` afterwards to have a clean shape
         :raises: ValueError if there is no solid to intersect with in the chain
-        :return: a CQ object with the resulting object selected
+        :return: a CQ object with the resulting iobject selected
```
jmwright, line 2996: What's an iobject?

cadquery/occ_impl/geom.py
```diff
@@ -852,3 +858,72 @@ def isInside(self, b2: "BoundBox") -> bool:
             return True
         else:
             return False
+
+
+class Location(object):
+    """Relative location in 3D space"""
```
jmwright, line 864: Should there be more background info in this docstring?

cadquery/occ_impl/geom.py
```diff
@@ -852,3 +858,72 @@ def isInside(self, b2: "BoundBox") -> bool:
...
+
+    @overload
+    def __init__(self, t: TopLoc_Location) -> None:
+        ...
+
+    @overload
+    def __init__(self, t: Vector, ax: Vector, angle: float) -> None:
+        ...
+
+    def __init__(self, *args):
```
jmwright, line 892: This chunk of code might benefit from a comment or two.

cadquery/cq.py
```diff
@@ -888,17 +1058,19 @@ def mirror(self, mirrorPlane="XY", basePointVector=(0, 0, 0)):
         newS = self.newObject([self.objects[0].mirror(mirrorPlane, basePointVector)])
         return newS.first()
 
-    def translate(self, vec):
+    def translate(self, vec: VectorLike) -> "Workplane":
```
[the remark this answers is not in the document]
adam-urbanczyk: FYI: that is the correct way when you are dealing with forward references. Only in newer python versions it is possible to use the more natural syntax: https://www.python.org/dev/peps/pep-0563/

cadquery/cq.py
```diff
@@ -272,16 +397,16 @@ def add(self, obj):
...
         return self
 
-    def val(self):
+    def val(self) -> CQObject:
         """
         Return the first value on the stack
 
         :return: the first value on the stack.
-        :rtype: A FreeCAD object or a SolidReference
+        :rtype: A CAD primitives
```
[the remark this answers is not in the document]
adam-urbanczyk: Done

cadquery/cq.py
adam-urbanczyk: Done

cadquery/occ_impl/geom.py
adam-urbanczyk: Yep, I added more info.

cadquery/occ_impl/geom.py
adam-urbanczyk: Added description for every overload.

adam-urbanczyk:
I think everything is fixed @jmwright

jmwright:
Great, thanks for all the work you've put into this! Merging.

maderero:
> 
> 
> I see now. It doesn't sound like as much of a breaking change as I expected. It would be nice to hear from someone who uses the plugin system about those changes.

I'm a few weeks late, but I just want to report that while the changes did initially break my plugins, it did not take long to implement a fix once I took a look at the changes.

adam-urbanczyk:
Good to hear @maderero ! Are you planning to share your code? It is always interesting to see what people are doing with CQ.

maderero:
> Good to hear @maderero ! Are you planning to share your code? It is always interesting to see what people are doing with CQ.

At some point I could share the plugins, which are essentially shortcuts to make [common structural steel shapes ](https://theconstructor.org/wp-content/uploads/2010/10/types-of-tension-members-in-structural-steel-construction-629x420.gif). 

Here's a sample of some [wide-flange beams](https://i.imgur.com/4Ufx2Sv.png) in the cq-editor window.

adam-urbanczyk:
Cool @maderero , thanks for sharing! BTW: you might be interested in the `importers.importDXF` function.

maderero:
I'm glad you like it, @adam-urbanczyk! I will keep that function in mind as well.

PR #381, opened 2020-06-19
Fix just a few unimportant typos

nopria:


jmwright:
Thanks for the contribution @nopria !

PR #384, opened 2020-06-22
Fix docstring example of 'slot2D'

nopria:


jmwright:
Thanks for the contribution @nopria 

`slot2D` was originally named `slot`, but got changed in commit https://github.com/CadQuery/cadquery/commit/c378d496ba05e0aa6c3b47310c4d78dcc402183b

Looks like the example never got updated to go with it.

Issue #218, opened 2019-11-25
Support alternative workplane centre options


abudden:
The default centre option is Centre of Mass (which makes a lot less sense to me than Centre of Bounding Box, but that's by-the-by and at least the bounding box centre is also an option).  Often though, I have a face of a workpiece and I want to start drawing at one of the corners of that face.

At the moment, I do this by calculating the dimensions of the face (as I tend to do everything in the model by calculation so fundamental dimensions can be tweaked) and adding a `.center` call or a `.moveTo` call to get to a given corner.  This is a lot of boilerplate and it would be nice if there were options to `.workplane` that made the starting/centre point any of the 8 bounding box extremes (North, North-East, East etc).

adam-urbanczyk:
We have selectors for that. Check out the docs: https://cadquery.readthedocs.io/en/latest/selectors.html#combining-selectors

NB: the default behaviour will change from "Center of Bounding Box" to "ProjectedOrigin"  cf. #202 

If you post a simple example I can offer more guidance.

abudden:
Thanks @adam-urbanczyk; I guess I need to better understand how the selectors work - I'd used them with faces all over the place and struggled a lot.  I found that e.g. '<Z[0]' produced error messages related to no objects found, so I've ended up with code like:

```
 .faces('+Z').faces('not (<Z)').faces('not (<Z)').faces('not (<Z)').faces('<Z').faces('not (<X)').faces('<X')
```

(that's a real example from my model).

I haven't got a "simple example" at the moment; I'll do some experimentation and try to find a simple demonstration of the sort of thing I'm trying to do.

As an aside, I'd be really really grateful if there was someone who could look at a not-very-simple example and give me general comments.  I'm fairly new at CQ and would love to have someone look at my full model and give any advice on better way of doing things.  However, the model is 396 lines long (of which about 100 lines are comments and about 90 lines are drawing angle-iron cross-sections over-and-over again), so I'd understand if no-one wants to read it!  The model is a personal one, so there's no problem posting it online if there is anyone willing to read it...

There are lots of things that feel wrong, for example:

* The long string of `.faces` that just seems like the wrong way to do it.
* The regular use of `moveTo` rather than using a selector to pick the corner of a face (ref this issue).
[...]

PR #386, opened 2020-06-23
more introduction to Workplanes in the docs

marcus7070:
I was trying to help a user ([here](https://github.com/CadQuery/cadquery/issues/218#issuecomment-648051515)) and I wanted to point them to somewhere in our docs where they could get the concept of a workplane and how solids are created relative to a workplane. I think they hadn't encountered the concept of workplanes before. I realised the docs don't say too much about the concept of workplanes (probably because back when they were written the link to FreeCAD was much stronger, so it was assumed knowledge). So I've added some more words to the docs. Do you think they help? I try to be brief because I know docs with too many words won't be read at all.

### DIFF
--- a/doc/primer.rst
+++ b/doc/primer.rst
@@ -39,16 +39,23 @@ See :ref:`apireference` to learn more.
 Workplanes
 ---------------------------
 
-Workplanes represent a plane in space, from which other features can be located. They have a center point and a local
-coordinate system.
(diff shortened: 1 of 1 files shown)


adam-urbanczyk:
Thanks @marcus7070 , it does look useful.

adam-urbanczyk:
Small docs modification, so just merging.

jmwright:
Yes, thanks @marcus7070

PR #387, opened 2020-06-23
Better BuildCurves3d_s parameters in makeHelix

### LINKED ISSUE #385: Incorrect makeHelix output
Method `makeHelix` gives incorrect result when the height, expressed in terms of  *number of turns*, is greater than a treshhold (42 in the following case):

```
import cadquery as cq

R = 50 # helix radius
N = 43 # helix turns - incorrect output for N > 42
P= 10 # helix pitch

show_object(cq.Wire.makeHelix(pitch=P, height=N*P, radius=R))
```

incorrect-helix-opt

adam-urbanczyk:
This will resolve #385 . There parameters are chosen "experimentally", so future changes are possible.

### DIFF
--- a/cadquery/occ_impl/shapes.py
+++ b/cadquery/occ_impl/shapes.py
@@ -1055,7 +1055,7 @@ class Wire(Shape, Mixin1D):
 
         # 4. Convert to wire and fix building 3d geom from 2d geom
         w = BRepBuilderAPI_MakeWire(e).Wire()
-        BRepLib.BuildCurves3d_s(w)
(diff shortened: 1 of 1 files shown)


adam-urbanczyk:
I think this PR can be merged @jmwright . Current settings produce a correct result for at least 1000 turns:
```python
import cadquery as cq

R = 500 # helix radius
N = 1000 # helix turns - incorrect output for N > 42
P= 10 # helix pitch

w = cq.Wire.makeHelix(pitch=P, height=N*P, radius=R)
show_object(w)
```
I think this is enough for a typical user. In the future we might want to assemble the final helix form multiple edges.

jmwright:
Thanks @adam-urbanczyk , merging.

PR #388, opened 2020-06-24
Fix docstring of method `radiusArc`

nopria:


jmwright:
Thanks for the contribution @nopria

PR #391, opened 2020-07-02
Type annotations for shapes

adam-urbanczyk:


adam-urbanczyk:
I think I'm done. Enough of MyPy for quite some time, but I hope it will be beneficial in the long run.

adam-urbanczyk:
One question @jmwright : I fixed `CombinedCenterOfBoundBox` (it was relying on a method that did not exist) but maybe it is better to completely remove it? No one ever complained about it not working so I guess is is effectively dead.

jmwright:
I'd kind of like to keep it. That's because I've implemented versions of something similar in other apps. I don't think I ever knew that it existed in CQ.

adam-urbanczyk:
Alright,  I'll add a test for it then.

adam-urbanczyk:
Done, merging.

Issue #393, opened 2020-07-06
Shell does not work without selected faces


adam-urbanczyk:
The OCCT kernel supports such an operation (i.e. creation of an empty but watertight object) and it is a useful feature for additive manufacturing related applications.

![obraz](https://user-images.githubusercontent.com/13981538/86612483-79fd1580-bfb0-11ea-8eee-63958034dc4d.png)

adam-urbanczyk:
Solved by #394

PR #394, opened 2020-07-08
Fully closed shelling

adam-urbanczyk:
This will solve #393 . I was motivated by the discussion on our google group and it is rather small fix. Now one can create hollow models easily:
```python
b = cq.Wokrplane().box(1,1,1).shell(-0.1)
```
obraz


adam-urbanczyk:
Merging.

Issue #347, opened 2020-05-07
Expose 2d offset operation


adam-urbanczyk:
Wrap `BRepOffsetAPI_MakeOffset` operation from OCCT:

![obraz](https://user-images.githubusercontent.com/13981538/81329422-d5895080-909e-11ea-9be8-1529331031c3.png)

adam-urbanczyk:
Solved by #397

PR #397, opened 2020-07-10
offset2D implementation

### LINKED ISSUE #208: Question: offset curve / tolerance
I want to expand a 2D polyline surface by 1mm.
Similar to https://forum.freecadweb.org/viewtopic.php?t=9137

Or, is it on the roadmap?
https://cadquery.readthedocs.io/en/latest/roadmap.html#d-operations

adam-urbanczyk:
Will resolve #208 and #347 

- [x] Implementation
- [x] Docs

obraz

### DIFF
--- a/cadquery/cq.py
+++ b/cadquery/cq.py
@@ -19,6 +19,7 @@
 
 import math
 from copy import copy
+from itertools import chain
 from typing import (
     overload,
     Sequence,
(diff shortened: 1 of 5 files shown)


cadquery/cq.py
```diff
@@ -1933,44 +1934,6 @@ def tangentArcPoint(
 
         return self.newObject([arc])
 
-    '''
-    def rotateAndCopy(self, matrix: Matrix) -> "Workplane":
```
jmwright, line 1937: Why was this commented out and now removed?

cadquery/cq.py
```diff
@@ -3675,6 +3648,26 @@ def toPending(self) -> "Workplane":
 
         return self
 
+    def offset2D(
+        self, d: float, kind: Literal["arc", "intersection", "tangent"] = "arc"
+    ) -> "Workplane":
+        """
+        Creates a 2D offset wire.
+        
+        :param float d: thickness. Negative thickness denotes offset to inside.
```
jmwright, line 3657: So this implies that a 0 offset can be used, even though it's not practical to do so. Does the kernel handle that properly?

tests/test_cadquery.py
```diff
@@ -2771,7 +2770,7 @@ def testMakeShellSolid(self):
                 lines.append(
                     Edge.makeLine(Vector(*vertices[v1]), Vector(*vertices[v2]))
                 )
-            wire = Wire.combine(lines)
+            wire = Wire.combine(lines)[0]
```
jmwright, line 2773: Why is only the first wire used here after the combine?

cadquery/cq.py
adam-urbanczyk: I just removed commented code. Looks like I might have commented it our by mistake in previous PRs - I'll revert this change.

cadquery/cq.py
adam-urbanczyk: 0 offset just works

tests/test_cadquery.py
adam-urbanczyk: I changed Wire.combine to return a list of Wires. Now it handles disconnected edges/wires properly, but obviously such a modification is needed.

cadquery/cq.py
adam-urbanczyk: I did some brief archeology and now it is clear why I commented it out: it was calling the method `rotateShapes` which was not implemented (i.e. throwing a `NotImplementedError`). Hence I removed both.

jmwright:
Ok. I think all discussion items have been resolved. Merging.

PR #404, opened 2020-07-13
Implement locationAt

### LINKED ISSUE #389: Support generating locations along a curve
For complex sweeps with varying sections and a complicated spine.

See: http://www.cppblog.com/eryar/archive/2018/04/03/OpenCASCADE_Trihedron_Law.html

adam-urbanczyk:
This will resolve #389 . This is exposed only at shapes level, I don't think there is a elegant way to use expose this at the fluen api level.

Here is an example of what can be done with this functions. It can be very useful for modeling thread ends.
obraz

```python
from cadquery import *

r = cq.Workplane().rect(3,1).extrude(0.1).faces('<Z').val()

s2 = Wire.makeHelix(7,20,5,angle=360).Edges()[0]
show_object(s2)
N = 20
locs = s2.Edges()[0].locations([0] + [i/N for i in range(1,N+1)])

show_object(Compound.makeCompound(r.located(l) for l in locs))
```

### DIFF
--- a/cadquery/occ_impl/shapes.py
+++ b/cadquery/occ_impl/shapes.py
@@ -36,7 +36,7 @@ from OCP.gp import (
 
 # collection of pints (used for spline construction)
 from OCP.TColgp import TColgp_HArray1OfPnt
-from OCP.BRepAdaptor import BRepAdaptor_Curve, BRepAdaptor_Surface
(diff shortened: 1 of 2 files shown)


adam-urbanczyk:
OK, merging.

Issue #360, opened 2020-05-24
How to export parts to SVG?


lalebarde:
Hi,

I found this function in the documentation:

> **exportSvg(fileName)**
> Exports the first item on the stack as an SVG file
> For testing purposes mainly.
> **Parameters: fileName** (String, absolute path to the file) – the filename to export

I have tested with this simple example:

```
import cadquery as cq
result = cq.Workplane("XY").box(80, 60, 10).faces(">Z").workplane().hole(22)
result.exportSvg("/home/lalebarde/test/test_export_svg_solid.svg")
result.faces(">Z").exportSvg("/home/lalebarde/test/test_export_svg_face.svg")
show_object(result)
```

If I import the face in say QCAD, I get:

![image](https://user-images.githubusercontent.com/3611049/82749823-2641be80-9dac-11ea-8dce-645611ac2097.png)

and for the solid:

![image](https://user-images.githubusercontent.com/3611049/82749850-5b4e1100-9dac-11ea-8a08-08796ecf3593.png)

There is only a useless 3D view. How can I export my design? I need to print plans for the faces.

lalebarde:
Right, I have [installed](https://github.com/jmwright/cadquery-freecad-module/blob/master/docs/index.md#introduction) CQ in FreeCAD and can use all FreeCAD export features.

Besides, thought it is off topic, I need to print 1:1 big pieces plans on polyester (>A0). For this purpose, I hire a professional printing service. They have 900mm or 594mm large polyester rolls, so in practice, there is no limit for the length, which is what I need for 3 meters long pieces.

I am new to FreeCAD. I could see that to get 2D plans from a 3D model, I have to use the _Drawing_ module. But the biggest it proposes is A0, requiring for my use case 3 pages, which is not desirable.

In addition, the printing service company requires pdf only.

So:

- **How could I get 3 meters long pdf exports at 1:1 scale precisely (no adjustments)?** If it is not possible, I would fall back on A0.

- **Is it possible to configure A0 overlaps so that _Drawing_ manages itself exports/prints in several pages with connection marks?** I have tried to project a selected 3m long piece in an A0 sheet with scale 1, but it has fitted it in the page.

lalebarde:
The following sequence: {FreeCAD → CQ → module Drawing A0 → Export SVG flattened → LibreOffice Draw or InkScape → Export pdf} works. If I reopen the pdf under InkScape and measure, it is equal exactly to the original parameters

[...]

Issue #340, opened 2020-05-04
Current SVG rendering not usable for CNC work


adam-urbanczyk:
While working on  #339 I noticed that the edges exported to SVG are not connected. This issue is orthogonal to #246 .

adam-urbanczyk:
This looks interesting: https://github.com/cduck/drawSvg

adam-urbanczyk:
2D DXF export solved by #415 , closing.

Issue #334, opened 2020-04-29
makeCompound needed to save all objects in STL


bragostin:
Why does this yield only the first object in the stl file as if ``each()`` had been ignored
```python
def air_fin(pos):
    return fin.translate(pos).val()

x_fin = cq.Workplane('XY').pushPoints(xy_af).each(air_fin)
x_fin.val().exportStl("test.stl", 0.05)
```
while this saves correctly all the patterned objects in the file?
```python
def air_fin(pos):
    return fin.translate(pos).val()
        
x_fin = cq.Workplane('XY').pushPoints(xy_af).each(air_fin)
x_fin = cq.Compound.makeCompound([o.vals()[0] for o in x_fin.all()])
x_fin.exportStl("test.stl", 0.05)
```

adam-urbanczyk:
`val()` return first value from the current `objects`. `cq.exporters.exportShape` does the same actually so I'd say it is a (functional) bug.

adam-urbanczyk:
This is related to #212 and #216

bragostin:
Then maybe a new ''exportStep'' (or ''exportStl'') could be in this form:
```python
    def exportStep(self, fileName):
        """
        surface_curve_mode:
            0: write without pcurves (2 times smaller STEP file)
            1 (Default): write with pcurves
        """
      
        obj = Compound._makeCompound([o.vals()[0] for o in self.all()])
        c = STEPControl_Controller()
        c.Init() 
        Interface_Static_SetCVal("write.step.schema", "AP214")
        Interface_Static_SetIVal('write.surfacecurve.mode', 0)
        writer = STEPControl_Writer()
        writer.Transfer(obj.wrapped, STEPControl_AsIs)
        
        return writer.Write(fileName)
``` 
This would save all the objects in the stack I presume, and it also gives control on more of the export STEP options of Opencascade. For example setting ''surfacecurve.mode'' to 0 divides the file size by 2.

adam-urbanczyk:
I guess yes, the place to put this change would be here:
https://github.com/CadQuery/cadquery/blob/1491e0ccf446a9cbeb4a6d7b19f0922918ba648d/cadquery/occ_impl/exporters.py#L70
We need to rethink the exporters API anyhow, the possibility to control `surfacecurve.mode` looks like a good idea.

michaelgale:
[...]

Issue #212, opened 2019-11-21
Export multiple objects to a single step file


abudden:
I've just discovered CQ2 and my first impressions are that it's brilliant.  I've been trying to figure out how to get step files for a multi-object model out of a script and have drawn a blank: there doesn't seem to be any documentation and I haven't figured it out by trawling through the code either.

Given a simple model like the one below, I can load it in CQ-editor and export the model with no issues.  This works fine, but I'd rather not use CQ-editor (it changes the line-endings in the text file resulting in huge diffs and moving the model around is painful without 3D mouse support).

So I know it's possible to export multiple objects in one STEP file as CQ-editor can do it, but how do I do the same in CQ python code?

Example code:

```python
#!/usr/bin/python3
# vim: set fileencoding=utf-8 :
import cadquery as cq

# Create a box
length, width, height = (60.0, 80.0, 10.0)
box1 = (cq.Workplane('XY')
        .box(length=length, width=width, height=height)
        # Get the face highest in the Z axis
        .faces('>Z').workplane()
        # Draw a centred rectangle, 20×40
        .rect(xLen=20, yLen=40)
        # Cut the rectangle through the box to make a square pocket
        .cutThruAll()
        )

# Make a second box
box2 = (cq.Workplane('XY')
        # Move the centre of the workplane away from the first box
        .center(200.0, 100.0)
        # Draw the box
        .box(length=20.0, width=30.0, height=40.0)
        )

box3 = (cq.Workplane('XY')
        # Move the centre of the workplane away from the first and second boxes
        .center(-100.0, 50.0)
        # Draw the box
        .box(length=50.0, width=100.0, height=100.0)
        )

# If in CQ-Editor GUI:
if 'show_object' in globals():
    globals()['show_object'](box1) # Odd way of calling it to avoid a pylint error
    globals()['show_object'](box2) # Odd way of calling it to avoid a pylint error
    globals()['show_object'](box3) # Odd way of calling it to avoid a pylint error
    # Now select all objects under "CQ models" in the Objects pane and hit Tools...Export as STEP
else:
    # If not in CQ-Editor GUI:
    # How do I make this export box1, box2 and box3 in place in a single step file?
    box1.val().exportStep("example2.step")
```

bragostin:
Can you try this?

```python
vals = []
for o in box1.all(): vals.extend(o.vals())
for o in box2.all(): vals.extend(o.vals())
for o in box3.all(): vals.extend(o.vals())
[...]

PR #415, opened 2020-07-27
DXF export (2D) and exporters cleanup

### LINKED ISSUE #352: Export to DXF for 2D sections
In relation to #340 it might easier to implement export to DXF. It is a more common format used for CNC work.

#122 is somewhat related to this.

adam-urbanczyk:
Will resolve #352 , #340, #212, #334 and #360 .

This adds dxf export. Moreover I implemented a new export function and marked exportShape as deprecated:
```python
def export(
    w: Union[Shape, Workplane],
    fname: str,
    exportType: Optional[ExportLiterals] = None,
    tolerance: float = 0.1,
): ...
```
Now exporting is more user friendly: `exporters.export(obj,'out.step')`


doc/apireference.rst
```diff
@@ -120,7 +120,7 @@ File Management and Export
 .. autosummary::
     importers.importStep
     importers.importDXF
-    exporters.exportShape
+    exporters.export
```
jmwright, line 123: I expect this change is going to break some existing scripts where users have used `exportShape` directly.

jmwright: I'm not saying "don't do it", I'm just pointing out that we may get issues filed over this change.

adam-urbanczyk: It is so far only marked as deprecated in the code. I'd propose to remove it in 2.2 so there is enough time for ppl to react on the deprecation warning.

jmwright: Sounds good, +1

PR #417, opened 2020-07-28
Fix rotate docstring

winksaville:


jmwright:
Thanks for this fix @winksaville

PR #418, opened 2020-07-30
Setup.py fix

adam-urbanczyk:
Use find_packages() in setup.py and perform azure builds on PRs


adam-urbanczyk:
This is a simple packaging fix so just merging.

PR #434, opened 2020-08-12
Updated the installation to reference OCP and pointed to the master installation option

jmwright:
I kept repeating the same installation method to help people get the latest features, so I updated the docs. There were a couple of outdated references to PythonOCC that I removed as well.


adam-urbanczyk:
Good idea, thanks!

Issue #20, opened 2018-12-06
add Color property for solid objects


dcowden:
<a href="https://github.com/easyw"><img src="https://avatars1.githubusercontent.com/u/3032347?v=4" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [easyw](https://github.com/easyw)**
_Saturday Aug 08, 2015 at 09:04 GMT_
_Originally opened as https://github.com/dcowden/cadquery/issues/105_

----

It would be useful to add Color property for solid objects and keep color for all functions (e.g. union cut etc.)
thanks

dcowden:
<a href="https://github.com/hyOzd"><img src="https://avatars2.githubusercontent.com/u/1479166?v=4" align="left" width="48" height="48" hspace="10"></img></a> **Comment by [hyOzd](https://github.com/hyOzd)**
_Sunday Aug 09, 2015 at 02:39 GMT_

----

I was going to create an issue for "exporting STEP with colors". Since I know we are both here for the same thing I thought I would just add this here : ) Yeah we want to export colorfuls STEPs from cadquery.

dcowden:
<a href="https://github.com/jmwright"><img src="https://avatars0.githubusercontent.com/u/1015439?v=4" align="left" width="48" height="48" hspace="10"></img></a> **Comment by [jmwright](https://github.com/jmwright)**
_Sunday Aug 09, 2015 at 03:23 GMT_

----

@easyw and @hyOzd - Thanks for the ideas. The CadQuery module for FreeCAD allows you to specify the color and transparency for an object when it's rendered. However, it sounds like you are both saying that you want the color to be embedded as an intrensic part of a solid. My thinking is that the color property would only matter when exporting the solid (STLs/STEPs with color), or displaying it (in FreeCAD). Are these the only use cases you both see for it?

On the surface this doesn't sound too bad, but operations that result in multiple solids (split), or operations on compound solids might get a little interesting. I'd need to research it a bit.

dcowden:
<a href="https://github.com/hyOzd"><img src="https://avatars2.githubusercontent.com/u/1479166?v=4" align="left" width="48" height="48" hspace="10"></img></a> **Comment by [hyOzd](https://github.com/hyOzd)**
_Sunday Aug 09, 2015 at 04:14 GMT_

----

Let me correct; I myself only care for colors when exporting or displaying. But it would be nice to be able to fuse parts, and conserve the colors of different faces. Like freecad does;

[...]

PR #440, opened 2020-08-17
Assembly support

### LINKED ISSUE #276: Implement lightweight container for assemblies
* Place existing parts in an assembly
* Allow to specify color/metadata per part
* Export STEP assemblies
* Import STEP assemblies

adam-urbanczyk:
This PR will eventually resolve #276, #20

What's left

- [x] ~~Exact jacobian~~ , finite diff jacobian by hand
- [x] Finalize solver choice -> BFGS it is
- [x] Implement parameters handling (e.g. specific angle or distance constraint)
- [x] Visually attractive example in docs/README


jmwright:
Tagging @bernhard-42 on this since jupyter-cadquery supports lightweight assemblies, and it may be useful to compare notes.

adam-urbanczyk:
Not yet mergable but getting close. Could you take a look and let me know what you think @jmwright ? BTW: preliminary CQ-editor is here https://github.com/CadQuery/CQ-editor/pull/181 .

jmwright:
This is amazing work @adam-urbanczyk thanks! I have an environment set up and have begun testing. Over the past few weeks I have been working on a script that generates about 60 large scale objects, with most of them being the same object (a cube representing a junction). CQ-editor crashes periodically when I run the script (I assume) because I'm running show_object so much with the same object, just copied in different locations. I really need to be using some sort of instancing. I plan to convert that script to use the new assembly functionality as I hope that will help with my issue and will give me a good feel for the new assemblies.

One thing I did notice on the CQ-editor assembly branch is that I'm getting an artifact in the corners of the cubes like part of a line that should be hidden is being drawn.

Screenshot_20200904_220609

adam-urbanczyk:
It was already there before assembly support. It's some kind of rendering glitch that shows up only when the helper axes are visible.
obraz

jmwright:
> It was already there before assembly support.

Ok, it must just be more noticeable for me with the gray object.

I converted my script to being assembly based and there was a big improvement. I could re-run the script in CQ-editor as many times as I wanted without any crashes. Being able to group things as assemblies and subassemblies now is really great.

How hard would it be to add color and alpha support? One part in my assembly fits inside of another one, so I had been using alpha to be able to see the positioning. It's also nice to have some differentiation of parts by color when there are a lot of parts in the assembly.

adam-urbanczyk:
I want add color support to the assembly class - it is supported by XDE (the underlying OpenCascade module: https://dev.opencascade.org/doc/overview/html/occt_user_guides__xde.html#occt_xde_2_6). To my understanding it will just render once it is defined, but we'll see once we get there.

adam-urbanczyk:
Yup, seems to be working. For now you can use it as follows:
```python
assy.add(s1, color=cq.Color("red"))
assy.add(s2, color=cq.Color(1, 0, 0))
assy.add(s3, color=cq.Color(1, 0, 0, 0.5))
```

obraz

jmwright:
Color works great, thanks!

adam-urbanczyk:
Do you maybe have some nice model to show in the docs / readme?

jmwright:
> Do you maybe have some nice model to show in the docs / readme?

Not right now, no. The script that I converted is for a mathematically generated tensegrity structure. It's not finished, so it just looks like a point cloud overlaid on a reference torus. The points are the connectivity nodes.

cqparts has a you car model they use as an example. I could recreate that unless you want a more sophisticated assembly.

https://cqparts.github.io/cqparts/doc/tutorials/assembly.html

adam-urbanczyk:
@Shimwell I think this PR will be very interesting for you.

Shimwell:
This looks most useful indeed.

marcus7070:
> Do you maybe have some nice model to show in the docs / readme?

 I can provide this one if you want. It doesn't use this fork yet, but that should be trivial.

screenshot2020-09-11-125732
screenshot2020-09-11-132428
screenshot2020-09-11-132240
screenshot2020-09-11-132152


It's the spindle off my CNC router. A vacuum hose is permantly attached to the black part. This is held on to the v-slot extrusion with magnets, so it can be quickly removed to change tools.

The model is an assembly of 6 parts. The relationships between parts are simple offsets and 90 degree rotations.

I am a bit worried that it's not obvious enough for docs. We shouldn't waste words explaining parts of my CNC router in the CadQuery docs. It's a bit difficult for me to step back and say what's obvious and what needs explanation though, so please let me know if you think it's suitable or not.

jmwright:
@marcus7070 Very cool! I assume the vacuum hose adapter is 3D printed? Great idea to attach it with magnets to make the tool changes easier.

I think a screenshot of this assembly would be great in the docs. I don't think we need to explain what it is or how it fits together, it would just show users that you can, in fact, make a real world assembly with this new CadQuery feature.

adam-urbanczyk:
It looks very nice @marcus7070 . The idea was indeed to just have an attractive picture to advertise this new feature - so no need to explain. Do you have the code somewhere?

marcus7070:
> @marcus7070 Very cool! I assume the vacuum hose adapter is 3D printed?

That's right. It will go through a few more iterations before I'm statisfied. I'm trying out some new ideas so it looks a bit different to the average dust boot.

@adam-urbanczyk, the code is [here](https://github.com/marcus7070/spindle-assy-example).

adam-urbanczyk:
Preliminary constraint solver seems to be working:
obraz
```python
b1 = cq.Workplane().box(1, 1, 2).faces('>Z').edges('>X').chamfer(0.4)
b2 = cq.Workplane().box(.1, 2, 1)
b3 = cq.Workplane().box(2, .1, .5)
b4 = cq.Workplane().box(1, 1, .2)

assy = cq.Assembly(b1, loc=cq.Location(cq.Vector(1, 1, 0)), name="TOP")
assy2 = cq.Assembly(b2, loc=cq.Location(cq.Vector(0, -3, 4)), name="SECOND")
assy3 = cq.Assembly(b3, loc=cq.Location(cq.Vector(0, 0, 5)), name="THIRD", color=cq.Color('orange'))

assy.add(assy2, color=cq.Color("green"))
assy.add(assy3)
assy.add(b4, loc=cq.Location(cq.Vector(0, 0, 5)), name="4th",color=cq.Color("blue1"))


assy.constrain("TOP@faces@>(1,0,1)", "THIRD@faces@>Z", "Plane")
assy.constrain("TOP@faces@<Y", "SECOND@faces@<Y", "Axis")
assy.constrain("SECOND@faces@>Z", "THIRD@faces@<Z", "Axis")
assy.constrain("THIRD@faces@<X", "SECOND@faces@>X", "Plane")
assy.constrain("4th@faces@>Z", "SECOND@faces@<Z", "Plane")
assy.constrain("4th@faces@>Y", "TOP@faces@<Y", "Axis")
assy.solve()

show_object(assy)
```

jmwright:
@adam-urbanczyk That's great!

adam-urbanczyk:
Any thoughts on the syntax? I think that having a pure string query will fit into the CQ mindset. There is of course additional overload if you want to hand roll a selector (see below). Another addition (for later) would be to have syntax for referencing tagged objects, maybe: `TOP@tag` or `TOP@?tag`.

```python
    @overload
    def constrain(
        self,
        id1: str,
        s1: Shape,
        id2: str,
        s2: Shape,
        kind: ConstraintKinds,
        param: Any = None,
    ) -> "Assembly":
        ...
```

michaelgale:
I like the string based syntax.  The only thing I might suggest is the ability to add constraints from list of constraints?  The constrain method could check if the incoming argument is a list and conveniently loop over them.  It's a minor thing, but I could see myself having a constraint list for every sub-assembly since there is often more than one anchor or constraint to satisfy in complicated assemblies.

marcus7070:
> Any thoughts on the syntax?

Looks neat to me. I appreciate you including the object based overload, I get more use out of those styles than string based selectors because I can't function without autocomplete in my IDE. :sweat_smile:

> Another addition (for later) would be to have syntax for referencing tagged objects

Awesome. For those who don't use tags, imagine a double sided PCB with 4 mounting holes. Selecting those holes around all the components hanging off the board would be difficult and fragile. However, if you start your model with a box for the PCB board, made 4 holes then tagged that state, your selector would be much more simple.

jmwright:
> Any thoughts on the syntax?

I like it, but I want some more time using it. I'll post something tomorrow (Friday) morning (US time UTC-5).

jmwright:
Looks good. I took part of the trivial toy car example that cqparts has and recreated it. It worked well, and the parser even seems to be able to handle combining string selectors.
```python
import cadquery as cq

axle_diameter = 10
length = 50
diameter = 30
width = 10

# Wheel
wheel = cq.Workplane('XY') \
            .circle(diameter / 2.0).extrude(width)
hole = cq.Workplane('XY') \
    .circle(2).extrude(width / 2.0).faces(">Z") \
    .circle(4).extrude(width/2)
wheel = wheel.cut(hole)

# Axle
axle = cq.Workplane('ZX', origin=(0, -length/2, 0)) \
            .circle(axle_diameter / 2).extrude(length)
cutout = cq.Workplane('ZX', origin=(0, -length/2, 0)) \
    .circle(1.5).extrude(10)
axle = axle.cut(cutout)

# Add components of assembly
axle_assy = cq.Assembly(axle, loc=cq.Location(cq.Vector(0, 0, 0)), name="AXLE")
wheel1_assy = cq.Assembly(wheel, loc=cq.Location(cq.Vector(0, 0, 0)), name="WHEEL1")
wheel2_assy = cq.Assembly(wheel, loc=cq.Location(cq.Vector(0, 0, 0)), name="WHEEL2")
axle_assy.add(wheel1_assy, color=cq.Color("black"))
axle_assy.add(wheel2_assy, color=cq.Color("black"))

# Constraints
axle_assy.constrain("WHEEL1@faces@<Z", "AXLE@faces@>Y and <Z", "Plane") # Just testing selector combining
axle_assy.constrain("WHEEL1@faces@<Z", "AXLE@faces@>Y", "Axis")
axle_assy.constrain("WHEEL2@faces@<Z", "AXLE@faces@<Y", "Plane")
axle_assy.constrain("WHEEL2@faces@<Z", "AXLE@faces@<Y", "Axis")
axle_assy.solve()

show_object(axle_assy)
```

Screenshot_20200918_072408

I like it. Thanks @adam-urbanczyk this is great! Is there anything left to add, or is this pretty much ready to merge?

dcowden:
Wow this is amazing!

adam-urbanczyk:
Almost done @jmwright,  but still want to tweak the solver a little bit.

bernhard-42:
I ported jupyter-cadquery over to OCP and now try to find out how I could support the new cadquery assemblies.

After looking at the code I figured that for nested assemblies I have to apply to a shape the product of its assembly's location with the locations of all assemblies being hierarchically above, e.g. for @adam-urbanczyk's example from above:

```python
a = assy.solve()
a0 = a.children[0]
a1 = a.children[1]
a2 = a.children[2]

from jupyter_cadquery.occ import show, Part, Assembly

show(Assembly([
    Part(a.shapes[0].located(a.loc).wrapped, "a"),
    Part(a0.shapes[0].located(a.loc * a0.loc).wrapped, "a0"),
    Part(a1.shapes[0].located(a.loc * a1.loc).wrapped, "a1"),
    Part(a2.shapes[0].located(a.loc * a2.loc).wrapped, "a2"),
]))
```

which seems to produce the right result:

image

Is this the correct way to retrieve all shapes from an assembly and apply the location calculated by the solver to visualise it?

bernhard-42:
When I look at the above examples, `cq.Location(cq.Vector(x, y, z))` appears quite often. Would it be possible to support short cuts like

```python
assy3 = cq.Assembly(b3, loc=(0, 0, 5), name="THIRD", color='orange')
```

and let the `__init__` function create the correct `Location` and `Color` object.

bernhard-42:
This is how I would visualize the example in jupyter-cadquery:
I see one TOP assembly and 3 sub assemblies (SECOND, THIRD and 4th). And every assembly can have several objects (object_0, object_1, ...):

image

From how it is built I would have seen a difference between SECOND/THIRD and 4th: SECOND and THIRD are assemblies itself, while 4th is a shape. I would have expected 4th to be on the same level as the shapes of TOP.
So it looks like

```python
assy = cq.Assembly(b1, loc=L(1, 1, 0), name="TOP")
assy2 = cq.Assembly(b2, loc=L(0, -3, 4), name="SECOND")
assy3 = cq.Assembly(b3, loc=L(0, 0, 5), name="THIRD", color=C('orange'))

assy.add(assy2, color=C("green"))
assy.add(assy3)
assy.add(b4, loc=L(0, 0, 5), name="4th",color=C("blue1"))
```

```python
assy = cq.Assembly(b1, loc=L(1, 1, 0), name="TOP")
assy2 = cq.Assembly(b2, loc=L(0, -3, 4), name="SECOND")
assy3 = cq.Assembly(b3, loc=L(0, 0, 5), name="THIRD", color=C('orange'))
assy4 = cq.Assembly(b4, loc=L(0, 0, 5), name="4th",color=C("blue1"))

assy.add(assy2, color=C("green"))
assy.add(assy3)
assy.add(assy4)
```

lead to identical results, which I found a little bit confusing (btw. `C` and `L` are just shortcut functions).

bernhard-42:
On a completely different topic, I think I have now stared for two hours into the one example above, commenting out constraints and visualising constraints and results - and I must admit, I don't understand it. Constraint based assembly is totally new to me. Is there any documentation around this?

bernhard-42:
btw. converting from cadquery assemblies to jupyter-cadquery assemblies is straightforward (however, one needs to use the Assembly and Part classes from the occ version of juypter-cadquery - need to check how to properly integrate later)

```python
from jupyter_cadquery.occ import Part, Assembly

def rgb(assy):
    def b(x):
        return int(255*x)
    
    if assy.color is None:
        return "#aaa"
    rgb = assy.color.wrapped.GetRGB()
    return "#%02x%02x%02x" % (b(rgb.Red()), b(rgb.Green()), b(rgb.Blue()))

def convert(assy, loc=None):
    loc = assy.loc if loc is None else loc * assy.loc
    color = rgb(assy)
    parent = [Part(shape.located(loc).wrapped, "object_%d" % (i), color=color) for i, shape in enumerate(assy.shapes)]
    children = [convert(c, loc) for c in assy.children]
    return Assembly(parent + children, assy.name)
```

@adam-urbanczyk Does the convert function look right for you?

adam-urbanczyk:
Thanks @bernhard-42 , looks cool! You are right regarding the product of locations.  The semantics of the Assembly class is essentially contained in this snippet:
```python
class Assembly(object):
    loc: Location
    name: str
    color: Optional[Color]
    metadata: Dict[str, Any]

    obj: Union[Shape, Workplane, None]
    parent: Optional["Assembly"]
    children: List["Assembly"]
```

So every child of an `Assembly` is an `Assembly` object. It may or may not contain a  `cq.Workplane` or `cq.Shape`. In the view of this `THIRD` and `4th` are actually equivalent.

Regarding constraints, this is actually a standard feature of CAD software. I have not seen it yet in scripted CAD packages (not that there are so many). For now it is highly experimental, may break and change at any time kind of thing. So no docs at the moment.

Convert looks good to me.

bernhard-42:
Thanks, Adam

bernhard-42:
I did my homework about constraint based assemblies looking at onshape and the Assembly3 FreeCad plugin. 
I think I get the idea, so I tried to build something and selected, inspired by cqparts, a bearing:

```python
def ring(inner_radius, outer_radius, width):
    ring = (cq.Workplane(origin=(0, 0, -width / 2))
        .circle(outer_radius).circle(inner_radius)
        .extrude(width)
    )
    return ring

tol = 0.05
ball_diam = 5

r1, r2, r3, r4 = 4, 6, 8, 10
r5 = (r3 + r2) / 2
inner_ring = ring(r1, r2, ball_diam)
outer_ring = ring(r3, r4, ball_diam)

torus = cq.CQ(cq.Solid.makeTorus(r5, ball_diam / 2 + tol))
ball = cq.Workplane().sphere(ball_diam / 2)

inner = inner_ring.cut(torus)
outer = outer_ring.cut(torus)
```

For the issue I had with this, see https://github.com/CadQuery/cadquery/issues/458

After having done all the steps below, I am not sure this is the best example for the assembly feature. However, you asked for some feedback on the API and so I thought to share some of my experiences:

- The API is clear and I like that it reuses the selector syntax. For the constraints I think I personally would prefer ("qos" meaning "query_or_shape"):

    ```python
    def constrain(
        self, 
        id1: str,
        qos1: Union[str, Shape], 
        id2: str,
        qos2: Union[str, Shape], 
        kind: ConstraintKinds, 
        param: Any = None
    ) -> "Assembly"  
    ```

    allowing to mix the two methods 

    ```python
    constrain("name1", "faces@>Z", "name2", "faces@>Y", kind)
    constrain("name1", "faces@>Z", "name2", shape2, kind)
    constrain("name1", shape1, "name2", "faces@>Y", kind)
    constrain("name1", shape1, "name2", shape2, kind)
    ```

    and to me `faces(">Y")` and `"faces@>Y"` would be a better visual match than `"name@faces@>Y"`

- With the query syntax I haven't found a way to use points that are no vertices (e.g. center of a shape, middle points of an edge, ...). But that's maybe due to my lack of knowledge around selectors.

- The "Plane" constraint seems to always place the two faces onto each other with the shapes being stacked. How would I assemble two shapes with both faces in one plane sharing the center of the faces and having their normals showing into the same direction? I had to introduce a helper object to achieve this. 

- My last example converged (after increasing `maxiter` in the code), however the result is not exactly correct. What is the best way to avoid this? 

Finally, I don't know whether the below approach using helper objects in assemblies is appropriate, but at least it worked. And I am happy to learn, how to avoid them if possible.


**The Bearing assembly**

For the way I wanted to create the constraints, I needed a way to create a constraint for the center of a shape and to
 get all results of an assembly query. So I built two helpers:

```python
def center(assy, name):
    return cq.Vertex.makeVertex(*assy.objects[name].obj.val().Center().toTuple())

def query_all(assy, q):
    name, kind, arg = q.split("@")

    tmp = cq.Workplane()
    obj = assy.objects[name].obj

    if isinstance(obj, (cq.Workplane, cq.Shape)):
        tmp.add(obj)
        res = getattr(tmp, kind)(arg)

    return res.objects if isinstance(res.val(), cq.Shape) else None
```

Given this, I could create the assembly and add the constraints by additionally adding two construction objects:
- `_center`: a cylinder to use with a "Plane" constraint to arrange the rings concentric on one plane (I didn't find another way to assemble two shapes with their faces on the same plane showing into the same direction and sharing the center of the faces).
- `_points`: a polygon on which upper vertices the ball centers would be placed.

```python
def balls(i):
    return "ball_%d" % i

number_balls = 5

# components
assy = cq.Assembly(outer, loc=L(0, 0, ball_diam/2), name="outer", color=C("orange"))
assy.add(inner, loc=L(20, 0, 0), name="inner", color=C("orange"))
for i in range(number_balls):
    assy.add(ball, loc=L(6*i, 20, 0), name=balls(i), color=C("black"))

# helpers
assy.add(cq.Workplane().circle(1).extrude(1), loc=L(0,-20,0), name="_center")
assy.add(cq.Workplane().polygon(number_balls, 2*r5).extrude(ball_diam/2), loc=L(20,-20,0), name="_points")

# contraints
points = query_all(assy, "_points@vertices@>Z")
cs = [
    ("outer@faces@<Z", "_center@faces@>Z", "Plane"),
    ("inner@faces@<Z", "_center@faces@>Z", "Plane"),
    ("_points@faces@<Z", "_center@faces@>Z", "Plane"),
] + [("_points", points[i],  
      balls(i), center(assy, balls(i)),  
      "Point") 
     for i in range(number_balls)]

for c in cs:
    assy.constrain(*c)
```

This is the assembly before solving:

image

And this is the result after solving:

image

However, when I added a 6th ball, the optimzation algorithm did not finish after after 1000 iterations. So I increased the max iterations to 10000. It finally converged

```
   N    Tit     Tnf  Tnint  Skip  Nact     Projg        F
   60   5274   5814      2     0     0   7.478D-05   2.777D-05
  F =  2.77707271466398639E-005

CONVERGENCE: REL_REDUCTION_OF_F_<=_FACTR*EPSMCH
```

but the result was not exact:

image

bernhard-42:
Even for 5 balls, the result is not exact:

The assembled center coordinates (after solving):

```python
loc = assy.loc
for i in range(number_balls):
    o = assy.objects[balls(i)]
    v = o.shapes[0].located(loc * o.loc).Center()
    print("%7.4f, %7.4f, %7.4f" % (v.x, v.y, v.z))

#  6.9230, -0.9758,  2.5239
#  3.0724,  6.2966,  2.4826
# -5.0340,  4.8817,  2.4654
# -6.1935, -3.2651,  2.4959
#  1.1963, -6.8853,  2.5321
```

The coordinates of the polygon vertices (also after solving)

```python
["%7.4f, %7.4f, %7.4f" % (v.X, v.Y, v.Z) for v in query_all(assy, "_points@vertices@>Z")]
# [' 7.0000,  0.0000,  2.5000',
#  ' 2.1631,  6.6574,  2.5000',
#  '-5.6631,  4.1145,  2.5000',
#  '-5.6631, -4.1145,  2.5000',
#  ' 2.1631, -6.6574,  2.5000']
```

I had expected that the "Point" constraint ensures that the points are colocated. 
Did I dow something wrong with my constraint?

adam-urbanczyk:
Thanks for the suggestions regarding the API. I think we can add those in future iterations of this functionality.

I changed the cost function slightly @bernhard-42 , seems to give better result now. Pleas pull and try. In general I use a numerical solver here so nothing is "exact".

obraz


BTW: please share complete code examples so that I can reproduce next time, it took me a while to glue the snippets together.

bernhard-42:
@adam-urbanczyk The results are much better now. The issue before the change was that the solver did not align the normals well enough. To compare I created a little metrics function. For the 6 ball bearing I get
```
Constraint                                 Normal-Dist    Point-Dist  
--------------------------------------------------------------------
Plane outer@faces@<Z - _center@faces@>Z      0.0001571     0.0000001
Plane inner@faces@<Z - _center@faces@>Z      0.0001115     0.0000000
Plane _points@faces@<Z - _center@faces@>Z    0.0000002     0.0000001
Point _points<Vertex> - ball_0<Vertex>                     0.0000002
Point _points<Vertex> - ball_1<Vertex>                     0.0000001
Point _points<Vertex> - ball_2<Vertex>                     0.0000001
Point _points<Vertex> - ball_3<Vertex>                     0.0000002
Point _points<Vertex> - ball_4<Vertex>                     0.0000002
Point _points<Vertex> - ball_5<Vertex>                     0.0000001
```
with `Point-Dist` being `norm(center1-center2)` and `Normal-Dist` being  `norm(normal1 - (-normal2))` after solving.

Before `Normal-Dist` was up to 0.017, so for my tests the result is more than factor 100 better now - and the algorithm is much faster for this example.

All code including the metrics function can be found in https://github.com/bernhard-42/jupyter-cadquery/blob/master/examples/cadquery-assemblies.ipynb

adam-urbanczyk:
Thanks @bernhard-42 . FYI I implemented parameter handling for `Axis` and `Point` - you can now specify the distance (default is 0) or angle (default is pi). Note that it can lead to non-unique solutions resulting in a funny behavior. That's maybe a story for another PR or maybe just a fact of life.

adam-urbanczyk:
Still need to add the pic, but I don't want touch the code. I think yo ucan start reviewing @jmwright .
obraz

cadquery/assembly.py
```diff
@@ -0,0 +1,415 @@
...
+            subassy.loc = kwargs["loc"] if kwargs.get("loc") else arg.loc
+            subassy.name = kwargs["name"] if kwargs.get("name") else arg.name
+            subassy.color = kwargs["color"] if kwargs.get("color") else arg.color
+            subassy.parent = self
+
+            self.children.append(subassy)
+            self.objects[subassy.name] = subassy
+            self.objects.update(subassy.objects)
+
+            arg.parent = self
```
jmwright, line 221: Why is the parent of arg set here when a copy of arg was made above and already had its parent set?

cadquery/occ_impl/shapes.py
```diff
@@ -314,8 +314,7 @@ def fix(obj: TopoDS_Shape) -> TopoDS_Shape:
 
 class Shape(object):
     """
-        Represents a shape in the system.
-        Wrappers the FreeCAD apiSh
+    Represents a shape in the system.Wraps TopoDS_Shape.
```
jmwright, line 317: I don't understand what `system.Wraps` means here.

jmwright: Oh, it's the end of a sentence without a space. I was reading it as a class.method.

cadquery/occ_impl/solver.py
```diff
@@ -0,0 +1,225 @@
...
+
+        x0 = array([el for el in self.entities]).ravel()
+        f, jac = self._cost()
+
+        res = minimize(
+            f,
+            x0,
+            jac=jac,
+            method="BFGS",
+            options=dict(disp=True, ftol=1e-12, maxiter=1000),
```
jmwright, line 217: When I run the tests I get the following warning about `ftol`.
```
tests/test_assembly.py::test_constrain
  /home/jwright/Downloads/repos/cadquery/cadquery/occ_impl/solver.py:217: OptimizeWarning: Unknown solver options: ftol
    options=dict(disp=True, ftol=1e-12, maxiter=1000),
```

environment.yml
```diff
@@ -8,15 +8,18 @@ dependencies:
   - ipython
   - ocp
   - pyparsing
-  - sphinx=2.4
+  - sphinx=3.2.1
```
jmwright, line 11: We'll want to make sure that this version plays nice with the the 3D views in the docs.

mypy.ini
```diff
@@ -9,3 +9,12 @@ ignore_missing_imports = True
...
 ignore_missing_imports = True
+
+[mypy-scipy.*]
+ignore_missing_imports = True
+
+[mypy-numpy.*]
+ignore_missing_imports = True
+
+[mypy-nptyping.*]
+ignore_missing_imports = True
```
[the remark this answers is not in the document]
jmwright: Bump

jmwright (approved):
@adam-urbanczyk I think this looks good. The warning about the ftol parameter is the only thing I see that might be introducing a bug or unexpected behavior. All the rest are minor things.

environment.yml
adam-urbanczyk: The docs build now in CI and it worked fine.

cadquery/occ_impl/solver.py
adam-urbanczyk: Solved

cadquery/assembly.py
adam-urbanczyk: Good catch, arg should be untouched!

adam-urbanczyk:
OK this is it, I think. You can merge @jmwright if everything looks good.

Thanks to all for feedback and testing, thanks to @marcus7070 for the cool spindle model and to @bernhard-42 for the bearing assy that allowed me to improve the solver.

jmwright:
Thanks for all the work you've put into this @adam-urbanczyk !

Shimwell:
Congrats on this really really nice feature

Issue #451, opened 2020-09-20
cadquery.Workplane.rarray arg handling implemented incorrectly.


JosiahBradley:
https://github.com/CadQuery/cadquery/blob/2d721d0ff8a195a0902eb9c3add88d07546c05b1/cadquery/cq.py#L1294

According to https://cadquery.readthedocs.io/en/latest/classreference.html#cadquery.Workplane.rarray
rarray's arguments need to be greater than 0. The implementation correctly does this for the integer arguments xCount and yCount, however xSpacing and ySpacing are floats and their check should be if arg less than or equal to 0
This simple code reproduces the error:

```
import cadquery as cq
cq.Workplane().box(5.0, 5.0, 5.0).faces(">Z").rarray(1.0, 0.625, 1, 3).hole(0.125)
```

Recommended Solution
```
if xSpacing <= 0 or ySpacing <= 0 or xCount < 1 or yCount < 1:
            raise ValueError("Spacing and count must be > 0 ")
```

jmwright:
@JosiahBradley Good catch. Would you be interested in creating a PR for this?

PR #454, opened 2020-09-22
rarray: bound spacings to greater than 0

JosiahBradley:
This addresses Issue #451  Instead of matching the conditional for the counts I left them alone as the data types are different.


adam-urbanczyk:
Thanks @JosiahBradley , could you rebase to get the latest Travis conf?

adam-urbanczyk:
Thanks for the fix @JosiahBradley , I just added a simple test.

PR #455, opened 2020-09-25
Fix dxf arc export for some partial circles

snagy:
Hey folks!

I'm not sure if this is the 'right' fix, but some partial circles (specifically, the ones with c_dz.XYZ().Z() < 0) were exporting flipped for me.

The changes are pretty minor, and fixed all the broken arcs on my data.  Basically, rotate the arc 180 degrees.  I also flipped the start/end and used the is_counterclockwise flag for those arcs.

Broken:
Screen Shot 2020-09-24 at 11 41 26 PM

Fixed:
Screen Shot 2020-09-24 at 11 41 57 PM

### DIFF
--- a/cadquery/occ_impl/exporters/dxf.py
+++ b/cadquery/occ_impl/exporters/dxf.py
@@ -14,8 +14,7 @@ CURVE_TOLERANCE = 1e-9
 def _dxf_line(e, msp, plane):
 
     msp.add_line(
-        e.startPoint().toTuple(),
-        e.endPoint().toTuple(),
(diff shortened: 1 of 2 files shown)


adam-urbanczyk:
Thanks a lot @snagy !  Could you fix the formatting (black is complaining in CI)? It would be also great if you could share some code that triggers this issue - maybe we could add it to the tests.

snagy:
@adam-urbanczyk done and done!

adam-urbanczyk (approved):

adam-urbanczyk:
Looks good, thanks @snagy !

PR #460, opened 2020-09-27
Nth selector always returns empty list

cactrot:
For a call like:
workplane.faces("<Y[1] or <Y[0]")

I would always get "list index out of range".

It appears that there was an extra call to the superclass filter, removing it seems to make it work per my understanding of the documentation.

Also the calculation for the number of digits to round self.TOLERANCE was definitely incorrect.


adam-urbanczyk:
Thanks for the PR but could you open an issue first and paste a minimal reproducible example? If I try the following code:
```python
import cadquery as cq
r = cq.Workplane().box(1,1,1).faces("<Y[1] or <Y[0]")
```
I get the correct output:
obraz

cactrot:
Sorry about the excerpt only, I'll have to work on making a minimal example I can share. Also I need to figure out how to setup an environment to run the tests, so I can see what works without posting each change here.

cactrot:
Oh, I didn't realize >Y implied |Y. I thought it was based on either center of mass or vertices of bounding box. I'll reopen just the rounding issue then.

PR #461, opened 2020-09-27
Nth selector rounding

cactrot:
I am almost certain that the number of digits to round to for self.TOLERANCE = 0.0001 should be 4 rather than 10000.

### DIFF
--- a/cadquery/occ_impl/shapes.py
+++ b/cadquery/occ_impl/shapes.py
@@ -140,6 +140,7 @@ from OCP.Font import (
     Font_FA_Regular,
     Font_FA_Italic,
     Font_FA_Bold,
+    Font_SystemFont,
 )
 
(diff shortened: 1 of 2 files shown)


cactrot:
That is weird that most platforms passed, including several versions of windows but not all.

I don't think the test that failed should even depend on DirectionNthSelector

adam-urbanczyk:
Let's see of it reproduces.

cactrot:
The coverage changes it reports are in cq.py and test_cadquery.py. I only edited one line in selectors.py.

https://codecov.io/gh/CadQuery/cadquery/pull/461/changes

It looks like codecov is failing because of other changes happening in the repo.

adam-urbanczyk:
Looks good @cactrot , thanks! Codecov sometimes gives false positives.

PR #475, opened 2020-10-12
Implement locations() for Wires

adam-urbanczyk:
Implements `locations()` and `locationAt()` for both Wires and Edges.

### DIFF
--- a/cadquery/occ_impl/shapes.py
+++ b/cadquery/occ_impl/shapes.py
@@ -918,10 +918,18 @@ class Vertex(Shape):
 
 
 class Mixin1DProtocol(ShapeProtocol, Protocol):
-    def _geomAdaptor(self) -> Adaptor3d_Curve:
+    def _geomAdaptor(self) -> Union[BRepAdaptor_Curve, BRepAdaptor_CompCurve]:
         ...
 
-    def _geomAdaptorH(self) -> Adaptor3d_HCurve:
+    def _geomAdaptorH(
+        self,
+    ) -> Tuple[
+        Union[BRepAdaptor_Curve, BRepAdaptor_CompCurve],
+        Union[BRepAdaptor_HCurve, BRepAdaptor_HCompCurve],
+    ]:
+        ...
+
+    def paramAt(self, d: float) -> float:
         ...
 
     def locationAt(
@@ -934,12 +942,14 @@ class Mixin1DProtocol(ShapeProtocol, Protocol):
 
 
 class Mixin1D(object):
-    def _geomAdaptor(self: Mixin1DProtocol) -> Adaptor3d_Curve:
+    def _geomAdaptor(
+        self: Mixin1DProtocol,
+    ) -> Union[BRepAdaptor_Curve, BRepAdaptor_CompCurve]:
         """
         Return the underlying geometry
         """
 
-        curve: Adaptor3d_Curve
+        curve: Union[BRepAdaptor_Curve, BRepAdaptor_CompCurve]
 
         if isinstance(self.wrapped, TopoDS_Edge):
             curve = BRepAdaptor_Curve(self.wrapped)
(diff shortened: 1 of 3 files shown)


trayracing:
Giving locations on wires would be a great help on placing objects around the perimeter of a face. Two comments:
1) There are three potentially interesting orientations to return for the points: tangential, locally perpendicular to the enclosed area, and perpendicular to the path, but locally in plane. (see cadquery google group for examples.)
2) In the PR, the orientation is not consistent around the wire, and instead appears to be per edge. This makes it harder to use. [Example image] from cadquery group. (https://groups.google.com/group/cadquery/attach/2f33032517546/Auto%20Generated%20Inline%20Image%201?part=0.1&view=1&authuser=0)

trayracing:
> In the PR, the orientation is not consistent ...
As Adam pointed out in the google cadquery group, using the `frame="corrected"` parameter corrects this flipping. Apologies for the distraction.

marcus7070:
`positionAt` looks great, I'm going to have a lot of use for that, thanks @adam-urbanczyk.

cadquery/occ_impl/shapes.py
```diff
@@ -959,7 +976,14 @@ def endPoint(self) -> Vector:
 
         return Vector(curve.Value(umax))
 
-    def tangentAt(self, locationParam: float = 0.5) -> Vector:
+    def paramAt(self: Mixin1DProtocol, d: float) -> float:
+
```
marcus7070, line 980: Can I suggest a doc string for this one? It was difficult to figure out what this does.
```
Compute length along the edge at the specified location.
:param d: location to use in [0, 1]
:return: length
```

adam-urbanczyk: I added a docstring, but note that it actually does something subtly different.

adam-urbanczyk:
@jmwright if everything passes I'd say it is ready for merging.

cadquery/occ_impl/shapes.py
marcus7070: Ah, now I think I understand. I wasn't clear on the definition of "parameter". For anyone else reading this, here is the definition of "parameter" for a circle in OCCT: https://old.opencascade.com/doc/occt-7.4.0/refman/html/class_geom___circle.html#details

tests/test_cadquery.py
```diff
@@ -3622,6 +3636,42 @@ def testNormal(self):
...
+
+        self.assertTupleAlmostEquals(n.toTuple(), (0, 0, 1), 6)
+
         with self.assertRaises(ValueError):
             edge = Workplane().rect(1, 2).edges().val()
             n = edge.normal()
+
+    def testPositionAt(self):
+
+        w = Workplane().lineTo(0, 1).lineTo(1, 1).wire().val()
```
jmwright, line 3650: I'm assuming it doesn't matter if this wire is closed, but thought I'd check to be sure.

jmwright (approved):
@adam-urbanczyk Thanks for doing this. I just have a few minor comments.

marcus7070:
Currently `Mixin1DProtocol` is not autosummarized in the HTML docs. It also looks like sphinx is set up to not autosummarize inherited methods. So while the methods of the `Edge` class appear in there, these new ones on `Mixin1DProtocol` do not.

Perhaps that's intentional, but I thought I'd point it out incase it's an oversight.

adam-urbanczyk:
> Perhaps that's intentional, but I thought I'd point it out incase it's an oversight.

Good catch, it was not the intention. I'll add `Mixin1D` and other mixins to the docs.

tests/test_cadquery.py
adam-urbanczyk: Good point, I extended the test with a closed wire case.

adam-urbanczyk:
I tried to add `Mixin1D` and `Mixin3D` to the docs, but it does not show up (due to my poor sphinx-fu). Does anyone know how to do it properly?

jmwright:
@Peque Do you have any ideas on the Spinx part of this (see previous comment from @adam-urbanczyk )? You've done quite a bit of work with Sphinx.

Peque:
@jmwright No idea, sorry. :sweat_smile: I may be able to have a look at it if you are still blocked in 1-2 weeks. Sounds like a good question for StackOverflow though. :stuck_out_tongue_winking_eye:

adam-urbanczyk:
It works now, but could be prettier. I propose to merge and maybe investigate autodoc alternatives in the future. @jmwright would you agree?

jmwright:
+1 to merge

PR #477, opened 2020-10-12
Fixed parametricCurve() to use correct stop value

just-georgeb:
parametricCurve() takes a start and stop value as inputs:
```
:param start: starting value of the parameter t
:param stop: final value of the parameter t
```
The final value in the current implementation is actually `start + stop`. This PR fixes it to end at the specified stop value


adam-urbanczyk:
Thanks @just-georgeb  , good catch!

Issue #149, opened 2019-06-13
Workplane() resets currently selected objects


adam-urbanczyk:
The following code does not work
```python
import cadquery as cq
result = cq.Workplane("XY").rect(10.0,10.0).extrude(10).faces("<Y")\
    .workplane().vertices('<XZ').circle(1).extrude(10)
```
I think it should -- it is quite clear what the intent if the user is here. This is also related to #24

Jojain:
We agree that the expected behaviour would be equal to this : 
```python
import cadquery as cq
result = cq.Workplane("XY").rect(10.0,10.0).extrude(10).faces("<Y")\
    .workplane().end(1).vertices('<XZ').circle(1).extrude(10)```
I have been bothered several time by this and while the `end` method solves the problem, it's super unnatural and definitely not novice friendly.

If this is still expected to be solved I will try to bring a PR

**Comment by adam-urbanczyk** [comment]
I think, that's it indeed.

**Comment by adam-urbanczyk** [comment]
Or maybe:
```python
s.objects = self.objects[:]
```

Jojain:
I have tried my solution, it breaks some tests, I have quickly looked, some tests could be rewritten, some need a bit more investigation why they break.

adam-urbanczyk:
Feel free to open a WIP PR and we can take a look together.

PR #480, opened 2020-10-13
Create a child workplane on a vertex

### LINKED ISSUE #24: Workplane orientation breaks when "Locating workplane on a vertex".
<a href="https://github.com/sander76"></img></a> **Issue by [sander76](https://github.com/sander76)**
_Saturday Aug 11, 2018 at 10:15 GMT_
_Originally opened as https://github.com/dcowden/cadquery/issues/289_

----

Comparable to the example displayed at http://dcowden.github.io/cadquery/examples.html#locating-a-workplane-on-a-vertex I am trying to draw a cilinder on a cube with the center located at one of the cube's vertices. However when I do that the workplane changes orientation unexpectedly.

Two examples below.
1. with cilinder on cube's face at the center
2. identical to above, but now trying to put it at one of the vertices of the cube.

## Example 1

```
import cadquery as cq

result = cq.Workplane("XY").rect(10.0,10.0).extrude(10)

# Draws a cilinder on the center of the XZ face of the cube
result = result.faces("<Y").workplane().circle(2.0).extrude(10)

show_object(result)
```

image

## Example Two

```
import cadquery as cq

result = cq.Workplane("XY").rect(10.0,10.0).extrude(10)

# Should draw the cilinder on the same face as above but with its center placed
# at the lower XZ coordinate of that face
# But instead the workplane orientation changes too.
result = result.faces("<Y").vertices("<XZ").workplane().circle(2.0).extrude(10)

show_object(result)
```

image

ArmoredBlood:
This PR should resolve #24 and #149 (I think).

This is also my first time coding anything in the CAD domain, so I'm open to criticism on the solution since I'm not very familiar with the domain. I have an intermediate level of knowledge of CAD as a hobby user, not as a programmer.

I'm also new to CadQuery, so the only potential issue I see is this only looks at the immediate parent workplane for a Face to use. I'm not sure if it should instead do some kind of recursive search through the entire stack for a Face (assuming there's a long stack of related workplanes)

### DIFF
--- a/cadquery/cq.py
+++ b/cadquery/cq.py
@@ -543,8 +543,10 @@ class Workplane(object):
                         if isinstance(obj, Shape)
                         else obj.Center()
                     )
-                if self.parent is not None and isinstance(self.parent.val(), Face):
-                    normal = self.parent.val().normalAt(center)
+
+                val = self.parent.val() if self.parent else None
+                if isinstance(val, Face):
+                    normal = val.normalAt(center)
                     xDir = _computeXdir(normal)
                 else:
                     normal = self.plane.zDir
--- a/cadquery/cq.py
+++ b/cadquery/cq.py
@@ -543,7 +543,7 @@ class Workplane(object):
                         if isinstance(obj, Shape)
                         else obj.Center()
                     )
-                if isinstance(self.parent.val(), Face):
+                if self.parent is not None and isinstance(self.parent.val(), Face):
                     normal = self.parent.val().normalAt(center)
                     xDir = _computeXdir(normal)
                 else:
--- a/cadquery/cq.py
+++ b/cadquery/cq.py
@@ -543,8 +543,12 @@ class Workplane(object):
                         if isinstance(obj, Shape)
                         else obj.Center()
                     )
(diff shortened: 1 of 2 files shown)


ArmoredBlood:
I'm not sure what mypy is wanting from me in the travisCI build. My logic should only be ran when the cq object is a Face.

jmwright:
The mypy stuff is still a mystery to me, but the signature in the lint error matches [CQObject](https://github.com/CadQuery/cadquery/blob/c3e9ef548576186598e82b139b4f11ebbc113a1d/cadquery/cq.py#L60).

So maybe mypy at least thinks you're grabbing the wrong level of object? @adam-urbanczyk will know, I'm sure.

adam-urbanczyk (approved):

adam-urbanczyk:
Thanks @ArmoredBlood . Mypy does not now that the result of `val()` has the same type between the two calls.

Looks good to me, @jmwright any thoughts?

jmwright:
+1 to merge

jmwright:
Thanks for the contribution @ArmoredBlood

PR #481, opened 2020-10-13
Adding docker publish action

Shimwell:
Just a thought but I'm wondering if the dockerfile that @jmwright made could be automatically build and published on dockerhub each time a push to master branch is made.

There is a Github action for automating such a task https://github.com/docker/build-push-action

This PR would need to be accompanied with some new github repository secrets that grant access to the dockerhub account
DOCKERHUB_USERNAME
DOCKERHUB_TOKEN

If another Dockerfile (perhaps with jupyter notebook) gets added then another git action can build that one as well

Co-authored-by: @jmwright

### DIFF
--- a/.github/workflows/dockerhub-publish-dev.yml
+++ b/.github/workflows/dockerhub-publish-dev.yml
@@ -2,7 +2,7 @@ name: dockerhub-publish
 
 on:
   push:
-    branches: develop
+    branches: master
 
 jobs:
   main:
@@ -26,7 +26,7 @@ jobs:
         uses: docker/build-push-action@v2
         with:
           push: true
-          tags: cadquery/cadquery:dev
+          tags: cadquery/cadquery
       -
         name: Image digest
         run: echo ${{ steps.docker_build.outputs.digest }}
--- a/.github/workflows/dockerhub-publish-dev.yml
+++ b/.github/workflows/dockerhub-publish-dev.yml
@@ -1,0 +1,32 @@
+name: dockerhub-publish
+
+on:
+  push:
+    branches: develop
+
+jobs:
+  main:
+    runs-on: ubuntu-latest
+    steps:
+      -
+        name: Set up QEMU
+        uses: docker/setup-qemu-action@v1
+      -
+        name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v1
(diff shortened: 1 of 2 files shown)


jmwright:
@dcowden has mentioned being interested in CadQuery's Docker infrastructure as well.

Shimwell:
I've thrown together an example repo that shows git actions automatically building and publishing dockerfile on pushes to develop, master and with each release. I have added @jmwright @dcowden  @adam-urbanczyk  to the repository in case you want to trigger any dockerhub builds and see the git actions working

https://github.com/Shimwell/example-dockerhub-upload

dcowden:
Thanks for this contribution! I'll have a look!

Over the holidays hope to make an official cqdocker image, this will be a
huge help.

RubenRubens:
Great work!

RubenRubens:
I don't think this is going to work as expected. This pipeline is triggered at the same time as the cq is build and upload to conda. Therefore, docker would be one commit behind master.

Possible fixes:
+ Use Azure and trigger the pipeline after cq is uploaded to conda.
+ Use [this docker image](https://github.com/RubenRubens/cq-testing/blob/master/Dockerfile) instead. It builds cadquery from source inside a container.

I have some other questions:
1. Is it more convenient to upload the image to Dockerhub or to Github Packages?
2. Is it more convenient to use Azure or Github Actions?

jmwright:
> Use this docker image instead. It builds cadquery from source inside a container.

That seems like a fine solution to me.

> Is it more convenient to upload the image to Dockerhub or to Github Packages?

I would expect that Dockerhub would be a more standard location, wouldn't it?

> Is it more convenient to use Azure or Github Actions?

There's no precedent yet for using GitHub Actions on this repo, so I would favor Azure unless there's a compelling reason to do otherwise.

Shimwell:
We should perhaps close this issue as it is going along the wrong track, the dockerfile is not as good as the one made by @RubenRubens and Github Actions is also not used on this repo. So it looks like this PR is not quite suitable. I shall close for now but perhaps we should make a new PR with the improved Dockerfile https://github.com/RubenRubens/cq-testing/blob/master/Dockerfile and Azure to push to dockerhub.

PR #491, opened 2020-10-24
`occ_impl` doc strings

marcus7070:
I've realised that there are a few useful methods in the `occ_impl` layer that I wasn't familar with. So I'm going through and adding docstrings to a few and fixing any inconsistent formatting or other small things I see.

### DIFF
--- a/cadquery/occ_impl/shapes.py
+++ b/cadquery/occ_impl/shapes.py
@@ -790,7 +790,7 @@ class Shape(object):
         of the geometry, but cannot handle skew transformations.
 
         :param tMatrix: The transformation matrix
-        :returns: a copy of the object, but with geometry transformed insetad
+        :returns: a copy of the object, but with geometry transformed instead
             of just rotated.
         """
         r = Shape.cast(
--- a/cadquery/occ_impl/shapes.py
+++ b/cadquery/occ_impl/shapes.py
@@ -383,6 +383,9 @@ class Shape(object):
         return writer.Write(self.wrapped, fileName)
 
     def exportStep(self, fileName: str) -> IFSelect_ReturnStatus:
+        """
+        Export this shape to a STEP file
+        """
 
         writer = STEPControl_Writer()
         writer.Transfer(self.wrapped, STEPControl_AsIs)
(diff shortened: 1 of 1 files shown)


marcus7070:
Before I get too far I thought I should just check, these methods don't have doc strings because no one has got around to it yet, correct? Not because they are supposed to be hidden from users or sphinx? Or because the OCCT docs are supposed to be the only source and we don't want to repeat this stuff in CQ?

marcus7070:
I tried to squash commits and something went horribly wrong. Successfully squashed with that second force push now.

marcus7070:
There might be a problem with sphinx auto type hints. When `self` is type hinted it does this:
screenshot2020-10-24-173549
Is that correct behaviour?

marcus7070:
I'm done for now. If I'm feeling masochistic again I'll try to do some more in another PR.

adam-urbanczyk:
Thanks @marcus7070  - I think it was due to all of the above reasons.

marcus7070:
> I think it was due to all of the above reasons.

Ok then, if there are reasons to not have doc strings on any of these methods then please do remove them from this PR or reject it entirely. I don't mind; I did the work to familiarise myself with the methods, it's not wasted effort.

adam-urbanczyk (approved):

adam-urbanczyk:
Thanks @marcus7070 , I think we should merge it before 2.1. In the end it does not hurt to have better docstrings.

cadquery/occ_impl/shapes.py
```diff
@@ -677,16 +779,19 @@ def transformShape(self, tMatrix: Matrix) -> "Shape":
...
-            WARNING: transformGeometry will sometimes convert lines and circles to splines,
-            but it also has the ability to handle skew and stretching transformations.
+        If your transformation is only translation and rotation, it is safer to
+        use :py:meth:`transformShape`, which doesnt change the underlying type
+        of the geometry, but cannot handle skew transformations.
 
-            If your transformation is only translation and rotation, it is safer to use transformShape,
-            which doesnt change the underlying type of the geometry, but cannot handle skew transformations
+        :param tMatrix: The transformation matrix
+        :returns: a copy of the object, but with geometry transformed insetad
```
jmwright, line 793: instead is misspelled here

jmwright (approved):
Other than my comment, this is ready to merge. Thanks @marcus7070

cadquery/occ_impl/shapes.py
marcus7070: Fixed, thanks.

adam-urbanczyk:
Merging, thanks @marcus7070 .

Issue #493, opened 2020-10-30
Investgate better methods of generating API docs


adam-urbanczyk:
See #175

PR #504, opened 2020-11-13
Radius selector

### LINKED ISSUE #501: Radius selector
I think there is a need for a radius selector. All of CQ's current selectors work of the centre of an object or the bounding box. So if you have the following shape for example:
```python
part = (
    cq
    .Workplane()
    .circle(10)
    .extrude(10)
    .faces(">Z")
    .workplane()
    .hole(10)
)
```
screenshot2020-11-10-201029
and you want to fillet just the outer edges and not the inner, the only selector that you can use is some ugly combination of bounding boxes. 

A radius selector in the style of `DirectionNthSelector` would be a good addition IMHO.

marcus7070:
Will close #501.

---

I've added an `Edge.radius` method to get the radius of a circular edge. 

The method uses `_geomAdaptor().Circle()`. When I used that code on a elipse, I got the kernel (7.4.0) to segfault. So I've restricted the method to just objects with `geomType() == "CIRCLE"` to be safe.

---

I've added a RadiusSelector in the style of DirectionNthSelector. 

I didn't add RadiusSelector to the string selector interface.

While I added it to the HTML docs, about half the selectors are not in the top level module and they don't show up due to #493.

### DIFF
--- a/cadquery/occ_impl/shapes.py
+++ b/cadquery/occ_impl/shapes.py
@@ -190,7 +190,7 @@ from OCP.GeomFill import (
 )
 
 # for catching exceptions
-from OCP.Standard import Standard_NoSuchObject
+from OCP.Standard import Standard_NoSuchObject, Standard_Failure
 
 from math import pi, sqrt
 import warnings
@@ -1069,7 +1069,7 @@ class Mixin1D(object):
         geom = self._geomAdaptor()
         try:
             circ = geom.Circle()
-        except Standard_NoSuchObject as e:
+        except (Standard_NoSuchObject, Standard_Failure) as e:
             raise ValueError("Shape could not be reduced to a circle") from e
         return circ.Radius()
 
--- a/cadquery/occ_impl/shapes.py
+++ b/cadquery/occ_impl/shapes.py
@@ -188,6 +188,10 @@ from OCP.GeomFill import (
     GeomFill_CorrectedFrenet,
     GeomFill_TrihedronLaw,
 )
+
+# for catching exceptions
+from OCP.Standard import Standard_NoSuchObject
+
 from math import pi, sqrt
 import warnings
 
@@ -1052,6 +1056,22 @@ class Mixin1D(object):
     def Length(self: Mixin1DProtocol) -> float:
 
         return GCPnts_AbscissaPoint.Length_s(self._geomAdaptor())
+
+    def radius(self: Mixin1DProtocol) -> float:
+        """
+        Calculate the radius.
+
+        Note that when applied to a Wire, the radius is simply the radius of the first edge.
+
+        :return: radius
+        :raises ValueError: if kernel can not reduce the shape to a circular edge
+        """
+        geom = self._geomAdaptor()
+        try:
+            circ = geom.Circle()
+        except Standard_NoSuchObject as e:
+            raise ValueError("Shape could not be reduced to a circle") from e
+        return circ.Radius()
 
     def IsClosed(self: Mixin1DProtocol) -> bool:
 
--- a/cadquery/selectors.py
+++ b/cadquery/selectors.py
@@ -300,6 +300,51 @@ class TypeSelector(Selector):
         return r
 
 
+class RadiusNthSelector(Selector):
+    """
+    Select the object with the Nth radius.
+
+    Applicability:
+        All Edge and Wires.
+
+    Will ignore any shape that can not be represented as a circle or an arc of
+    a circle.
+    """
+
+    def __init__(self, n, directionMax=True, tolerance=0.0001):
+        self.N = n
+        self.directionMax = directionMax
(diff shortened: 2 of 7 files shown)


jmwright:
Thanks for working on this @marcus7070 

@adam-urbanczyk  The errors with Travis, Appveyor and the doc builds seem to be related to font/text capability. Items like this:
```
ImportError: cannot import name 'Font_BRepTextBuilder'
```
Did something change with the deployed versions of OCP? Maybe related to the 7.5 beta?

adam-urbanczyk:
Indeed @jmwright , should be fine now.

jmwright:
All green now.

@marcus7070 I'm assuming this is ready for review now?

marcus7070:
Yes, ready for review @jmwright.

cadquery/occ_impl/shapes.py
```diff
@@ -1322,6 +1322,17 @@ def makeLine(cls: Type["Edge"], v1: Vector, v2: Vector) -> "Edge":
         """
         return cls(BRepBuilderAPI_MakeEdge(v1.toPnt(), v2.toPnt()).Edge())
 
+    def radius(self) -> float:
```
adam-urbanczyk, line 1325: Could you move this method to the Mixin1D? I think it'll be then automatically usable for wires with one edge.

cadquery/selectors.py
```diff
@@ -300,6 +300,45 @@ def filter(self, objectList):
         return r
 
 
+class RadiusSelector(TypeSelector):
```
adam-urbanczyk, line 303: I'd consider renaming to RadiusNthSelector , but I'm open for discussion.

adam-urbanczyk (changes requested):

cadquery/occ_impl/shapes.py
marcus7070: `.geomType()` of a Wire always seems to be `"Wire"`, so as it is currently written it won't work for a wire with a single `"CIRCLE"` edge.

I could always add the following though:
```python
class Wire():
    ...
    def radius(self):
        radii = [el.radius() for el in self.Edges()]
        if not all(radii):
            raise ValueError('Not a circular wire')
        # calculate average
        # if deviation is above tolerance
            raise ValueError('Wire does not have a consistent radius')
        return avg_radius
```

cadquery/selectors.py
marcus7070: I agree, that's will be more consistent with the other selectors.

cadquery/occ_impl/shapes.py
adam-urbanczyk: True, though you can still call `_geomAdaptor().Circle().Radius()` on a wire with one edge. Maybe catch `Standard_NoSuchObject` and rethrow it as a ValueError instead of checking `geomType`?

marcus7070: I've been avoiding doing that because I get segfaults with certain non-circular edges. Can you recreate this segfault?

```
>>> w0 = cq.Wire.makeEllipse(20, 10, (0, 0, 0), (0, 0, 1), (1, 0, 0))
>>> w0._geomAdaptor().Circle()
zsh: segmentation fault (core dumped)  python
```

The only method I could think of to prevent this segfault is that `geomType() == "CIRCLE"` check. Which sucks, because it's pretty cool that OCCT can create a circle from anything that resembles a circle.

I am still on OCCT 7.4.0 (I'm not using Conda, but I will get the update done soon, especially if it's relevant to this PR).

adam-urbanczyk: I get this with OCCT 7.4:
```python
Standard_NoSuchObject: GeomAdaptor_Curve::Circle() - curve is not a Circle
```

marcus7070: Oh damn, I must have messed something up in my install. I'll move to a VM and Conda for the rest of this PR. Thanks for testing that for me.

I'll move the method over to Mixin1D and raise a ValueError from the OCP exception.

marcus7070:
So I think when you request the radius of a Wire with multiple Edges, OCCT just takes the radius of the first Edge and calls it a day. You can see code regarding this in `test_cad_objects.py`. This was pretty unexpected for me, so I've mentioned it in the `radius` docstring.

It wouldn't be too hard to do some more logic in the `radius` method. We could grab the radius of every edge in a wire, and if they are all within a tolerance return the average value as the radius, otherwise raise ValueError. But I've stuck with the OCCT logic for now. Let me know if you have opinions either way.

adam-urbanczyk (approved):
Thanks, looks good!

adam-urbanczyk:
@jmwright are OK with merging this PR?

jmwright (approved):

jmwright:
Thanks @marcus7070 ! This is a great addition to the selectors.

Issue #469, opened 2020-10-03
Assembly docs


adam-urbanczyk:
Add a docs chapter about using assemblies and constrains.

marcus7070:
Carrying on from #503, I think it would be good to show in the docs how to make a brace that connects between parts in an assembly. The kind of parts where it would be really annoying to calculate yourself where the mounting faces wind up in the final assembly.

1. Add several parts to an assembly and constrain them
2. Solve
3. Select a wire from the mounting face in one part of the assembly
4. Use assy.objects["part_a"].loc to position the mounting face in it's final position
5. Repeat steps 3 & 4 for part_b
6. A simple loft to create the brace.

adam-urbanczyk:
Sounds good. I want to first work on #467 and then move to this.

jmwright:
We need to make sure to include the string selector additions from #514 when this gets done.

adam-urbanczyk:
Indeed, I was planing to use them in the docs too.

adam-urbanczyk:
BTW: after solving this we could release 2.1, what do you think @jmwright ?

jmwright:
> BTW: after solving this we could release 2.1

I think that would be great.

PR #514, opened 2020-11-20
Implement tag based constraint definition

### LINKED ISSUE #467: Support tags for defining assembly constraints

adam-urbanczyk:
Will resolve #467 .

### DIFF
--- a/cadquery/assembly.py
+++ b/cadquery/assembly.py
@@ -260,9 +260,12 @@ class Assembly(object):
         
             name[?tag][@kind@args]
             
-        for example:
+        valid example include:
         
-            obj_name@faces@>Z
+            obj_name @ faces @ >Z
+            obj_name?tag1@faces@>Z          
+            obj_name ? tag
+            obj_name
         
         """
 
--- a/cadquery/assembly.py
+++ b/cadquery/assembly.py
@@ -29,8 +29,8 @@ def _define_grammar():
     Separator = Literal("@").suppress()
     TagSeparator = Literal("?").suppress()
 
-    Name = Word(alphas, alphanums).setResultsName("name")
-    Tag = Word(alphas, alphanums).setResultsName("tag")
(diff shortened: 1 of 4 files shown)


adam-urbanczyk:
I think this PR is ready for merging, could you take a look @jmwright ?

adam-urbanczyk:
@marcus7070 you are probably also interested

cadquery/cq.py
```diff
@@ -390,15 +393,15 @@ def _getTagged(self, name: str) -> "Workplane":
 
         :param name: the tag to search for
         :type name: string
-        :returns: the first CQ object in the parent chain with tag == name
-        :raises: ValueError if no object tagged name in the chain
+        :returns: the a CQ object with tag == name
```
jmwright, line 396: "the a" should probably just be "the" here

jmwright (approved):
@adam-urbanczyk Looks good, thanks!

cadquery/assembly.py
```diff
@@ -231,24 +258,40 @@ def _query(self, q: str) -> Tuple[str, Optional[Shape]]:
...
+        name: str = query.name
 
-        tmp = Workplane()
         obj = self.objects[name].obj
 
-        if isinstance(obj, (Workplane, Shape)):
-            tmp.add(obj)
-            res = getattr(tmp, kind)(arg)
+        if isinstance(obj, Workplane) and query.tag:
+            tmp = obj.ctx.tags[query.tag]
```
marcus7070, line 281: I'm a bit late here, but wouldn't it be better to use `tmp = obj._getTagged(query.tag)`? Try to keep a single method to access tags to be DRY, consistent behaviour (note this will raise a KeyError, while the rest of CQ will raise a ValueError with a bad tag), etc.

Otherwise looks great to me.

adam-urbanczyk: Good point, will fix it together with #469 .

Issue #521, opened 2020-11-26
Mirror about the workplane of a face. 


blooop:
I am making a function that mirrors about a face.  From looking at the code this is not possible, as it only supports string arguments for the basic XY XZ etc workplanes.

https://github.com/CadQuery/cadquery/blob/b22436c59e1c1d44e2bf832df52d6ebcca6cdbaf/cadquery/cq.py#L1024

my attempt:

```
obj = cq.Workplane("XY").rect(1, 2).extrude(1)

def mirrorAboutFace(obj, face):
    planeString = "XY"  # hardcoded, cannot work out how to get this via code
    planeOffset = face.objects[0].Center()
    return obj.mirror(planeString, planeOffset)


obj = mirrorAboutFace(obj, obj.faces("<Z"))  # works as expected
obj = mirrorAboutFace(obj, obj.faces(">Z"))  # works as expected

obj = mirrorAboutFace(obj, obj.faces("<X"))  # does not work because of hardcoded "XY" plane
```

Even if its not possible to mirror about an arbitrary workplane, is it possible to extract the string "XY" parameter from a face object?

Thanks

adam-urbanczyk:
You could modify `Shape.mirror` to accept `cq.Plane` objects.  Note that the current plane is stored in `cq.Workplane.plane`.

https://github.com/CadQuery/cadquery/blob/b22436c59e1c1d44e2bf832df52d6ebcca6cdbaf/cadquery/occ_impl/shapes.py#L467

blooop:
Thanks for the help.  I've implemented the feature as two new functions and left the original mirror alone while i was debugging and testing.  I could make a pull request for the new functionality.  Should I leave them as separate functions, or integrate them into the existing mirror function?

```
    def mirrorAboutPlane(self, mirrorPlaneNormal=(1, 0, 0), basePointVector=(0, 0, 0), union=False):
        """
        Mirror a single CQ object about an arbitrary plane defined by a point and a normal

        :param mirrorPlaneNormal: the normal vector of the plane to mirror about
        :type mirrorPlaneNormal: tuple
        :param basePointVector: the base point to mirror about
        :type basePointVector: tuple
        :param union: If true will perform a union operation on the mirrored object
        :type union: bool
        """
        newS = self.newObject([self.objects[0].mirrorAboutPlaneNormal(mirrorPlaneNormal, basePointVector)])
        if union:
            return self.union(newS.first())
        else:
            return newS.first()

[...]

PR #526, opened 2020-11-29
Add the ability to mirror from a selected face

blooop:
Hi,

Apologies if I have not some something correctly, this is my first pull request so still working out how to do things. 

I added this feature after asking a question on in the cadquery github page:
https://github.com/CadQuery/cadquery/issues/521

The featured adds support for mirroring about a selected face and adds an option to perform a union as part of the mirror operation.  When mirroring about a selected face I think that in most cases a mirror union will be the desired operation rather than leaving two touching but unconnected bodies. 

I have added some tests and updated the examples on how to mirror about a face.  In the tests, I perform checks against the bounding box, but I think checking the volume would be better. I wasn't able to to work out how to call the method that returns the volume. 

Please let me know if I have done anything incorrectly or need to add anything

Thanks,
Austin

### DIFF
--- a/cadquery/cq.py
+++ b/cadquery/cq.py
@@ -1021,17 +1021,38 @@ class Workplane(object):
             ]
         )
 
-    def mirror(self, mirrorPlane="XY", basePointVector=(0, 0, 0)):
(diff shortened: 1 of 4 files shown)


blooop:
I have looked at the conflict and do not understand what is wrong or how I can fix it.

PR #532, opened 2020-12-04
Change to ProjectedOrigin

### LINKED ISSUE #314: Change centerOption to ProjectedOrigin and remove warning

adam-urbanczyk:
Resolves #314

### DIFF
--- a/cadquery/cq.py
+++ b/cadquery/cq.py
@@ -413,14 +413,13 @@ class Workplane(object):
 
         return self.val().wrapped
 
-    @deprecate_kwarg("centerOption", "ProjectedOrigin")
     def workplane(
         self,
         offset: float = 0.0,
         invert: bool = False,
         centerOption: Literal[
             "CenterOfMass", "ProjectedOrigin", "CenterOfBoundBox"
-        ] = "CenterOfMass",
+        ] = "ProjectedOrigin",
         origin: Optional[VectorLike] = None,
     ) -> "Workplane":
         """
--- a/doc/examples.rst
+++ b/doc/examples.rst
@@ -406,7 +406,7 @@ how deep the part is
 .. cadquery::
 
     result = cq.Workplane("front").box(3,2, 0.5)                 #make a basic prism
-    result = result.faces(">Z").vertices("<XY").workplane()  #select the lower left vertex and make a workplane
+    result = result.faces(">Z").vertices("<XY").workplane(centerOption='CenterOfMass')  #select the lower left vertex and make a workplane
     result = result.circle(1.0).cutThruAll()                 #cut the corner out
 
 .. topic:: Api References
--- a/doc/examples.rst
+++ b/doc/examples.rst
@@ -406,7 +406,7 @@ how deep the part is
 .. cadquery::
 
     result = cq.Workplane("front").box(3,2, 0.5)                 #make a basic prism
-    result = result.faces(">Z").vertices("<XY").workplane(centerOption='CenterOfMass')  #select the lower left vertex and make a workplane
+    result = result.faces(">Z").vertices("<XY").workplane(centerOption="CenterOfMass")  #select the lower left vertex and make a workplane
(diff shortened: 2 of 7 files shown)


adam-urbanczyk:
It should pass now. In principle after merging this we could release (see: https://github.com/CadQuery/cadquery/projects/4), but maybe we want to merge other open PRs too.

adam-urbanczyk:
Still need to fix some examples it seems.

jmwright:
@adam-urbanczyk These changes look ok, but I can take another look once the doc tests are passing if you want.

adam-urbanczyk:
Ok ready with all fixes.

cadquery/cq.py
```diff
@@ -413,14 +413,13 @@ def toOCC(self) -> Any:
 
         return self.val().wrapped
 
-    @deprecate_kwarg("centerOption", "ProjectedOrigin")
```
jmwright, line 416: I'm assuming we want to leave it for future use so it's probably not a problem, but it looks like `deprecate_kwarg` won't be used anywhere after this removal.

doc/examples.rst
```diff
@@ -406,7 +406,7 @@ how deep the part is
 .. cadquery::
 
     result = cq.Workplane("front").box(3,2, 0.5)                 #make a basic prism
-    result = result.faces(">Z").vertices("<XY").workplane()  #select the lower left vertex and make a workplane
+    result = result.faces(">Z").vertices("<XY").workplane(centerOption="CenterOfMass")  #select the lower left vertex and make a workplane
```
jmwright, line 409: Is specifying `CenterOfMass` necessary in all the examples and tests? Normally the `ProjectedOrigin` and `CenterOrMass` options yield nearly identical results until material is removed in a non-symmetrical way.

tests/__init__.py
```diff
@@ -25,12 +25,15 @@ def makeUnitSquareWire():
     )
 
 
-def makeUnitCube():
-    return makeCube(1.0)
+def makeUnitCube(centered=True):
```
jmwright, line 28: If a unit cube is supposed to be centered, should it be centered on the Z axis too so that the center of the cube is at the origin (0, 0, 0)? The current method only centers on the X and Y. If that's the way it was before, just disregard this comment.

jmwright (approved):
@adam-urbanczyk I made a few comments. The PR looks good. Thanks for fixing the selector descriptions too.

cadquery/cq.py
adam-urbanczyk: Correct, it'll be neede sooner or later.

tests/__init__.py
adam-urbanczyk: It was not centered at all, I'll adjust it.

doc/examples.rst
adam-urbanczyk: I only changed if an error occurred.

tests/__init__.py
adam-urbanczyk: I changed the param to `xycentered`. Complete centering was breaking more tests.

adam-urbanczyk:
OK, I'm going ahead with the merge.

bernhard-42:
@adam-urbanczyk Could it be that this breaks the [assembly tutorial](https://cadquery.readthedocs.io/en/latest/assy.html#defining-reusable-components)? It seems to work again when one e.g. adds `centerOption` to the second `workplane` call in `make_connector`:

```python
def make_connector():

    rv = (
        cq.Workplane()
        .box(20, 20, 20)
        .faces("<X")
        .workplane()
        .cboreHole(6, 15, 18)
        .faces("<Z")
        .workplane(centerOption="CenterOfMass")
        .cboreHole(6, 15, 18)
    )

    # tag mating faces
    rv.faces(">X").tag("X").end()
    rv.faces(">Z").tag("Z").end()

    return rv
```

adam-urbanczyk:
Yest it does, thanks @bernhard-42 !

Issue #533, opened 2020-12-04
Combine() has no glue=True option


bragostin:
`combine(self, clean: bool = True)` does not have the glue=True option, which was recently implemented in `union()` and is very useful.
Would you accept that I do a pull request to implement it?

jmwright:
+1 from me, but I would wait to see what @adam-urbanczyk says as well.

adam-urbanczyk:
+1

adam-urbanczyk:
Solved by #535

PR #535, opened 2020-12-06
Add glue option to combine()

bragostin:
This is the pull request solving issue #533

### DIFF
--- a/cadquery/cq.py
+++ b/cadquery/cq.py
@@ -2855,12 +2855,19 @@ class Workplane(object):
 
         return self.newObject([r])
 
-    def combine(self, clean: bool = True) -> "Workplane":
+    def combine(
+        self,
+        clean: bool = True,
+        glue: bool = False,
+        tol: Optional[float] = None,
+    ) -> "Workplane":
         """
         Attempts to combine all of the items on the stack into a single item.
         WARNING: all of the items must be of the same type!
 
         :param boolean clean: call :py:meth:`clean` afterwards to have a clean shape
+        :param boolean glue: use a faster gluing mode for non-overlapping shapes (default False)
+        :param float tol: tolerance value for fuzzy bool operation mode (default None)
         :raises: ValueError if there are no items on the stack, or if they cannot be combined
         :return: a CQ object with the resulting object selected
         """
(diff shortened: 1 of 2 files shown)


bragostin:
Codecov complains that "Added line #L2874 was not covered by tests" in cadquery/cq.py#L2874
even though I added in line 2168 of `tests/test_cadquery.py`
```python
        objects1.combine(glue=True, tol=None)
        self.assertEqual(11, objects1.faces().size())
```
to cover the new `glue=True` option in `combine()`.
Any idea what is missing?

jmwright:
@bragostin Unless it's something serious, I take what codecov says with a grain of salt (not too seriously). It often complains about things that are either non-issues or not anything major.

adam-urbanczyk:
Codecov does not make that kind of mistakes AFAIK. Probably you have only one item (cq. Compound) in objects. Just put a breakpoint in this test and verify yourself.

bragostin:
You are right @adam-urbanczyk , the problem was that I had one Compound only in objects, now Codecov passes.
What I don't understand though, is that initially I simply copied the original test, which has also one Compound only when I print `objects1.vals()` before combine is used.

adam-urbanczyk:
You touched this line and that's why it was taken into account in the delta coverage calculation.

adam-urbanczyk:
BTW: is this ready?

bragostin:
@adam-urbanczyk   yes it is

adam-urbanczyk (approved):

cadquery/cq.py
```diff
@@ -2855,20 +2853,25 @@ def _cutFromBase(self, obj: Shape) -> "Workplane":
...
         :raises: ValueError if there are no items on the stack, or if they cannot be combined
         :return: a CQ object with the resulting object selected
         """
+
         items: List[Shape] = [o for o in self.objects if isinstance(o, Shape)]
         s = items.pop(0)
 
         if items:
-            s = s.fuse(*items)
+            s = s.fuse(*items, glue=glue, tol=tol)
```
jmwright, line 2875: Having the tol parameter here won't cause problems when glue is false will it?

jmwright (approved):
Thanks @bragostin !

cadquery/cq.py
adam-urbanczyk: They are orthogonal, so no issue.

adam-urbanczyk:
Merging, thanks for the contribution @bragostin !

PR #555, opened 2020-12-27
Expose makeWire to parametricCurve

just-georgeb:
The use case - creating a complex wire made up of multiple parametricCurves or a combination of parametricCurves and other 2d operations. I think the only way currently would be to create each parametricCurve individually (creating separate spline wires) and then using Wire.combine() at the end, which seems messy.

### DIFF
--- a/cadquery/cq.py
+++ b/cadquery/cq.py
@@ -1746,6 +1746,7 @@ class Workplane(object):
         N: int = 400,
         start: float = 0,
         stop: float = 1,
+        makeWire: bool = True,
     ) -> "Workplane":
         """
         Create a spline interpolated through the provided points.
@@ -1755,6 +1756,7 @@ class Workplane(object):
         :param N: number of points for discretization
         :param start: starting value of the parameter t
         :param stop: final value of the parameter t
+        :param makeWire: convert the resulting spline edge to a wire
         :return: a Workplane object with the current point unchanged
 
         """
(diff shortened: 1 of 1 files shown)


adam-urbanczyk:
You can already achieve this by calling `.wire()`, why this PR?

just-georgeb:
Seems odd to me that the `Spline()` function has makeWire as an option, but `parametricCurve()` (which calls spline under the hood) does not.

For me I am trying to make gear geometry - so I'm using `parametricCurve()` for the tooth flanks, with `radiusArc()` in between. I actually did this a little while ago so it's not as fresh in memory as I'd like - but from memory the issue was that with `makeWire=True`, the current position on the workplane doesn't change, so you then need to use `moveTo()` as well as combining the wires at the end to make a continuous profile.

If not wanted then that's fine though - it was simple to add and helped me out so thought I'd add it as a PR

adam-urbanczyk:
Thanks @just-georgeb , making things consistent with `spline` sounds like a good rationale. If all tests pass we'll take a look and merge it.

BTW: if you manage to make a nice gear with CQ please share the code if possible.

just-georgeb:
Definitely will do! It's working at the moment but the code is a real mess and a bit bandit so needs some work. 

Also sorry have messed this commit up - needs another change before you review so it actually uses the makeWire value!

just-georgeb:
Should be good to go now by the way (not sure how I managed to make such a mess of 3 lines!)

adam-urbanczyk (approved):

jmwright (approved):
Looks good, thanks @just-georgeb

Issue #568, opened 2021-01-07
Syntactic sugar for boolean operations


RubenRubens:
Short and sweet.

```python
import cadquery as cq

Box = cq.Workplane('XY').box(1, 1, 1, centered=(False, False, False))
Sphere = cq.Workplane('XY').sphere(1)

# Union
op1 = Box + Sphere

# Difference
op2 = Box - Sphere

# Intersection
op3 = Box * Sphere
```

Here's the [fork](https://github.com/RubenRubens/cadquery/tree/boolean).

michaelgale:
Oh wow!  Why haven't we thought of that before!  I like it.  I would propose adding support for the "|" (union) and the "&" (intersection) operators so that it is consistent with python's use of these operators for the same operations of the `set` builtin type.

michaelgale:
Question:  do expressions such as the following also work:
```python
op1 = Box.rotate((0,0,0), (0,0,1), 45).translate((-1, 0, 0)) + Sphere.translate((0, 0, 2))
```

adam-urbanczyk:
They have to -> `rotate` and `translate` return a `Workplane`.

michaelgale:
@adam-urbanczyk Of course--I realized just after posting--but I left it as a "dumb" question to remind myself to think before posting!  :)

michaelgale:
This is likely a bit of rabbit hole, but could we also consider extending this syntactic sugar to support operations such as:
```python
# translation
op1 = Box + (0, 0, 2)
op1 = (Box - (0, 0, 2)) + (Box + (0, 0, 2))

# scaling
op1 = 3 * Box
# op1 is a box scaled 3x larger in all axes
op1 = (1, 1, 3) * Box
# op1 is box scaled 3x taller in Z (X,Y remain unscaled)
```
This would require a bit more parsing of argument types to determine if the "`+/-`" operators apply a CSG operation or simply a translation.  

Lastly, would we also consider adding support for "`+=`" and "`-=`" too?  e.g.
```python
Box += Sphere
Box -= Box.translate((0, 0, -3))
```

RubenRubens:
Short and sweet is the priority for me. I strongly disagree with making operations overcomplicated. From my point of view we have two options:

## (A) Support `+`, `-` and `*` symbols

[...]

PR #569, opened 2021-01-07
Boolean operation syntactic sugar

RubenRubens:


marcus7070:
It might be worth mentioning in the docs that with `r = a + b`, `r.parent == a` and r has the same CQContext object as a.

Or perhaps just state `r = a + b` is equivalent to `r = a.union(b)`.

jmwright:
@RubenRubens It looks like the lint check is failing.

RubenRubens:
I think it's OK now. Thanks for the helpful comments.

RubenRubens:
There is disscussion in #568 about which operators we should include. Suming up, there are multiple options but... is this the way to go?

+ Union: `|`
+ Cut: `-`
+ Intersection `&`

I'm happy with it and hopefully @adam-urbanczyk would be too. 😄

adam-urbanczyk:
I think I'd still add `+` as an alternative to `|`. What do you think @jmwright ?

jmwright:
> I think I'd still add + as an alternative to |.

+1

marcus7070:
Sphinx doesn't autosummary dunder methods (except `__init__`), so these new methods don't appear in the HTML docs at all. Relevant stackoverflow: https://stackoverflow.com/questions/62142793/autosummary-generated-documentation-missing-all-dunder-methods-except-for-ini 

Looks easier to just write a section manually than to make sphinx autosummary do what we want.

Linking to #493.

jmwright:
There's also this that shows up just after the lint check.
```
$ mypy cadquery
cadquery/cq.py:3772: error: Unsupported left operand type for + ("Type[Workplane]")
Found 1 error in 1 file (checked 24 source files)
The command "mypy cadquery" exited with 1.
```

adam-urbanczyk:
Yes, MyPy does not like monkey patching. Will fix it.

adam-urbanczyk (approved):

adam-urbanczyk:
@jmwright looks ready to me. Codecov complains about a docstring line...

jmwright:
Ok, merging.

jmwright (approved):

PR #574, opened 2021-01-12
Fix split docs

RubenRubens:


jmwright:
Thanks again @RubenRubens

PR #584, opened 2021-01-16
remove unused variable

RubenRubens:
`width` was never used.


jmwright:
@RubenRubens The MacOS builds failed because of something that seems internal to Azure, so I'm re-running them. Hopefully that fixes it. This change looks good otherwise, thanks.

PR #592, opened 2021-01-18
fix unconstrained assembly.solve()

### LINKED ISSUE #591: crash calling .solve() on an assembly with no constraints
```
line 139, in main
    asy.solve()
  File "/usr/lib/python3.9/site-packages/cadquery/assembly.py", line 408, in solve
    locs_new = solver.solve()
  File "/usr/lib/python3.9/site-packages/cadquery/occ_impl/solver.py", line 212, in solve
    res = minimize(
  File "/usr/lib/python3.9/site-packages/scipy/optimize/_minimize.py", line 614, in minimize
    return _minimize_bfgs(fun, x0, args, jac, callback, **options)
  File "/usr/lib/python3.9/site-packages/scipy/optimize/optimize.py", line 1164, in _minimize_bfgs
    gnorm = vecnorm(gfk, ord=norm)
  File "/usr/lib/python3.9/site-packages/scipy/optimize/optimize.py", line 166, in vecnorm
    return np.amax(np.abs(x))
  File "<__array_function__ internals>", line 5, in amax
  File "/usr/lib/python3.9/site-packages/numpy/core/fromnumeric.py", line 2705, in amax
    return _wrapreduction(a, np.maximum, 'max', axis, None, out,
  File "/usr/lib/python3.9/site-packages/numpy/core/fromnumeric.py", line 87, in _wrapreduction
    return ufunc.reduce(obj, axis, dtype, out, **passkwargs)
ValueError: zero-size array to reduction operation maximum which has no identity
```

greyltc:
Fixes #591

### DIFF
--- a/cadquery/assembly.py
+++ b/cadquery/assembly.py
@@ -400,16 +400,17 @@ class Assembly(object):
         constraints = []
         for c in self.constraints:
             constraints.append(((ents[c.objects[0]], ents[c.objects[1]]), c.toPOD()))
+        
+        if not constraints == []:
+            # instantiate the solver
+            solver = ConstraintSolver(locs, constraints, locked=[lock_ix])
 
-        # instantiate the solver
-        solver = ConstraintSolver(locs, constraints, locked=[lock_ix])
+            # solve
+            locs_new = solver.solve()
 
-        # solve
-        locs_new = solver.solve()
-
-        # update positions
-        for loc_new, n in zip(locs_new, ents):
-            self.objects[n].loc = loc_new
+            # update positions
+            for loc_new, n in zip(locs_new, ents):
+                self.objects[n].loc = loc_new
 
         return self
 
--- a/cadquery/assembly.py
+++ b/cadquery/assembly.py
@@ -400,17 +400,20 @@ class Assembly(object):
         constraints = []
         for c in self.constraints:
             constraints.append(((ents[c.objects[0]], ents[c.objects[1]]), c.toPOD()))
-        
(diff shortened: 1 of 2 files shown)


cadquery/assembly.py
```diff
@@ -400,16 +400,17 @@ def solve(self) -> "Assembly":
         constraints = []
         for c in self.constraints:
             constraints.append(((ents[c.objects[0]], ents[c.objects[1]]), c.toPOD()))
+        
+        if not constraints == []:
```
marcus7070, line 404: Can't you just do `if not constraints:`?

greyltc: yeah, you can[1]. I just find this way to be more clear on what's being checked.
[1]: https://docs.python.org/3/library/stdtypes.html#truth-value-testing

adam-urbanczyk (approved):

adam-urbanczyk:
I reworked to raise a ValueError with a clear message. Anything against @greyltc ?

greyltc:
Yeah, it's very likely the user has made a mistake if they call solve() without setting up constraints.

I can imagine a case where it might be convenient for solve() to be a noop (or just throw a warning), like if you're troubleshooting constraints and you've commented them all out or something.

I'm fine with leaving it as an error though.

adam-urbanczyk:
I'm on the being explicit side on that one - solve only makes sense with constraints.

adam-urbanczyk:
Are you fine with merging @jmwright ?

jmwright:
+1

PR #596, opened 2021-01-20
Expose SVG Export Options and Add Some Useful Extras

jmwright:
I'm still working on this, but wanted to get it out in the open to get feedback.

Exposed the `getSVG()` 'opt' parameter so that things like height, width and margins could be controlled. Also added a projection direction opt to control the projection view of the shape. The following code currently works with this fork/branch:
```python
import cadquery as cq

result = cq.Workplane("XY").circle(1.0)

exporters.export(result,
                 "/home/jwright/Downloads/out.svg",
                 opt={"width": 100,
                      "height": 100,
                      "marginLeft": 10,
                      "marginTop": 10,
                      "projectionDir": (0, 0, 1)})
```
There's been some discussion around SVGs and HLR lately, so feedback is welcome. I'll share on Discord to get feedback as well.

Tagging @adam-urbanczyk and @marcus7070

### DIFF
--- a/cadquery/occ_impl/exporters/svg.py
+++ b/cadquery/occ_impl/exporters/svg.py
@@ -263,16 +263,8 @@ def getSVG(shape, opts=None):
         {
             "unitScale": str(unitScale),
             "strokeWidth": str(strokeWidth),
-            "strokeColor": str(strokeColor[0])
-            + ","
-            + str(strokeColor[1])
-            + ","
-            + str(strokeColor[2]),
-            "hiddenColor": str(hiddenColor[0])
-            + ","
-            + str(hiddenColor[1])
-            + ","
-            + str(hiddenColor[2]),
+            "strokeColor": ",".join([str(x) for x in strokeColor]),
+            "hiddenColor": ",".join([str(x) for x in hiddenColor]),
             "hiddenContent": hiddenContent,
             "visibleContent": visibleContent,
             "xTranslate": str(xTranslate),
--- a/cadquery/occ_impl/exporters/svg.py
+++ b/cadquery/occ_impl/exporters/svg.py
@@ -61,7 +61,7 @@ class UNITS:
 
 def guessUnitOfMeasure(shape):
     """
-        Guess the unit of measure of a shape.
+    Guess the unit of measure of a shape.
     """
     bb = BoundBox._fromTopoDS(shape.wrapped)
 
@@ -83,7 +83,7 @@ def guessUnitOfMeasure(shape):
 
 def makeSVGedge(e):
     """
-
+    Creates an SVG edge from a OCCT edge.
     """
 
     cs = StringIO.StringIO()
@@ -108,7 +108,7 @@ def makeSVGedge(e):
 
 def getPaths(visibleShapes, hiddenShapes):
     """
-
+    Collects the visible and hidden edges from the CadQuery object.
     """
 
     hiddenPaths = []
@@ -127,21 +127,22 @@ def getPaths(visibleShapes, hiddenShapes):
 
 def getSVG(shape, opts=None):
     """
-        Export a shape to SVG
+    Export a shape to SVG
     """
 
     # Available options and their defaults
-    d = {"width": 800,
-         "height": 240,
-         "marginLeft": 200,
-         "marginTop": 20,
-         "projectionDir": (-1.75, 1.1, 5),
-         "showAxes": True,
-         "strokeWidth": -1.0, # -1 = calculated based on unitScale
-         "strokeColor": (0, 0, 0), # RGB 0-255
-         "hiddenColor": (160, 160, 160), #RGB 0-255
-         "showHidden": True
-        }
+    d = {
+        "width": 800,
+        "height": 240,
+        "marginLeft": 200,
+        "marginTop": 20,
+        "projectionDir": (-1.75, 1.1, 5),
+        "showAxes": True,
+        "strokeWidth": -1.0,  # -1 = calculated based on unitScale
+        "strokeColor": (0, 0, 0),  # RGB 0-255
+        "hiddenColor": (160, 160, 160),  # RGB 0-255
+        "showHidden": True,
+    }
 
     if opts:
         d.update(opts)
@@ -237,11 +238,7 @@ def getSVG(shape, opts=None):
     # If the caller wants the axes indicator and is using the default direction, add in the indicator
     if showAxes and projectionDir == (-1.75, 1.1, 5):
         axesIndicator = AXES_TEMPLATE % (
-            {
-                "unitScale": str(unitScale),
-                "textboxY": str(height - 30),
-                "uom": str(uom)
-            }
+            {"unitScale": str(unitScale), "textboxY": str(height - 30), "uom": str(uom)}
         )
     else:
         axesIndicator = ""
@@ -250,8 +247,16 @@ def getSVG(shape, opts=None):
         {
             "unitScale": str(unitScale),
             "strokeWidth": str(strokeWidth),
-            "strokeColor": str(strokeColor[0]) + "," + str(strokeColor[1]) + "," + str(strokeColor[2]),
-            "hiddenColor": str(hiddenColor[0]) + "," + str(hiddenColor[1]) + "," +  str(hiddenColor[2]),
+            "strokeColor": str(strokeColor[0])
+            + ","
+            + str(strokeColor[1])
+            + ","
+            + str(strokeColor[2]),
+            "hiddenColor": str(hiddenColor[0])
+            + ","
+            + str(hiddenColor[1])
+            + ","
+            + str(hiddenColor[2]),
             "hiddenContent": hiddenContent,
             "visibleContent": visibleContent,
             "xTranslate": str(xTranslate),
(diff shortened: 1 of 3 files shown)


marcus7070:
It looks like you're planning to add an option to display the axis indicators?

cadquery/occ_impl/exporters/svg.py
```diff
@@ -128,7 +129,7 @@ def getSVG(shape, opts=None):
         Export a shape to SVG
     """
```
marcus7070, line 130: I personally need docs on these opts here, particularly a reference for `projectionDir`. Something like
> a `projectionDir` of (-1, 1, 1) displays the model in an isometric view, with the z-axis pointing upwards on the page, x to the right and y to the left

I'll forget every time if this direction points from the camera to the model or vice versa and have to check the doc string. :sweat_smile:

cadquery/occ_impl/exporters/svg.py
```diff
@@ -128,7 +129,7 @@ def getSVG(shape, opts=None):
...
 
-    d = {"width": 800, "height": 240, "marginLeft": 200, "marginTop": 20}
```
marcus7070, line 131: This is just a minor style thing, but if we intend for a user to modify these options, I'd suggest just making them keyword args to the function. CQ already uses much more arguments in https://github.com/CadQuery/cadquery/blob/53a76d4f288c4babd416dfb3b2c3e7b9153f754f/cadquery/occ_impl/shapes.py#L1761-L1805, I don't see any reason to handle them a different way here.

jmwright:
@marcus7070 

> It looks like you're planning to add an option to display the axis indicators?

I'm actually thinking of taking the axis indicator out because it's too hard to draw it dynamically for each possible projection direction. I'm not sure how much value it adds either.

cadquery/occ_impl/exporters/svg.py
jmwright: I'm actually hoping to write a new import/export section of the docs that expands on the current references. I can update the doc string of getSVG too.

cadquery/occ_impl/exporters/svg.py
jmwright: I was trying to be consistent with what has been done previously with the `opt` parameter. However, I can structure it according to the consensus. I don't have any real attachment to the current `opt` method.

adam-urbanczyk: I think I also prefer flat options. Maybe a deprecation warning and temporary support for both would be a good consensus?

adam-urbanczyk:
> @marcus7070
> 
> > It looks like you're planning to add an option to display the axis indicators?
> 
> I'm actually thinking of taking the axis indicator out because it's too hard to draw it dynamically for each possible projection direction. I'm not sure how much value it adds either.

We could use CQ to render the indicator, say a cube with embossed +-XYZ text on appropriate faces. But maybe that is something for a separate PR and definitely should be optional.

Another thing I realized recently - maybe it'd be good to make hidden dashed line optional. For complex models they add quite some visual noise:

image

MarcSallent:
Looks good!
I've been not able to test it yet, but I consider the following options interesting for many folks:
- Hidden lines:
  - visible / hidden (not transparent, but entirely not drawn)
  - stroke-width
  - stroke (color)
- Visible lines:
  - stroke-width
  - stroke (color)
  
These are actually the ones that I found myself editing to preview the SVGs in the browser.
I just found out that you can add both stroke-width and color to the [style attribute of the SVG tag](https://www.w3.org/TR/SVG/styling.html#:~:text=As%20with%20HTML%2C%20SVG%20supports,to%20support%20element%2Dspecific%20styling.&text=The%20'class'%20attribute%20assigns%20one,addressing%20by%20the%20styling%20language.&text=The%20'style'%20attribute%20is%20used,CSS%20declaration%20of%20an%20element.), so an option could be to send the CSS attributes instead of many options in the function call? But I don't how compatible is that with existing SVG readers

jmwright:
Thanks for the feedback everyone.

> We could use CQ to render the indicator, say a cube with embossed +-XYZ text on appropriate faces. But maybe that is something for a separate PR and definitely should be optional.

I'll add a deprecation warning. I definitely think we can do something better with that indicator in the future. I'm wondering if for now I just go ahead and add an option to show/hide the axes indicator and make it so that it only works with the default direction, and then note that in the docstring. It will be a little confusing for now, but everything will fall into place when a new indicator is added later.

> I think I also prefer flat options.

@adam-urbanczyk By "flat options" you mean keyword arguments, correct?

@adam-urbanczyk @marcus7070 Am I understanding correctly that the dictionary `opt` will still be passed to `exporters.export()`, but that the dict will be destructured to be passed to the `exportSVG`/`getSVG` methods, which will accept keyword arguments instead of a dict?

adam-urbanczyk:
> > I think I also prefer flat options.
> 
> @adam-urbanczyk By "flat options" you mean keyword arguments, correct?
> 
> @adam-urbanczyk @marcus7070 Am I understanding correctly that the dictionary `opt` will still be passed to `exporters.export()`, but that the dict will be destructured to be passed to the `exportSVG`/`getSVG` methods, which will accept keyword arguments instead of a dict?

Indeed, that is what I meant. Though after looking at the code I'm not so sure if I still have that opinion. Maybe let's park it for now.

jmwright:
There's still work to do, but I think I've got all the options added unless there's a desire to have hidden line stroke-width set independently of the visible line stroke-width.

The `opts` dictionary has to stay in some capacity because of how the exporters interface works, but I may take a shot at getting us set up to benefit from type annotations via keyword arguments on `getSVG()`. With the way the interface is the `opts` dict makes sense, but I'd like to be able to have a standard docstring with explanations of each option and type annotations too.

The axes indicator only works if you set `showAxes` to `True` and don't change the `projectionDir` (let it stay at  the default). If you don't pass any opts in, the exported SVG will look exactly like it did before I added these changes. The following code shows what options are available, and how it looks. 

```python
import cadquery as cq

result = cq.Workplane("XY").box(10, 10, 10)#.circle(1.0)

exporters.export(result,
                 "/home/jwright/Downloads/out.svg",
                 opt={"width": 100,
                      "height": 100,
                      "marginLeft": 10,
                      "marginTop": 10,
                      "showAxes": False,
                      "projectionDir": (-1.75, 1.1, 5),
                      "strokeWidth": 0.25,
                      "strokeColor": (255, 0, 0),
                      "hiddenColor": (0, 0, 255),
                      "showHidden": True})

exporters.export(result,
                 "/home/jwright/Downloads/out2.svg")
```

out.svg:
Screenshot from 2021-01-21 17-50-55


out2.svg:
Screenshot from 2021-01-21 17-49-31

cadquery/occ_impl/exporters/svg.py
jmwright: Added a docstring and #598 to keep track of this.

cadquery/occ_impl/exporters/svg.py
jmwright: Adding keyword args ended up being a dead-end for now, as @adam-urbanczyk expected. I would have to rework too much of the `export.exporters` interface and I wasn't willing to do that at this time. I tried to put as much detail in the `getSVG()` docstring as possible to offset this at least a little bit.

cadquery/occ_impl/exporters/svg.py
```diff
@@ -250,8 +247,16 @@ def getSVG(shape, opts=None):
...
+            "strokeColor": str(strokeColor[0])
+            + ","
+            + str(strokeColor[1])
+            + ","
+            + str(strokeColor[2]),
+            "hiddenColor": str(hiddenColor[0])
+            + ","
+            + str(hiddenColor[1])
+            + ","
+            + str(hiddenColor[2]),
```
marcus7070, line 275: Yikes, black has not helped here. Maybe:

```suggestion
            "strokeColor": ",".join([str(x) for x in strokeColor]),
            "hiddenColor": ",".join([str(x) for x in hiddenColor]),
```

jmwright: Thanks @marcus7070 I committed that change. Much, much better. Let me know if you see anything else.

jmwright:
@adam-urbanczyk codecov refuses to see that line 260 is covered, even though I can test it manually and it is hit. Please review at your convenience. If you don't find anything else, this should be ready to merge unless you want to wait until after the 2.1 release.

jmwright:
I really wasn't paying enough attention. I copied the STL test and added the SVG options to it. I've fixed that now and hopefully everything will go green, including the lint error for line 260.

adam-urbanczyk:
Great, let's merge this before the final 2.1 . Just give me some time for the review.

adam-urbanczyk (approved):

adam-urbanczyk:
Looks good @jmwright , thanks! Shall I merge it?

jmwright:
> Looks good @jmwright , thanks! Shall I merge it?

Yes, please do.

jmwright:
I'm adding this here for future reference, and to let the community know about it.

The Paramak team is already making good use of these new SVG export options [here](https://github.com/ukaea/paramak/pull/716).

They have created an animation script that looks really useful [here](https://github.com/ukaea/paramak/pull/716/files#diff-dfdc5576dc306845e1ee22e2ea92dc23f2eab337d9dfef89c13e502fc57f484b) as well.

PR #597, opened 2021-01-21
Fix ellipse documentation formatting

adam-trhon:
An empty line is missing after ellipse short summary, causing the whole
text to be inserted to API Reference
(http://cadquery.readthedocs.io/en/latest/apireference.html)


jmwright:
Thanks @adam-trhon !

PR #607, opened 2021-01-30
fix .rect calls in tests

marcus7070:
Just noticed these two lines looked wrong.


jmwright (approved):
@marcus7070 Thanks for fixing this. @adam-urbanczyk Are you ok with merging this?

adam-urbanczyk:
+1

Issue #567, opened 2021-01-07
Same options for centered in methods .rect and .box


marcus7070:
I think we can change the way we specify `centered` in both `.rect` and `.box` to be more consistent and more useful. Basically: 
* Keep existing behaviour for `.rect(1, 1, centered=True)` and `.box(1, 1, 1, centered=(True, True, True)`
* `.rect(10, 10, centered=(True, False)` produces a rectangle with corners on (-5, 0) and (5, 10)
* `.box(1, 1, 1, centered=True` is the same as `.box(1, 1, 1, centered=(True, True, True)`

Thoughts?

marcus7070:
These methods also use a 2 or 3d centered option, and could be given the same treatment:

* `rarray`
* `sphere`
* `wedge`

RubenRubens:
I think it's a good idea. Making stuff more consistent is always welcome.

michaelgale:
I like this suggestion.  I can't count how many times I adjust objects by `(-xyz/2)`!  As long as the default behaviour isn't broken, this would be useful.

adam-urbanczyk:
Sounds good, do you mind waiting till after 2.1?

marcus7070:
> Sounds good, do you mind waiting till after 2.1?

No worries, that is fine.

PR #611, opened 2021-01-31
Fix wedge centering

### LINKED ISSUE #610: wedge centering
I think the centering in `Workplane.wedge` is incorrect. I'd like to fix it but just want to make sure I'm not breaking anything, especially given I've never used `wedge` before.

```python
w = cq.Workplane().wedge(10, 10, 10, 4, 4, 6, 6, centered=(True, True, True))
```

screenshot2021-01-31-203112

Note that the bottom left corner is on the origin, which is what I would expect from `centered=(False, False, False)`.

Change centered to `(False, True, True)` and you get:

screenshot2021-01-31-203359

Now the shape doesn't touch the origin at all. The Y and Z directions also behave the same.

Shall I change it such that `centered=(True, True, True)` puts the center of a bounding box around the wedge on the origin?

FYI the last time the `wedge` method was significantly changed was in #228.

adam-urbanczyk:
Will resolve #610

### DIFF
--- a/cadquery/cq.py
+++ b/cadquery/cq.py
@@ -3529,14 +3529,14 @@ class Workplane(object):
 
         (xp, yp, zp) = (0.0, 0.0, 0.0)
 
-        if not centered[0]:
-            xp += dx / 2.0
+        if centered[0]:
+            xp -= dx / 2.0
 
-        if not centered[1]:
-            yp += dy / 2.0
+        if centered[1]:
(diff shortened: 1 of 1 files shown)


marcus7070:
LGTM, I'll add some tests for this when I do #567 after release 2.1.

jmwright (approved):
Thanks @adam-urbanczyk

PR #613, opened 2021-02-01
Wedge center

marcus7070:
I should have read #611 more closely, I missed this as well. I've added a test for good measure.


adam-urbanczyk (approved):

jmwright (approved):
Thanks @marcus7070

adam-urbanczyk:
Thanks for the quick fix and the tests @marcus7070 !

PR #617, opened 2021-02-03
Centered option

### LINKED ISSUE #567: Same options for centered in methods .rect and .box

marcus7070:
Will resolve #567. 

Also fixes some errors in the HTML docs for box and wedge.

### DIFF
--- a/cadquery/cq.py
+++ b/cadquery/cq.py
@@ -1331,8 +1331,10 @@ class Workplane(object):
         :param ySpacing: spacing between points in the y direction ( must be > 0)
         :param xCount: number of points ( > 0 )
         :param yCount: number of points ( > 0 )
-        :param center: if true, the array will be centered at the center of the workplane. if
-            false, the lower left corner will be at the center of the work plane
+        :param center: If True, the array will be centered around the workplane center.
+          If False, the lower corner will be on the reference point and the array will
+          extend in the positive x and y directions. Can also use a 2-tuple to specify
+          centering along each axis.
         """
 
         if xSpacing <= 0 or ySpacing <= 0 or xCount < 1 or yCount < 1:
@@ -2226,8 +2228,10 @@ class Workplane(object):
         :type xLen: float > 0
         :param yLen: length in yDirection ( in workplane coordinates )
         :type yLen: float > 0
-        :param centered: should the center or the lower bound be on the reference point? A single
-          bool can also be given, eg. True is equivalent to (True, True).
+        :param centered: If True, the rectangle will be centered around the reference
+          point. If False, the corner of the rectangle will be on the reference point and
+          it will extend in the positive x and y directions. Can also use a 2-tuple to
+          specify centering along each axis.
         :param forConstruction: should the new wires be reference geometry only?
         :type forConstruction: true if the wires are for reference, false if they are creating part
             geometry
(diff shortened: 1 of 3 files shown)


cadquery/cq.py
```diff
@@ -2224,8 +2226,8 @@ def rect(
         :type xLen: float > 0
         :param yLen: length in yDirection ( in workplane coordinates )
         :type yLen: float > 0
-        :param boolean centered: true if the rect is centered on the reference point, false if the
-            lower-left is on the reference point
+        :param centered: should the center or the lower bound be on the reference point? A single
```
RubenRubens, line 2229: What?

RubenRubens:
Fantastic job!

cadquery/cq.py
marcus7070: Yeah, that's confusing and I should rewrite it. I should probably use the same text for all these methods as well.

marcus7070: This text is going to be used a lot, so I want to make sure it is as concise and clear as possible. Can anyone improve on the following?

```suggestion
        :param centered: If True, the rectangle will be centered around the reference
          point. If False, the corner of the rectangle will be on the reference point and
          it will extend in the positive x and y directions. Can also use a 2-tuple to
          specify centering along each axis.
```

jmwright: I think it sounds fine. "reference point" makes sense (I think) because you can specify a list of points, and not use the workplane center.

cadquery/cq.py
```diff
@@ -2239,22 +2241,32 @@ def rect(
         Creates 4 circles at the corners of a square centered on the origin.
 
         Future Enhancements:
-            better way to handle forConstruction
-            project points not in the workplane plane onto the workplane plane
+            * better way to handle forConstruction
```
jmwright, line 2244: I think this is a hold-over from long ago, and I'm not even sure what it means. Does it make sense to keep it?

cadquery/cq.py
```diff
@@ -3432,43 +3455,45 @@ def sphere(
...
 
         One sphere is created for each item on the current stack. If no items are on the stack, one
         box using the current workplane center is created.
 
-        If combine is true, the result will be a single object on the stack:
-            If a solid was found in the chain, the result is that solid with all spheres produced
-            fused onto it otherwise, the result is the combination of all the produced boxes
+        If combine is true, the result will be a single object on the stack. If a solid was found
+        in the chain, the result is that solid with all spheres produced fused onto it otherwise,
+        the result is the combination of all the produced boxes.
```
jmwright, line 3476: I think the "boxes" reference here may be a very old copy-paste error.

cadquery/cq.py
```diff
@@ -3499,24 +3524,26 @@ def wedge(
...
+        :param clean: True to attempt to have the kernel clean up the geometry, False otherwise
         :return: A wedge object for each point on the stack
 
         One wedge is created for each item on the current stack. If no items are on the stack, one
         wedge using the current workplane center is created.
 
-        If combine is true, the result will be a single object on the stack:
-            If a solid was found in the chain, the result is that solid with all wedges produced
-            fused onto it otherwise, the result is the combination of all the produced wedges
+        If combine is true, the result will be a single object on the stack. If a solid was found
```
jmwright, line 3542: Might want to make true capitalized here to follow the case changes from directly above.

jmwright:
@marcus7070 I re-ran one of the Azure CI checks. It had failed because it said it couldn't reach github.com. All the checks pass now. Let us know when it's ready for review. Thanks!

cadquery/cq.py
marcus7070: Yeah, I think it should be removed. AFAIK there are no issues open for `forConstruction` and no plans to change it's action.

cadquery/cq.py
marcus7070: Thanks, I must have read that 10+ times and it didn't click.

cadquery/cq.py
marcus7070: Done, thanks.

marcus7070:
Ready for review now @jmwright.

jmwright (approved):
Looks good. Thanks @marcus7070 

I've requested a review from @adam-urbanczyk as well.

adam-urbanczyk:
Thanks, looks good @marcus7070 !

PR #630, opened 2021-02-11
force dxf layer name case agreement

### LINKED ISSUE #624: dxf layer import exclusion not matching layer names properly
I'm having a problem with https://github.com/CadQuery/cadquery/pull/442 in that ezdxf is returning layer names that are all upper case, so none of my layer name exclusions are matching.

Not really sure yet why this happening now, maybe exdxf changed or the DXF file format I'm using is somehow causing the layer names to become upper case.

Need to investigate a bit more. Maybe we should just force all the layer names we deal with to lower case to avoid case issues.

greyltc:
fixes #624


jmwright:
@greyltc Please have a look at the black lint check error when you get a chance.

adam-urbanczyk (approved):

jmwright:
Merging. Thanks @greyltc !

PR #636, opened 2021-02-14
Add Support for Specifying Arbitrary Tangents, etc. in spline()

pavpen:
* Added support for specifying the tangents at arbitrary interpolation points when interpolating a B-spline curve with `Workplane.spline()`.
* Added support for specifying whether the tangents should be automatically scaled. (I.e., only use the tangent vector directions, rather than their magnitudes.)
* Added support for specifying the value of the curve function parameter at the interpolation points.
* Added support for specifying the interpolator's tolerance value.

Q: _There are a number of whitespace, and other formatting changes introduced by `black`. Is there a specific list of parameters that you use when running code formatting?_

### DIFF
--- a/cadquery/cq.py
+++ b/cadquery/cq.py
@@ -1769,7 +1769,7 @@ class Workplane(object):
             periodic=periodic,
             parameters=parameters,
             scale=scale,
-            **({'tol' : tol} if tol else {})
+            **({"tol": tol} if tol else {}),
         )
 
         if makeWire:
--- a/cadquery/cq.py
+++ b/cadquery/cq.py
@@ -1672,7 +1672,7 @@ class Workplane(object):
         listOfXYTuple: Iterable[VectorLike],
         tangents: Optional[Sequence[VectorLike]] = None,
         periodic: bool = False,
-        parameters: Optional[List[float]] = None,
+        parameters: Optional[Sequence[float]] = None,
         scale: bool = True,
         tolerance: Optional[float] = None,
         forConstruction: bool = False,
--- a/cadquery/cq.py
+++ b/cadquery/cq.py
@@ -1674,7 +1674,7 @@ class Workplane(object):
         periodic: bool = False,
         parameters: Optional[Sequence[float]] = None,
         scale: bool = True,
-        tolerance: Optional[float] = None,
+        tol: Optional[float] = None,
         forConstruction: bool = False,
         includeCurrent: bool = False,
         makeWire: bool = False,
@@ -1712,7 +1712,7 @@ class Workplane(object):
             
             I.e., set this to True, if you want to use only the direction of
             the tangent vectors specified by ``tangents``, but not their magnitude.
-        :param tolerance: tolerance of the algorithm (consult OCC documentation)
+        :param tol: tolerance of the algorithm (consult OCC documentation)
             
             Used to check that the specified points are not too close to each
             other, and that tangent vectors are not too short. (In either case
@@ -1743,9 +1743,6 @@ class Workplane(object):
 
         *WARNING*  It is fairly easy to create a list of points
         that cannot be correctly interpreted as a spline.
-
-        Future Enhancements:
-          * provide access to control points
         """
 
         vecs = [self.plane.toWorldCoords(p) for p in listOfXYTuple]
@@ -1766,23 +1763,14 @@ class Workplane(object):
         else:
             tangents_g = None
 
-        if tolerance is not None:
-            e = Edge.makeSpline(
-                allPoints,
-                tangents=tangents_g,
-                periodic=periodic,
-                parameters=parameters,
-                scale=scale,
-                tol=tolerance,
-            )
-        else:
-            e = Edge.makeSpline(
-                allPoints,
-                tangents=tangents_g,
-                periodic=periodic,
-                parameters=parameters,
-                scale=scale,
-            )
+        e = Edge.makeSpline(
+            allPoints,
+            tangents=tangents_g,
+            periodic=periodic,
+            parameters=parameters,
+            scale=scale,
+            **({'tol' : tol} if tol else {})
+        )
 
         if makeWire:
             rv_w = Wire.assembleEdges([e])
--- a/cadquery/cq.py
+++ b/cadquery/cq.py
@@ -245,7 +245,7 @@ class Workplane(object):
 
             # drill a hole in the side
             c = Workplane().box(1,1,1).faces(">Z").workplane().circle(0.25).cutThruAll()
-            
+
             # now cut it in half sideways
             c = c.faces(">Y").workplane(-0.5).split(keepTop=True)
         """
@@ -327,7 +327,7 @@ class Workplane(object):
 
     def size(self) -> int:
         """
-         Return the number of objects currently on the stack
+        Return the number of objects currently on the stack
         """
         return len(self.objects)
 
@@ -444,7 +444,7 @@ class Workplane(object):
            * The centerOption paramter sets how the center is defined.
              Options are 'CenterOfMass', 'CenterOfBoundBox', or 'ProjectedOrigin'.
              'CenterOfMass' and 'CenterOfBoundBox' are in relation to the selected
-             face(s) or vertex (vertices). 'ProjectedOrigin' uses by default the current origin 
+             face(s) or vertex (vertices). 'ProjectedOrigin' uses by default the current origin
              or the optional origin parameter (if specified) and projects it onto the plane
              defined by the selected face(s).
            * The Z direction will be normal to the plane of the face,computed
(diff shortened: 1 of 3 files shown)


adam-urbanczyk:
What is exactly the use case for `parameters`? Does not scaling change the geometry or result in an error in the general case?

Regarding black, we use the defaults of `19.10b0`.

cadquery/cq.py
```diff
@@ -1681,8 +1684,33 @@ def spline(
...
+            interpolating.
+                Each tangent is scaled, so it's length is equal to the derivative of
+            the Lagrange interpolated curve.
+                I.e., set this to True, if you want to use only the direction of
+            the tangent vectors specified by ``tangents``, but not their magnitude.
+        :param tolerance: tolerance of the algorithm (consult OCC documentation)
+                Used to check that the specified points are not too close to each
+            other, and that tangent vectors are not too short. (In either case
+            interpolation may fail.)
+                Set to None to use the default tolerance.
```
marcus7070, line 1713: Could you please fix the leading whitespace? If you check the details of the Read The Docs test, you'll see a build of the HTML docs and some bad formatting resulting from this whitespace. https://cadquery--636.org.readthedocs.build/en/636/classreference.html#cadquery.Workplane.spline

marcus7070:
Github was being annoying when I tried to comment on the docstring for `Edge.makeSpline`, since the comment touched on lines @pavpen hadn't written, so I added a commit instead.

Aside from my comments this is all looking pretty good to me. Nice first contribution @pavpen!

adam-urbanczyk:
@pavpen any thoughts regarding my questions above?

pavpen:
> What is exactly the use case for `parameters`? Does not scaling change the geometry or result in an error in the general case?

I mostly set the value of the curve's parameter at different interpolation points, so that I can easily select a point between two interpolation points I'm interested in.

I don't quite understand your second question. Scaling the parameter has no effect on the shape of the curve (provided that the tangents are scaled correspondingly, if you're specifying tangents). It's not an error to choose any parameter values at the various interpolation points, as long as the parameter is monotonically increasing (theoretically, it could work if it's decreasing, too, but OCC seems to throw an error in this case) as you go from interpolation points at the beginning of the curve towards points at the end of the curve.

Here's an example of approximating an Archimedean spiral with a B-spline. (You can see which curve looks best.)

```Python
# -*- coding: utf-8 -*-

from cadquery.cq import Workplane
import cadquery as cq

import math

# Helper functions for defining an Archimedean spiral [Δr(θ) ∝ Δθ]:

# Spiral parameters:
center = cq.Vector(0, 0)
r0 = 1
angle0 = 0
dr_dthetaRadians = 10 / (2 * math.pi)


def point_at_angle(angular_distance):
    r = r0 + dr_dthetaRadians * angular_distance
    angle = angle0 + angular_distance

    return center + cq.Vector(r * math.cos(angle), r * math.sin(angle))


def tangent_at_angle(angular_distance):
    r = r0 + dr_dthetaRadians * angular_distance
    angle = angle0 + angular_distance

    return cq.Vector(
        (-r * math.sin(angle) + dr_dthetaRadians * math.cos(angle)),
        (r * math.cos(angle) + dr_dthetaRadians * math.sin(angle)),
    )


# Calculate interpolation points for the spiral:
angles = [i * math.pi / 2 for i in range(4 * 3)]
points = [point_at_angle(a) for a in angles]
tangents = [tangent_at_angle(a) for a in angles]
parameters_point_index = range(len(points))
parameters_angle = angles

# Connect the interpolation points with B-splines, using different values for
# the B-spline parameter at the interpolation points:
spiral_with_default_parameters = cq.Workplane("XY").spline(
    points, tangents=tangents, scale=True, includeCurrent=False,
)
spiral_with_point_index_parameters = cq.Workplane("XY").spline(
    points,
    tangents=tangents,
    parameters=parameters_point_index,
    scale=False,
    includeCurrent=False,
)
spiral_with_scaled10_point_index_parameters = cq.Workplane("XY").spline(
    points,
    tangents=tangents,
    parameters=[p * 10 for p in range(len(points))],
    scale=True,
    includeCurrent=False,
)
spiral_with_scaled100_point_index_parameters = cq.Workplane("XY").spline(
    points,
    tangents=tangents,
    parameters=[p * 100 for p in range(len(points))],
    scale=True,
    includeCurrent=False,
)
spiral_with_angle_parameters = cq.Workplane("XY").spline(
    points,
    tangents=tangents,
    parameters=parameters_angle,
    scale=False,
    includeCurrent=False,
)
spirals_with_scaled_parameters = (
    Workplane("XY")
    .add(spiral_with_scaled10_point_index_parameters)
    .add(spiral_with_scaled100_point_index_parameters)
)


# Save the B-splines to files:
if __name__ == "__main__":
    workplanes_dict = dict(
        [(k, v) for k, v in locals().items() if isinstance(v, cq.Workplane)]
    )
    for var_name, workplane in workplanes_dict.items():
        cq.exporters.export(workplane, f"$Demo_{__file__}-{var_name}.svg")
```

___

$Demo_spline_parameter_archimedean_spiral py-spiral_with_default_parameters
_Image above: spiral_with_default_parameters, i.e. no parameters specified when interpolating_
___
$Demo_spline_parameter_archimedean_spiral py-spiral_with_point_index_parameters
_Image above: spiral_with_point_index_parameters, i.e. parameter (at an interpolation point) = interpolation point index_
___
$Demo_spline_parameter_archimedean_spiral py-spiral_with_scaled10_point_index_parameters
_Image above: spiral_with_scaled10_point_index_parameters, i.e. parameter (at an interpolation point) = 10 × interpolation point index_
___
$Demo_spline_parameter_archimedean_spiral py-spiral_with_scaled100_point_index_parameters
_Image above: spiral_with_scaled100_point_index_parameters, i.e. parameter (at an interpolation point) = 100 × interpolation point index_
___
$Demo_spline_parameter_archimedean_spiral py-spiral_with_angle_parameters

_Image above: spiral_with_angle_parameters, i.e. parameter = spiral angle_
___
$Demo_spline_parameter_archimedean_spiral py-spirals_with_scaled_parameters
_Image above: spirals_with_scaled_parameters, i.e. two spirals overlayed: parameter (at an interpolation point) = 10 × interpolation point index, and parameter = 100 × interpolation point index._

_Don't be confused, if you don't see two spirals 😊._
___

> Regarding black, we use the defaults of `19.10b0`.

I saw that when the Jenkins test failed. I think `environment.yml` should be updated to pin the version of `black`. It's also worth mentioning in the documentation (i.e., `README.md`, unless there are other paces, too). I could write another issue for that when I get a chance.

tests/test_cadquery.py
```diff
@@ -590,6 +590,118 @@ def testSpline(self):
...
+        spline = (
+            Workplane("XY")
+            .spline(points, tangents=tangents, parameters=parameters, scale=True)
+            .consolidateWires()
+        )
+
+        test_point = spline.edges().val().positionAt(0.5, mode="parameter")
+        expected_test_point = Vector(0.375, 0.875, 0.0)
+
+        self.assertAlmostEqual((test_point - expected_test_point).Length, 0)
```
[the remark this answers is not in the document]
pavpen: Hm. I think there may be some misunderstanding. The `scale` parameter doesn't have much to do with whether the tangents have unit length, or not. The length of a tangent vector has physical meaning, i.e. it changes the shape of the curve. Let me overexplain.

If your trajectory in space is a vector-valued function of a parameter time, then the tangent to your trajectory is your velocity. Compare two situations, in both of which you were at point A at some time, and at point B one second later. If your velocity at point A is in the same direction in both situations, but is 10 times larger the first time than the second time, your trajectory in the two cases is not going to be the same, even if both times your trajectory is represented by a polynomial function of the time parameter of the same degree.

Unless you have a very uniform curve (like a circle with a parameter that's the right proportion to the radius, depending on your units of length and angle), your tangent vector is not going to have a unit length.

Setting `scale = True` makes the interpolator ignore the lengths of the tangent vectors you specify, and use only their direction. The lengths themselves are interpolated.

I realized I broke up the two test cases related to the `scale` parameter. Otherwise, they use the same points, and tangent vectors. However, the test point has different coordinates at the same parameter value. That's enough to verify that `scale = True` and `scale = False` is working properly.

I'll add a commit to put the test cases together, and  add a comment.

Here's a visualization of the test cases:

___
$Demo_spline_tangent_scaling py-spline_with_scaled_tangents
_Above: Spline with scaled tangents._
___
$Demo_spline_tangent_scaling py-spline_with_unscaled_tangents
_Above: Spline with unscaled tangents._
___
$Demo_spline_tangent_scaling py-overlayed_splines
_Above: The two splines overlaid._
___

cadquery/cq.py
pavpen: `sphinx-build` is currently failing on my machine. I'll need to figure out how to test this.

pavpen: A added a commit with an attempted fix.

adam-urbanczyk:
@pavpen thanks, second question was about the tangents. You answered it in the review comments.

tests/test_cadquery.py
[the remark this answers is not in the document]
marcus7070: > I realized I broke up the two test cases related to the `scale` parameter. Otherwise, they use the same points, and tangent vectors. However, the test point has different coordinates at the same parameter value. That's enough to verify that `scale = True` and `scale = False` is working properly.
> 
> I'll add a commit to put the test cases together, and add a comment.

Thanks, that's easier to follow now.

cadquery/cq.py
marcus7070: Read The Docs build LGTM.

marcus7070:
If we check the length of tangents and raise a ValueError, should we be doing the same for parameters? I don't think it's valid to have a different number of parameters to interpolation points.

marcus7070:
I'm a bit worried I have not handled `periodic=True` properly in that latest commit. I can't get to my computer at the moment, I'll check it in an hour or so.

marcus7070:
Damn, my code was passing black but I was using version 20.8b1 to check. Needs 8d2c0b9 to pass with 19.10.

pavpen:
> If we check the length of tangents and raise a ValueError, should we be doing the same for parameters? I don't think it's valid to have a different number of parameters to interpolation points.

👍 I see you added a commit for this.

I think it's a good idea. The OCC library should throw an exception, if we don't check, but I think Python exceptions are much easier to understand. The ones coming from C++ often don't explain the reason for the exception.

marcus7070 (approved):

adam-urbanczyk (approved):

marcus7070:
Thankyou very much @pavpen, this is a great contribution. Thanks also for the detailed comments and examples, they made reviewing much easier.

adam-urbanczyk:
+1, big thanks for the contributions and patience @pavpen .

pavpen:
Thanks for all the help.

I have a few more minor suggestions that I can write issues for.

Also, I kind of hacked the SVG exporter so I can generate the SVGs that used for the examples above. I changed the styling, so the graphic works both on bright and on dark backgrounds (like GitHub's dark theme). I was thinking it may be useful to add that as a feature, as well. Although it may require a bit of discussion.

adam-urbanczyk:
Sounds good, let's discuss in a new issue @pavpen .

PR #654, opened 2021-02-23
Fix deprecation warning on regex

RubenRubens:
This simple fix removes a warning in `pytest`.

### DIFF
--- a/cadquery/selectors.py
+++ b/cadquery/selectors.py
@@ -198,7 +198,7 @@ class ParallelDirSelector(BaseDirSelector):
         Linear Edges
         Planar Faces
 
-    Use the string syntax shortcut |(X|Y|Z) if you want to select based on a cardinal direction.
+    Use the string syntax shortcut \|(X|Y|Z) if you want to select based on a cardinal direction.
 
     Example::
 
@@ -679,7 +679,7 @@ class _SimpleStringSyntaxSelector(Selector):
     def filter(self, objectList):
         r"""
         selects minimum, maximum, positive or negative values relative to a direction
-        [+|-|<|>|] <X|Y|Z>
+        ``[+|-|<|>|] <X|Y|Z>``
         """
         return self.mySelector.filter(objectList)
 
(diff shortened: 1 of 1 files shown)


marcus7070:
These characters now get rendered as a link to something :shrug: in the HTML docs. https://cadquery--654.org.readthedocs.build/en/654/classreference.html#cadquery.selectors.ParallelDirSelector

Does pytest still throw warnings if you keep the raw string but add the backslash back in?

RubenRubens:
AFAIK `r"\|(X|Y|Z)"` would be displayed as `\|(X|Y|Z)`. So, I don't find it ideal neither. 😕 Nice catch by the way. Should we open an issue on sphinx-doc/sphinx?

marcus7070:
I agree, the best solution would be to tell Sphinx that the pipe symbol is not... whatever it thinks it is (variable substitution? But why does that make a link?). I'm sure there is some way to do that, but sphinx is so complicated and I'll take the easy way out of escaping it.

Looks like r"\|(X|Y|Z)" is rendered correctly by sphinx in my local build, so I've added a commit.

marcus7070:
This looks good to me, I can't see any more weird links:

https://cadquery--654.org.readthedocs.build/en/654/classreference.html#cadquery.ParallelDirSelector

And the pytest warnings (which didn't show up on Azure, but did show up on Travis) are gone.

Feel free to try something different if you want to @RubenRubens and I'll have another look at it, or let me know if you want to merge as is.

marcus7070 (approved):

adam-urbanczyk (approved):

jmwright:
All checks are green. Shall I merge?

marcus7070:
@jmwright, I'd just wait for @RubenRubens. If they are feeling like a deep dive into the sphinx internals, there might be a way to do this better. If they don't feel like a deep dive then this will do, I'm certainly not volunteering to figure out why sphinx thinks `|X` is a link to nowhere.

RubenRubens:
Great! Let's just merge this then.

jmwright:
Thanks @RubenRubens !

Issue #621, opened 2021-02-09
Improved error when pendingWires/pendingEdges is empty


marcus7070:
Minimal example:
```python
w = (
    cq.Workplane()
    .box(10, 10, 10)
    .faces(">Z")
    .edges()
    .extrude(3, taper=40)
)
```

Result is `IndexError: list index is out of range` in CQ-editor.

A more detailed trace from the interpreter is still pretty opaque:
```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/nix/store/sgx51ii0rf6872aj5ab5qvf5fw3kdll4-python3-3.8.6-env/lib/python3.8/site-packages/cadquery/cq.py", line 2745, in extrude
    r = self._extrude(
  File "/nix/store/sgx51ii0rf6872aj5ab5qvf5fw3kdll4-python3-3.8.6-env/lib/python3.8/site-packages/cadquery/cq.py", line 3177, in _extrude
    thisObj = Solid.extrudeLinear(ws[0], [], eDir, taper)
IndexError: list index out of range
```

To be clear, the error is caused by `ctx.pendingWires` and `pendingEdges` being empty. I need to push the results from `.edges()` to them by calling `toPending()` before extruding. 

Here are some issues where other users have run into this, and I'm sure I'm missing a few more: 
* https://github.com/CadQuery/cadquery/issues/619#issue-802350008 
* https://github.com/CadQuery/cadquery/issues/546#issue-765532217 (although this one has Edges in `pendingEdges`, just an empty `pendingWires` causing the error)

So question 1, we could use a more informative error than `IndexError` for when we expect a pendingWire or Edge but get nothing, right?

I think these methods would all fail with an IndexError if used like above:
* twistExtrude
* cutThruAll
* loft
* _extrude
* _revolve
* _sweep

You could check in each Workplane method if `ctx.pendingWires` was empty, but to save code repetition I suggest we add a new method to `CQContext` that either will either return a pending wire/edge or raise an error like `RuntimeError('Requested pending wires but none present in modelling context')`. Then change the relevant methods to use this getter instead of directly accessing `ctx.pendingWires`. Also this would need a `pop=True` keyword option, considering that almost every method does:
```python
        wires = self.ctx.pendingWires
        self.ctx.pendingWires = []
```

Question 2: Does this sound like a good approach?

jmwright:
@marcus7070 
[...]

PR #655, opened 2021-02-24
findSolid errors

marcus7070:
`Workplane.findSolid` had an incorrect docstring about it raising a ValueError when it didn't find any solids. It actually just returns None, and this is behaviour that a few other methods rely on (`union` for example). So I removed the line in the docstring.

There were several methods that treated the return from `findSolid` as just a Solid, and hence were trying to call methods on the `None` object and failing with an AttributeError. I added checks for `None` and ValueError exceptions to these methods.

Then I thought to myself, "This is mypy's job, why am I doing it?", and realised that `findSolid` needed the type hint Optional added to it's return value.

I also added tests for all the new exceptions. I realised `cutEach` isn't tested at all (except for probably in the examples), so I added a basic test for that as well.

### DIFF
--- a/tests/test_cadquery.py
+++ b/tests/test_cadquery.py
@@ -1844,18 +1844,15 @@ class TestCadQuery(BaseTest):
         """
         r = Workplane("XY").box(1, 1, 1)
         dim = r.largestDimension()
-
         self.assertAlmostEqual(1.76, dim, 1)
 
         r = Workplane("XY").rect(1, 1).extrude(1)
         dim = r.largestDimension()
-
         self.assertAlmostEqual(1.76, dim, 1)
 
         r = Workplane("XY")
-        dim = r.largestDimension()
-
-        self.assertEqual(-1, dim)
+        with raises(ValueError):
+            r.largestDimension()
 
     def testOccBottle(self):
         """
@@ -3443,9 +3440,10 @@ class TestCadQuery(BaseTest):
         self.assertEqual(len(s.Solids()), 2)
         self.assertTrue(isinstance(s, Compound))
 
-        # if no solids are found, should return None
+        # if no solids are found, should raise ValueError
         w = Workplane().hLine(1).close()
-        self.assertEqual(w.findSolid(), None)
+        with raises(ValueError):
+            w.findSolid()
 
     def testSlot2D(self):
 
--- a/tests/test_cadquery.py
+++ b/tests/test_cadquery.py
@@ -4241,14 +4241,17 @@ class TestCadQuery(BaseTest):
             w0.cutBlind(1)
 
     def testFindFace(self):
-        # if there are no faces to find, should return None
+        # if there are no faces to find, should raise ValueError
         w0 = Workplane()
-        self.assertEqual(w0.findFace(), None)
+        with raises(ValueError):
+            w0.findFace()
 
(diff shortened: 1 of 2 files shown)


marcus7070:
I forgot about the `Workplane.findFace` method, if I've added `Optional` to `findSolid` I should do it for `findFace` as well.

adam-urbanczyk:
Thanks for picking this up. Open question: isn't it better if `findSolid` raises? This would result in less code duplication.

marcus7070:
> isn't it better if `findSolid` raises? This would result in less code duplication.

Firstly, if I had of realised how large this PR was going to be I would have made an issue first and discussed this there. But honestly I was intending to take a 10 minute break from work and fix one line in the `findSolid` docstring and it just snowballed from there. :sweat_smile: 

I thought about raising an exception in `findSolid`, but:

* that's a user facing method and it's had the behaviour of returning `None` for [8 years now](https://github.com/CadQuery/cadquery/blob/e6c25473293401ebaaeddcc7f2104b4760abc1a5/cadquery/CQ.py#L376-L405), and
* it seemed like the smaller change to add a few `raise ValueError`s than to change the behaviour of `findSolid`.

On the other hand:

* if no user has complained about the error in the docstring for 8 years, it's probably rarely used and a change won't be too disruptive,
* raising in `findSolid` would bring it in line with how I intend to do `getPending` in #621 (which I'm also working on now),
* it turned out to be more than just a few `raise ValueError`s, so raising in `findSolid` would reduce a lot of duplication,
* at present `findSolid` is just an alias for `Workplane._findType((Solid, Compound))`, so it will be trivial to work around if you need the old behaviour. 

Now I've written all that out, I prefer raising in `findSolid`. @adam-urbanczyk, @jmwright, are you OK with me changing the behaviour of `findSolid` to raise an exception instead of returning `None`?

jmwright:
> are you OK with me changing the behaviour of findSolid to raise an exception instead of returning None?

I am, yes.

adam-urbanczyk:
+1

marcus7070:
`Workplane.largestDimension` previously returned -1 when it couldn't find a solid/compound. Now it also raises a ValueError.

tests/test_cadquery.py
```diff
@@ -4177,3 +4216,42 @@ def testEnd(self):
...
+            w1.cutEach(lambda loc: c.located(loc))
+
+    def testCutBlind(self):
+        # cutBlind is already tested in several of the complicated tests, so this method is short.
+        # test ValueError on no solid found
+        w0 = Workplane().hLine(1).vLine(1).close()
+        with raises(ValueError):
+            w0.cutBlind(1)
+
+    def testFindFace(self):
```
jmwright, line 4243: Do you still want to add this test based on the decision in #665 ?

jmwright (approved):
@marcus7070 Looks good, thanks!

tests/test_cadquery.py
marcus7070: Yeah, might as well keep the test while `findFace` is marked deprecated, then remove the test along with `findFace` in the next release.

adam-urbanczyk (approved):

adam-urbanczyk:
Thanks @marcus7070

PR #662, opened 2021-02-24
Simplify findSolid and findFace

RubenRubens:
These are breaking changes. I believe they are worth considering thought. Few users would be affected and makes cq code much cleaner. For example, both Fx Bricks and Paramak have 0 references to `findSolid`.

### DIFF
--- a/cadquery/cq.py
+++ b/cadquery/cq.py
@@ -655,7 +655,7 @@ class Workplane(object):
 
         return rv
 
-    def _findType(self, types, search: Literal["stack", "parents"]):
+    def _findType(self, types, search):
 
         if search == "stack":
             rv = [s for s in self.objects if isinstance(s, types)]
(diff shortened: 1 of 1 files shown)


marcus7070:
Ok, so you've simplified `findSolid` to two use cases:

1. search stack and parents
2. do not search stack but search parents

Instead of the string argument, can we just keep the `searchStack` bool argument and drop the `searchParents`?

I agree with the simplification, I've never used `findSolid(searchParents=False)`, and if I ever needed to do that job I think I'd use a more general python technique like `next(x for x in wp.vals() if isinstance(x, cq.Solid))`.

I'm about to change the behaviour of `findSolid` in #662 anyway, so good timing for this.

RubenRubens:
This might not be that interesting after all. So, don't waste time in here. I might consider reopening this PR soon with some interesting ideas. :)

PR #668, opened 2021-02-25
Adding dedicated import-export doc

jmwright:
The skeleton of a new import-export doc to help users know what formats available and how to use the API.


jmwright:
@adam-urbanczyk and @marcus7070 I think this is ready for review. Please pay special attention to the `tolerance` and `angularTolerance` parameter descriptions. I'm not sure that I grasp the nuances of each of those well enough to describe them properly.

doc/importexport.rst
```diff
@@ -0,0 +1,263 @@
...
+..  image:: _static/importexport/box_custom_options.svg
+
+Exporting STL
+##############
+
+The STL exporter is capable of adjusting the quality of the resulting mesh, and accepts the following parameters.
+
+* ``fileName`` - The path and file name to write the STL output to.
+* ``tolerance`` - A linear deflection setting which limits the distance between a curve and its tessellation. Setting this value too low will result in large meshes that can consume computing resources. Setting the value too high can result in meshes with a level of detail that is too low. Default is 0.1, which is good starting point for a range of cases.
+* ``angularTolerance`` - Angular deflection setting which limits the angle between subsequent segments in a polyline. Default is 0.1.
```
marcus7070, line 155: LGTM. I would kind of rather this information was on the `exporters.export` docstring and use automethod to stick it in here, but that is just minor.

doc/importexport.rst
```diff
@@ -0,0 +1,263 @@
...
+
+    result = cq.Workplane().box(10, 10, 10)
+
+    exporters.export(result, '/path/to/file/mesh.amf', tolerance=0.01, angularTolerance=0.1)
+
+
+Exporting TJS
+##############
+
+The TJS (ThreeJS) exporter is capable of adjusting the quality of the resulting JSON-based mesh, and accepts the following parameters.
```
marcus7070, line 194:
```suggestion
The TJS (ThreeJS) exporter produces a file in JSON format that describes a scene for the ThreeJS WebGL renderer. The objects in the first argument are converted into a mesh and then form the ThreeJS geometry for the scene. The mesh can be adjusted with the following parameters.
```

I think anyway. The terminology used in ThreeJS is significantly different to CAD. Also not helped by ThreeJS V3 JSON model format being retired and these terms changing a lot for the current V4.

marcus7070:
The last two commits were me (incompetently) trying to make git diff display the SVGs as "new file", instead of printing out a complete text diff for a file we'll never edit by hand.

doc/importexport.rst
```diff
@@ -0,0 +1,263 @@
...
+
+Importing DXF
+##############
+
+DXF files can be imported using the :py:meth:`importers.importDXF` method. There are 3 parameters that can be 
+passed to this method to control how the DXF is handled.
+
+* ``fileName`` - The path and name of the DXF file to be imported.
+* ``tol`` - The tolerance used for merging edges into wires (default: 1e-6).
+* ``exclude`` - A list of layer names not to import (default: []).
```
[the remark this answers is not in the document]
jmwright: @marcus7070 Thanks, it will be good not to have to keep those updated in two places.

doc/importexport.rst
jmwright: @marcus7070 My rst-fu is weak so I didn't know that parameters from the docstring could be included. I'll add docstrings and do the same thing you did with importDXF.

jmwright: The best I could do was add a docstring to `Shape.exportStl()` and include that. AMF and VRML are tessellated inside the `exporters()` method, so there's no place to add a docstring for them. Those parameter listings will just have to be maintained separately.  At least `exportStl()` got a docstring out of it.

jmwright:
@adam-urbanczyk The [tolerance default](https://github.com/CadQuery/cadquery/blob/dc9b0e70d15e04c6fa9eafc4f181646f4a13c774/cadquery/occ_impl/exporters/__init__.py#L38) for `exporters.export()` is 0.1, but the [default for tolerance](https://github.com/CadQuery/cadquery/blob/master/cadquery/occ_impl/shapes.py#L400) in `exportStl()` is 1e-3. It seems like `export()` is always going to override the default for `exportStl()`, which seems like it might not be intended. I wanted to bring it up to confirm.

marcus7070 (approved):

adam-urbanczyk (approved):

adam-urbanczyk:
Looks good, but you _might_ want to state that `angularTolerance` is a maximum tangent error between the mesh and the original shape. I need to look into the tolerance value, that does not block the PR, right?

https://dev.opencascade.org/doc/overview/html/occt_user_guides__mesh.html

jmwright:
> is a maximum tangent error between the mesh and the original shape.

Is that in addition to, or instead of, what's in the doc now?

adam-urbanczyk:
@jmwright after reading the docs again I conclude that I might've misinterpreted what is stated (and drawn) there. +1 for merging as-is.

marcus7070:
Thanks @jmwright!

adam-urbanczyk:
Thanks indeed, it'll be very helpful for the users!

PR #678, opened 2021-03-08
popPendingEdges and Wires

### LINKED ISSUE #621: Improved error when pendingWires/pendingEdges is empty

marcus7070:
Will close #621. To summarise:

* More informative exception when pendingWires or pendingEdges is empty
* Minimise code duplication
* More consistent method to pop pending Edges/Wires

I'm keeping this a draft PR until I see the test coverage, if it goes down then I'll fix that.

### DIFF
--- a/cadquery/cq.py
+++ b/cadquery/cq.py
@@ -93,6 +93,30 @@ class CQContext(object):
         self.tolerance = 0.0001  # user specified tolerance
         self.tags = {}
 
+    def popPendingEdges(self, errorOnEmpty: bool = True) -> List[Edge]:
+        """
+        Get and clear pending edges.
+
+        :raises ValueError: if errorOnEmpty is True and no edges are present.
+        """
+        if errorOnEmpty and not self.pendingEdges:
+            raise ValueError("No pending edges present")
+        out = self.pendingEdges
+        self.pendingEdges = []
+        return out
+
+    def popPendingWires(self, errorOnEmpty: bool = True) -> List[Wire]:
+        """
+        Get and clear pending wires.
+
+        :raises ValueError: if errorOnEmpty is True and no wires are present.
+        """
+        if errorOnEmpty and not self.pendingWires:
+            raise ValueError("No pending wires present")
+        out = self.pendingWires
+        self.pendingWires = []
+        return out
+
 
 class Workplane(object):
     """
--- a/cadquery/cq.py
+++ b/cadquery/cq.py
@@ -2122,6 +2122,8 @@ class Workplane(object):
 
     def _consolidateWires(self) -> List[Wire]:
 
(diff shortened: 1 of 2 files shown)


marcus7070:
Just FYI, I tried two ways of solving this. 

1. A single method, `def getPending(edges: bool = False, wires: bool = True):`
2. The two methods, `popPendingEdges` and `popPendingWires`, that you can see in this PR.

The first way needed a lot of overloads and casts to satisfy mypy. The experience has made me realise that two methods is the better solution, each method returning a single type.

jmwright (approved):
Thanks @marcus7070 !

adam-urbanczyk (approved):

Issue #684, opened 2021-03-11
How to cut a lip on a case seam?


fedorkotov:
**What is the proper way of making lips on a case seam like this?**
![image](https://user-images.githubusercontent.com/1106650/110783040-eedbca00-8278-11eb-9988-2b0bc2bc1d81.png)
Note that there is a gap between lips of bottom and top half that is necessary to account for 3D printing inaccuracies.

Below I describe how I do it. Maybe there is a better way? 
I thought about sweeping lip cross section along seam boundary but did not find a way to do that without manually redrawing sweeping path which would be tedious for anything but a simple box with sharp corners. 
Or maybe someone knows a library that does this automatically?

My method is
1. Create closed case shell
![image](https://user-images.githubusercontent.com/1106650/110786710-82af9500-827d-11eb-8ecf-188b1092a979.png)
1. Cut it into overlapping halves   
![image](https://user-images.githubusercontent.com/1106650/110786471-3c5a3600-827d-11eb-99cb-2f3af773d903.png)
1. Manually create "instrument" bodies to cut half (or slightly more) of overlapping parts to produce a lip 
![image](https://user-images.githubusercontent.com/1106650/110786181-dff71680-827c-11eb-8159-ab78dbe17584.png)

(see code below)

![image](https://user-images.githubusercontent.com/1106650/110785400-f81a6600-827b-11eb-8b24-046f0a6524eb.png)

```python
def get_inner_volume(wp):
    return\
      wp\
      .rect(50,50)\
      .extrude(50)\
      .edges()\
      .fillet(10)      
    

def split_bottom(
      wp,
      cut_zoffset):
    return\
        wp\
        .workplane(cut_zoffset)\
        .split(keepBottom=True)
        
def split_top(
      wp,
      cut_zoffset):
    return\
        wp\
        .workplane(cut_zoffset)\
        .split(keepTop=True)        
        
def z_slice(
        wp,
        min_z_offset,
        max_z_offset):
    
    delta_z = \
        max_z_offset-min_z_offset
    
    # two calls to split_top
    # because faces(..).workplane()
    # changes Z orientation
    
    return \
        split_top(
            split_top(
                wp,
                min_z_offset)\
              .faces("<Z"),
            -delta_z)
        
def filled_shell(wp, thickness):
    return \
        get_inner_volume(wp)\
        .union(
            get_inner_volume(wp)
            .shell(thickness))
    

def get_case_bottom(wp):
    return\
      split_bottom(
[...]

Issue #646, opened 2021-02-19
Length selector


marcus7070:
I think it's worth making a length selector (Nth selector style, since that also covers min/max with `[0]`/`[-1]`).

The [Paramak project have their own version](https://github.com/ukaea/paramak/blob/785c3ed7304e22eac7d58bb1bdc6515fbd20b9a8/paramak/utils.py#L696-L736) and so does [cq-kit](https://github.com/michaelgale/cq-kit/blob/fd3868593d9610446a4400ab64b08eddd895ba79/cqkit/cq_selectors.py#L254-L288).

A length selector also would help out in #638, where @mbway had to select the inner of two concentric b-splines. 

Thumbs up for yes, thumbs down if you think it's just bloat.

jmwright:
+1 from me

Tagging @michaelgale (cq-kit) and @Shimwell (Paramak) so they can add their thoughts if they want.

Shimwell:
Yes please that sounds most useful. I have the situation where I have to find edges and I don't know much about it. Having a selector for the Nth largest and Nth smallest would be most useful.

I tend to search for these edges and then perform operations like fillet and currently the Paramak has a selector for a given length +- a tolerance which we would be delighted to move to CQ if it is better placed there.

The only complication I can think of is that sometimes the edges are the same length. So perhaps it would be interesting to have another filter that returns edges that are the same length (within a tolerance). Otherwise it might be a bit unpredictable as to which of the similar length edges are returned.

I think that the creation of selectors in these two code bases that rely on CQ is an indication that such tools are universally useful. I would happily use a selector from CQ and remove the one in the Paramak.

Keen to help with this effort if I can be of assistance.

ps great work @michaelgale :light_rail: :train: :train: :train: :train:

michaelgale:
Yes, a LengthSelector is indeed a useful tool and often used and therefore a helpful addition to CQ.  My only concern in terms of CQ-Kit is the namespace collision of the `LengthSelector` classname.  Since CQ-Kit uses the same classname there could be issues depending on the order of imports etc.

I could rename my base classname `LengthSelector` to avoid conflict if that is the name the CQ project would like to adopt.  Either way, let me know so that I can make changes.  It looks like it will be a painless and non-breaking change for CQ-Kit.
p.s. thanks @Shimwell !

fedorkotov:
[...]

PR #688, opened 2021-03-14
Added AreaNthSelector

fedorkotov:
Added AreaNthSelector that is useful for nested features selection. Especially to select one of coplanar nested wires for subsequent extrusion, cutting or filleting.

This selector was developed as a result of discussion in #684

### DIFF
--- a/cadquery/selectors.py
+++ b/cadquery/selectors.py
@@ -453,6 +453,64 @@ class DirectionNthSelector(ParallelDirSelector, CenterNthSelector):
         objectlist = _NthSelector.filter(self, objectlist)
         return objectlist
 
+class AreaNthSelector(_NthSelector):
+    """
+    Selects object(s) with Nth area.
+
+    Applicability:
+        Faces, Shells, Solids - Shape.Area() is used to compute area
+        planar Wires - a temporary face is created to compute area
+
+    Among other things can be used to select one of 
+    the nested coplanar wires or faces. 
+
+    For example to create a fillet on a shank:
+
+        cq.Workplane("XY")\
+            .circle(5)\
+            .extrude(2)\
+            .circle(2)\
+            .extrude(10)\
+            .faces(">Z[-2]")\
+            .wires(AreaNthSelector(0))\
+            .fillet(2)
+
+    Or to create a lip on a case seam:
+        
+        cq.Workplane("XY")\
+            .rect(20, 20)\
+            .extrude(10)\
+            .edges("|Z or <Z")\
+            .fillet(2)\
+            .faces(">Z")\
+            .shell(2)\
+            .faces(">Z")\
+            .wires(AreaNthSelector(-1))\
+            .toPending()\
+            .workplane()\
+            .offset2D(-1)\
+            .extrude(1)\
+            .faces(">Z[-2]")\
+            .wires(AreaNthSelector(0))\
+            .toPending()\
+            .workplane()\
+            .cutBlind(2)    
+    """    
+    def key(self, obj: Shape) -> float:        
+        shape_type = obj.ShapeType()
+
+        if shape_type in ("Face", 
+                          "Shell", 
+                          "Solid"):
+            return obj.Area()
+        elif shape_type == "Wire":
+            return Face.makeFromWires(obj).Area()
+        else:
+            raise TypeError(
+                "AreaNthSelector supports only Wires, "\
+                "Faces, Shells and Solids, not {}"\
+                .format(shape_type))       
+
 
 class BinarySelector(Selector):
     """
--- a/cadquery/selectors.py
+++ b/cadquery/selectors.py
@@ -47,7 +47,7 @@ from pyparsing import (
     Keyword,
 )
 from functools import reduce
-from typing import List, Union, Sequence, cast
+from typing import List, Union, Sequence
 
 
 class Selector(object):
--- a/cadquery/selectors.py
+++ b/cadquery/selectors.py
@@ -453,6 +453,7 @@ class DirectionNthSelector(ParallelDirSelector, CenterNthSelector):
         objectlist = _NthSelector.filter(self, objectlist)
         return objectlist
 
+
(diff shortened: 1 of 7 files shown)


fedorkotov:
Examples for Solids and Shells are a bit contrived. Is there a way to get more than one Solid or Shell to select from while using Workplane? Or maybe using this selector for Solids and Shells makes no sense and support for them should be removed?

fedorkotov:
Black is complaining about formatting. IMHO it makes the code **less readable** by merging long sequences of fluent API calls that I format as one call per line into long lines with too many calls. But if that is the accepted formatting style of the project I will implement black's recommendations

marcus7070:
> Black is complaining about formatting. IMHO it makes the code **less readable** by merging long sequences of fluent API calls that I format as one call per line into long lines with too many calls. But if that is the accepted formatting style of the project I will implement black's recommendations

Yes please @fedorkotov. I know black isn't perfect, but it's much easier to read one consistent style.

fedorkotov:
Two questions
1. `AreaNthSelector` is very similar to `LengthNthSelector` proposed in #646. Maybe I should add length selector in this pull request? 
2. Maybe I should add `A` (area) and `L` (length) should be added to string selectors grammar with allowed usage like this
   1. `>A`  - objects with the largest area
   2. `<L[-2]`  - objects with second smallest length

Or is it better to keep things separated and both of these additional features should have their separate pull requests?

marcus7070:
> 1. `AreaNthSelector` is very similar to `LengthNthSelector` proposed in #646.

I was thinking the same thing so I actually did #646 earlier today, thinking that the process would help me review this PR. I just put up the code as #690.

> 2. Maybe I should add `A` (area) and `L` (length) to string selectors grammar with allowed usage like this
>    
>    1. `>A`  - objects with the largest area
>    2. `<L[-2]`  - objects with second smallest length
> 
> Or is it better to keep things separated and both of these additional features should have their separate pull requests?

In this case I think we should do another PR, because we can roll area, length and radius selectors into one review.

fedorkotov:
> I was thinking the same thing so I actually did #646 earlier today, thinking that the process would help me review this PR. I just put up the code as #690.

Ok. I will include LengthNthSelector into my pull request.
But first one more question. I see in your `LengthNthSelector` you use `isinstance(obj, (...))` to determine what kind of `Shape` was passed into `_NthSelector.key(..)` implementation. I used `Shape.ShapeType()` in my `AreaNthSelector` for the same purpose. I assumed that comparing strings would be faster than type information manipulation but did not measure that. Should I change to `isinstance`? Or are these variants roughly equivalent?

> In this case I think we should do another PR, because we can roll area, length and radius selectors into one review.

Ok.

marcus7070 (changes requested):
You can include the LengthNthSelector code from #690 in here if you want, or we can work on #690 after this gets merged (which would be less work for you), I don't mind.

marcus7070:
> Unit tests for Solids and Shells are a bit contrived. Is there a way to get more than one Solid or Shell to select from while using Workplane?

Something like
```python
shell = (
    cq.Workplane()
    .box(1, 1, 1)
    .faces(">Z")
    .shell(0.1)
    .val()
)
workplane_shells = (
    cq.Workplane()
    .rarray(2, 2, 3, 4)
    .eachpoint(lambda loc: shell.located(loc))
)
```

You can also union different sized shells/solids, which would give you more variety to test an area selector on.

>  Or maybe using this selector for Solids and Shells makes no sense and support for them should be removed?

Solids and shells have clearly defined areas, so I support this selector working on them.

cadquery/selectors.py
```diff
@@ -454,6 +454,64 @@ def filter(self, objectlist: Sequence[Shape]) -> List[Shape]:
...
+            .wires(AreaNthSelector(-1))\
+            .toPending()\
+            .workplane()\
+            .offset2D(-1)\
+            .extrude(1)\
+            .faces(">Z[-2]")\
+            .wires(AreaNthSelector(0))\
+            .toPending()\
+            .workplane()\
+            .cutBlind(2)
```
marcus7070, line 498: A black compatible version here too please.

marcus7070:
If you check https://raw.githubusercontent.com/CadQuery/cadquery/master/doc/classreference.rst you'll see where we need to add `AreaNthSelector` to make it appear in the list of selectors at https://cadquery--688.org.readthedocs.build/en/688/classreference.html#selector-classes

fedorkotov:
- Switched Shape.ShapeType() to isinstance to determine what kind of Shape was passed into _NthSelector.key(..) implementation in AreaNthSelector for consistency with other selectors based on _NthSelector
- Added LengthNthSelector developed by @marcus7070 in #690 with somewhat modified unit tests
- Added new selector references to documentation
- Black compatible code examples in AreaNthSelector docstring
- Explicitly marked _NthSelector class and _NthSelector.key(..) method as abstract using standard abc package


Can't figure out what caused Windows pipelines to fail. Don't have access to Windows machine right now.

The log says 
> DLL load failed: The specified module could not be found

But which module and which dll? And why?
Have you experienced something like that before?

marcus7070 (approved):
I think all our Windows CI are failing, it's not due to your code. Once we get that fixed and can see all the tests pass I think this looks good to merge.

Thanks for doing this @fedorkotov, there is a lot of work here especially the tests.

fedorkotov:
Forgot to check for non-planar or non-closed wires. Fixed. 
For that I catch all exceptions that  `Face.makeFromWires(..)` might throw. In general catching everything is a code smell. Should I catch something more specific?

adam-urbanczyk (approved):

adam-urbanczyk:
@fedorkotov I propose to leave the catch as-is for now. `makeFromWires` needs some refactoring anyhows.

marcus7070 (approved):
Good catch with that non-planar wire test.

marcus7070:
Thanks @fedorkotov!

PR #690, opened 2021-03-15
Length selector

### LINKED ISSUE #646: Length selector

marcus7070:
Will close #646.

### DIFF
--- a/cadquery/selectors.py
+++ b/cadquery/selectors.py
@@ -358,6 +358,21 @@ class _NthSelector(Selector):
         return clustered
 
 
+class LengthNthSelector(_NthSelector):
+    """
+    Select the object(s) with the Nth length
+
+    Applicability:
+        All Edge and Wire objects.
+    """
+
+    def key(self, obj:Shape) -> float:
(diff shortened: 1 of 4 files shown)


jmwright:
@marcus7070 @adam-urbanczyk My Windows builds have been failing this morning (Monday) with the same error about not being able to load the OCP dll.

marcus7070:
Code added in #688.

PR #710, opened 2021-04-01
Check if wires are coplanar

### LINKED ISSUE #709: Placing a Circle On Each Face of Cube and Extruding Results In Invalid Solid
The following code leads to a mangled tessellation in the 3D view of CQ-editor. Exporting as a STEP file leads to an invalid file according to FreeCAD.

```python
result = (cq.Workplane("XY")
            .box(10.0,10.0,10.0,centered=(True,True,True))
            .faces()
            .circle(1.0,forConstruction=False)
            .extrude(1.0,combine=True,clean=False))
show_object(result)
```
I would have expected that to work at least with the `combine=True`, even if it's not a standard way to use that selector. Here is the result in CQ-editor 0.2.

Screenshot from 2021-03-31 15-41-46

And here is the result in the latest master of CQ-editor (OCCT 7.5.1).

Screenshot from 2021-03-31 17-48-50

Here it is with the box shown as debug for reference.

Screenshot from 2021-03-31 17-51-29

I have tried different values for `combine` and `clean`, and have also tried different `centerOption` values. I get the same results.

adam-urbanczyk:
Will resolve #709

### DIFF
--- a/cadquery/occ_impl/shapes.py
+++ b/cadquery/occ_impl/shapes.py
@@ -2095,12 +2095,21 @@ class Face(Shape):
         Makes a planar face from one or more wires
         """
 
+        # check if wires are coplanar
+        ws = Compound.makeCompound([outerWire] + innerWires)
+        if not BRepLib_FindSurface(ws.wrapped, OnlyPlane=True).Found():
+            raise ValueError("Cannot build face(s): wires not planar")
+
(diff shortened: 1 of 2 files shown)


jmwright (approved):
Thanks!

marcus7070 (approved):
Looks great, thanks @adam-urbanczyk.

Issue #479, opened 2020-10-13
Assemblies with Two Planar Constraints


jmwright:
I was attempting to put together an example so I could ask about adding offsets to planar constraints, and came across what seems like a constraint solving bug, at least on the surface.
This code works fine:
```python
import cadquery as cq
from cadquery import Vector, Color, Assembly, Location

w = 10
d = 10
h = 10

part1 = cq.Workplane().box(2*w,2*d,h)
part2 = cq.Workplane().box(w,d,2*h)

assy = (
    Assembly(part1, name="FIRST")
    .add(part2, color=Color(0,0,1,0.5), name="SECOND")
)

assy.constrain("FIRST@faces@>Z", "SECOND@faces@<Z", "Plane")
# assy.constrain("FIRST@faces@>Y", "SECOND@faces@<Y", "Plane")

assy.solve()

show_object(assy)
```
![Screenshot from 2020-10-13 14-07-12](https://user-images.githubusercontent.com/1015439/95898686-6dc9e680-0d5d-11eb-8f3a-9ef528369aac.png)
But this code ends up with with the second part inside the first with seemingly no relation to the intended relationship to the first part or the origin.
```python
import cadquery as cq
from cadquery import Vector, Color, Assembly, Location

w = 10
d = 10
h = 10

part1 = cq.Workplane().box(2*w,2*d,h)
part2 = cq.Workplane().box(w,d,2*h)

assy = (
    Assembly(part1, name="FIRST")
    .add(part2, color=Color(0,0,1,0.5), name="SECOND")
)

assy.constrain("FIRST@faces@<Z", "SECOND@faces@<Z", "Plane")
assy.constrain("FIRST@faces@>Y", "SECOND@faces@<Y", "Plane")

assy.solve()

show_object(assy)
```
![Screenshot from 2020-10-13 14-09-21](https://user-images.githubusercontent.com/1015439/95899314-5fc89580-0d5e-11eb-8997-744e688fb736.png)

I've tried multiple other variations, and each set of two similar planar constraints cause this behavior. I would expect the assembly to solve to something where no part is inside another, but maybe I'm missing something obvious that makes my constraints unsolvable.

marcus7070:
When I recreate your last example @jmwright, I consistently get the blue box up the other way.

![screenshot2020-10-14-092402](https://user-images.githubusercontent.com/50230945/95924580-5f340b80-0dff-11eb-9909-e22de506b778.png)

[...]

PR #712, opened 2021-04-03
InPlane constraint

marcus7070:
I've added a constraint that positions a point within a 2D plane.

Proper docs are coming in a later PR, but I want to update SphinxCadquery to handle assemblies first.

### DIFF
--- a/cadquery/assembly.py
+++ b/cadquery/assembly.py
@@ -173,8 +173,8 @@ class Assembly(object):
         """
         construct an assembly
 
-        :param obj: root object of the assembly (deafault: None)
-        :param loc: location of the root object (deafault: None, interpreted as identity transformation)
+        :param obj: root object of the assembly (default: None)
+        :param loc: location of the root object (default: None, interpreted as identity transformation)
         :param name: unique name of the root object (default: None, reasulting in an UUID being generated)
         :param color: color of the added object (default: None)
         :return: An Assembly object.
@@ -230,7 +230,7 @@ class Assembly(object):
         Add a subassembly to the current assembly.
 
         :param obj: subassembly to be added
-        :param loc: location of the root object (deafault: None, resulting in the location stored in
+        :param loc: location of the root object (default: None, resulting in the location stored in
           the subassembly being used)
         :param name: unique name of the root object (default: None, resulting in the name stored in
           the subassembly being used)
@@ -251,7 +251,7 @@ class Assembly(object):
         Add a subassembly to the current assembly with explicit location and name.
 
         :param obj: object to be added as a subassembly
-        :param loc: location of the root object (deafault: None, interpreted as identity
+        :param loc: location of the root object (default: None, interpreted as identity
           transformation)
         :param name: unique name of the root object (default: None, resulting in an UUID being
           generated)
@@ -291,7 +291,7 @@ class Assembly(object):
 
     def _query(self, q: str) -> Tuple[str, Optional[Shape]]:
         """
-        Execute a selector query on the assembly. 
+        Execute a selector query on the assembly.
         The query is expected to be in the following format:
 
             name[?tag][@kind@args]
@@ -443,9 +443,9 @@ class Assembly(object):
     ) -> "Assembly":
         """
         save as STEP or OCCT native XML file
-        
+
         :param path: filepath
-        :param exportType: export format (deafault: None, results in format being inferred form the path)
+        :param exportType: export format (default: None, results in format being inferred form the path)
         """
 
         if exportType is None:
--- a/cadquery/assembly.py
+++ b/cadquery/assembly.py
@@ -5,7 +5,7 @@ from uuid import uuid1 as uuid
 
 from .cq import Workplane
 from .occ_impl.shapes import Shape, Face, Edge
-from .occ_impl.geom import Location, Vector
+from .occ_impl.geom import Location, Vector, Plane
 from .occ_impl.assembly import Color
 from .occ_impl.solver import (
     ConstraintSolver,
@@ -106,6 +106,14 @@ class Constraint(object):
 
         return rv
 
+    def _getPlane(self, arg: Shape) -> Plane:
+
+        # This is very duck typing. Might want to restrict types later to
+        # prevent confusion eg. making a plane from an spline Edge.
+        normal = self._getAxis(arg)
+        origin = arg.Center()
+        return Plane(origin, normal=normal)
+
     def toPOD(self) -> ConstraintPOD:
         """
         Convert the constraint to a representation used by the solver.
@@ -113,7 +121,7 @@ class Constraint(object):
 
         rv: List[Tuple[ConstraintMarker, ...]] = []
 
-        for arg, loc in zip(self.args, self.sublocs):
+        for idx, (arg, loc) in enumerate(zip(self.args, self.sublocs)):
 
             arg = arg.located(loc * arg.location())
 
@@ -123,6 +131,11 @@ class Constraint(object):
                 rv.append((arg.Center().toPnt(),))
             elif self.kind == "Plane":
                 rv.append((self._getAxis(arg).toDir(), arg.Center().toPnt()))
+            elif self.kind == "InPlane":
+                if idx == 0:
+                    rv.append((self._getPlane(arg).toPln(),))
+                else:
+                    rv.append((arg.Center().toPnt(),))
             else:
                 raise ValueError(f"Unknown constraint kind {self.kind}")
 
--- a/cadquery/assembly.py
+++ b/cadquery/assembly.py
@@ -18,7 +18,7 @@ from .selectors import _expression_grammar as _selector_grammar
 
 # type definitions
 AssemblyObjects = Union[Shape, Workplane, None]
-ConstraintKinds = Literal["Plane", "Point", "Axis", "InPlane"]
+ConstraintKinds = Literal["Plane", "Point", "Axis", "PointInPlane"]
 ExportLiterals = Literal["STEP", "XML"]
 
 PATH_DELIM = "/"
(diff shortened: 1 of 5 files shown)


marcus7070:
Codecov pointed out I only test when the plane is locked and the point is free, I should test the other way around too.

marcus7070:
> Codecov pointed out I only test when the plane is locked and the point is free, I should test the other way around too.

At the moment this constraint requires the first object to be a plane (or at least quack like one), and the second object to be a point. I started changing the code so the constraint could be specified either way around but I didn't like what I was writing. I would have to type check to figure out which argument is a plane, and in doing so prevent the use case of constraining the centre of one face to the plane formed by another face. It was also ugly code.

At the moment the responsibility is on the user to specify the plane as the first object and the point as the second.

I'll keep it in the back of my head and see if I can come up with a better way to write these non-symetrical constraints (plane/point vs the symetrical axis/axis and point/point constraints).

marcus7070:
BTW @jmwright, I think this constraint does what you expected "Plane" to do back in #479?

jmwright:
> BTW @jmwright, I think this constraint does what you expected "Plane" to do back in #479?

Yes, this seems like what I was looking for. I'm not sure when I'll get a chance to do it, but I plan to go back to your spindle assembly and retry mounting the spindle base plate to the aluminum extrusion with this constraint in place.

https://raw.githubusercontent.com/marcus7070/spindle-assy-example/master/screenshot.png

Melding both the way I would do it in something like SolidWorks and CadQuery's existing constraints, I expect this is what I'll try.

1. inPlane of "<Y" on extrusion and ">Y" on base plate
2. inPlane of "<Z" on extrusion and "<Z" on base plate
3. Use translate to create an offset for the base plate until it sort of looks like it's in the correct position away from the end of the extrusion
4. Create an axial constraint to (hopefully) center the base plate in the X axis in relation to the extrusion - not sure if this will work

jmwright:
Got a chance to play around with this a little bit and I like it. I expect I will use this constraint constantly when working with assemblies in CQ. I created the following simplified assembly to accomplish what caused my confusion in #479 

```python
import cadquery as cq

block1 = cq.Workplane("XY").box(200, 100, 20)
block2 = cq.Workplane("XY").box(50, 100, 10)

assy = cq.Assembly()

assy.add(block1, name="block1", color=cq.Color(1.0, 0, 0, 1.0))
assy.add(block2, name="block2", color=cq.Color(0.0, 0, 1, 1.0))

assy.constrain("block1@faces@>Z", "block2@faces@<Z", "InPlane", param=0)
assy.constrain("block1@faces@>X", "block2@faces@>X", "InPlane", param=20)
assy.constrain("block1@faces@<X", "block2@faces@>X", "Axis")

assy.solve()

show_object(assy)
```
Results in:

Screenshot from 2021-04-06 09-01-21

I do have two questions though.

* Why is `param` named that, instead of something like `offset`?
* What if I want to use a negative `param`, say to have one part overhang another? In the example below I've arranged the Y face inPlane constraint in such a way that I "push" the block2 off the the other side, but what if I wanted to apply a negative offset to "pull" it to hang off the opposite edge? Or if I want two plates to face each other, but to be floating 5mm apart? This is probably a moot point because you should always be able to set the constraint up to achieve the desired effect, but I thought I'd bring it up anyway to be thorough.

```python
import cadquery as cq

block1 = cq.Workplane("XY").box(200, 100, 20)
block2 = cq.Workplane("XY").box(50, 80, 10)

assy = cq.Assembly()

assy.add(block1, name="block1", color=cq.Color(1.0, 0, 0, 1.0))
assy.add(block2, name="block2", color=cq.Color(0.0, 0, 1, 1.0))

assy.constrain("block1@faces@>Z", "block2@faces@<Z", "InPlane", param=0)
assy.constrain("block1@faces@>X", "block2@faces@>X", "InPlane", param=20)
assy.constrain("block1@faces@>Y", "block2@faces@>Y", "InPlane", param=30)
#assy.constrain("block1@faces@<X", "block2@faces@>X", "Axis")

assy.solve()

show_object(assy)
```

Screenshot from 2021-04-06 09-09-54

cadquery/assembly.py
```diff
@@ -211,12 +224,15 @@ def add(
...
         """
-        add a subassembly to the current assembly.
-        
+        Add a subassembly to the current assembly.
+
         :param obj: subassembly to be added
-        :param loc: location of the root object (deafault: None, resulting in the location stored in the subassembly being used)
-        :param name: unique name of the root object (default: None, resulting in the name stored in the subassembly being used)
-        :param color: color of the added object (default: None, resulting in the color stored in the subassembly being used)
+        :param loc: location of the root object (deafault: None, resulting in the location stored in
```
jmwright, line 230: Extra a in "default" here

jmwright (approved):
Looks great, thanks @marcus7070 !

marcus7070:
> * What if I want to use a negative `param`, say to have one part overhang another? In the example below I've arranged the Y face inPlane constraint in such a way that I "push" the block2 off the the other side, but what if I wanted to apply a negative offset to "pull" it to hang off the opposite edge? Or if I want two plates to face each other, but to be floating 5mm apart? This is probably a moot point because you should always be able to set the constraint up to achieve the desired effect, but I thought I'd bring it up anyway to be thorough.


If you have a look at the function that is minimized: https://github.com/CadQuery/cadquery/blob/a71a93ea274089ddbd48dbbd84d84710fc82a432/cadquery/occ_impl/solver.py#L102-L104

(`param` winds up as `val` in that), the solver shuffles around the locations until that returned value is minimized (normally it winds up as 0). So when it comes to distance relationships it is an offset, but it doesn't have any inherent direction, so I think with pnt to pnt constraint you wind up with a sphere around the fixed pnt where the function would return 0.

So InPlane with a `param != 0` would be minimised at two parallel planes, offset `param` from the argument. To make sure my part winds up in the plane I want, I position the part roughly (but definitely on the correct side) before solving (with a location argument in `Assembly.add`) and trust in the solver to hit the nearest minimum and not jump over it to the other side.

An alternative is to use the 5 argument `Assembly.constrain` (instead of 3 string arguments), where you can do things like `part.vertices(">X and >Y and >Z").val().translate((5, 0, 0))`.

marcus7070:
In case there is any misunderstanding I'll just make this clear.

```python
assy.constrain("block1@faces@>Z", "block2@faces@<Z", "InPlane", param=0)
assy.constrain("block1@faces@>X", "block2@faces@>X", "InPlane", param=20)
assy.constrain("block1@faces@>Y", "block2@faces@>Y", "InPlane", param=30)
```

I'm planning to make this quite clear in the docs, but the InPlane relationship is between a plane and a point. I've just duck typed it so that the faces you are selecting on block2 can be represented as a point.

adam-urbanczyk:
Still need to read the code but three general things pop up.

1. given the naming I'd expect first thing to be "point" and the second thing to be plane ( ... `inPlane`)
2. wires seem not to be supported, though you can [often] construct a plane from a wire
3. I'd propose to make the param sign sensitive - you can offset the plane by `d*normal` and optimize to zero

marcus7070:
> 1. given the naming I'd expect first thing to be "point" and the second thing to be plane ( ... `inPlane`)

Sure, the order was always pretty arbitrary. 

This was in the back of my mind while writing this PR: the order of arguments in a constraint is important, right? eg.

```python
assy = cq.Assembly()

assy.add(cq.Workplane().box(1, 1, 1), name="box")
assy.add(cq.Workplane().sphere(1), name="sphere")

assy.constrain(
    "sphere",
    cq.Vertex.makeVertex(0, 0, 1),
    "box",
    cq.Vertex.makeVertex(0, 0, 0),
    "Point",
)

assy.solve()
show_object(assy)
```
screenshot2021-04-08-153430

But when you flip the constraint order to:

```python
assy.constrain(
    "box",
    cq.Vertex.makeVertex(0, 0, 0),
    "sphere",
    cq.Vertex.makeVertex(0, 0, 1),
    "Point",
)
```

screenshot2021-04-08-153516

So a user could need the reverse of this constraint, ie. `"PointOnPlane"` and `"PlaneOnPoint"`. 

> 2. wires seem not to be supported, though you can [often] construct a plane from a wire

Thanks, I had planned to do wires but forgot.

Should I extend `Constraint._getAxis` to also handle Wires?

> 3. I'd propose to make the param sign sensitive - you can offset the plane by `d*normal` and optimize to zero

Oh, so I can! That's going to be so much better, thanks! :rocket:

marcus7070:
I now handle Wires and introduced tests for _getPlane.

I decided to make a plane from any Face (as opposed to checking for planar Faces first). Not 100% confident in that descision, let me know if you have any opinions.

Still to do:

- [ ] Use signed `param`
- [ ] Think about how to do the `"PointOnPlane"`/`"PlaneOnPoint"` without repeating code, and considering there will probably be more "non-symetrical" constraints coming along later.

cadquery/assembly.py
marcus7070: :+1:

marcus7070:
I said above:

> This was in the back of my mind while writing this PR: the order of arguments in a constraint is important, right? eg.
> ...
> So a user could need the reverse of this constraint, ie. `"PointOnPlane"` and `"PlaneOnPoint"`.

I now think the order of arguments to constraints is not important. 

What was happening in my examples was that I was starting with an empty assembly (`assy = cq.Assembly()`), so the first entity of the first constraint was set to locked by: https://github.com/CadQuery/cadquery/blob/a71a93ea274089ddbd48dbbd84d84710fc82a432/cadquery/assembly.py#L379-L396

If the user wants a part to be locked, then they should specify it like `assy = cq.Assembly(part_to_be_locked, name="part")`, and then the order of constraint arguments shouldn't matter and there is no need for a "PlaneOnPoint" constraint.

marcus7070:
I think I'm done now, so ready for you to look at @adam-urbanczyk. @jmwright you can check out the changes if you like.

adam-urbanczyk (approved):

Issue #717, opened 2021-04-06
Location silently inits to an empty location if the args are not understood


marcus7070:
There are a lot of places in our code where a 3-tuple can be used in place of a vector. The Location init is not one of them. `Location((1, 2, 3))` silently produces an empty location.

I think the behaviour of `Location()` producing empty locations is good, however if an argument is passed in but can't be used, then it should probably raise a TypeError.

marcus7070:
Done in #723.

PR #722, opened 2021-04-13
Prevent location tuple

RubenRubens:
Let me know if this was what you had on mind @marcus7070. 

PR requested on #717 .


fedorkotov:
I don't know for sure if @marcus7070 meant that tuple should be accepted or that exception should be raised when it is passed. But either way I think that `if ... elif ..` chain should end with `else` (not `elif`) which raises an exception when anything of unsupported type is passed.

fedorkotov:
Ups. This PR that was related to  #717 was closed while I was writing my comment.

PR #723, opened 2021-04-13
raise TypeError when Location param is a tuple

RubenRubens:
Let me know if this was what you had on mind @marcus7070.

PR requested on #717 .

### DIFF
--- a/cadquery/occ_impl/geom.py
+++ b/cadquery/occ_impl/geom.py
@@ -928,6 +928,12 @@ class Location(object):
                 return
             elif isinstance(t, gp_Trsf):
                 T = t
+            elif isinstance(t, (tuple, list)):
+                raise TypeError(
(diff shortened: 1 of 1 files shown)


fedorkotov:
(copying my comment from now closed #722)
I don't know for sure if @marcus7070 meant that tuple should be accepted or that exception should be raised when it is passed. But either way I think that `if ... elif ..` chain should end with `else` (not `elif`) which raises an exception when anything of unsupported type is passed.

RubenRubens:
Good point @fedorkotov. I have just choose to end with `elif` because this is tuple specific issue. But it can end with an `else` as you said. Thanks!

marcus7070 (approved):
Looks good, thanks @RubenRubens!

jmwright (approved):
Thanks @RubenRubens !

PR #727, opened 2021-04-16
Minor Comment on Documentation

andreubernadserra:
Expected cutout in make_panel function for "Initial assembly". Gathered from "Final Result".


jmwright (approved):
Thanks @RubenRubens I see now that `SLOT_D` was specified in one of the individual steps, but was not included in the final result. Without that the example gave an error.

PR #733, opened 2021-04-18
Remove not used file about Pypy

RubenRubens:
Is `registering.txt` relevant today? Back in the day it was valuable to Dave Cowden. Maybe we should just remove it.


jmwright:
Thanks again @RubenRubens

PR #734, opened 2021-04-18
Remove authors file

RubenRubens:
Currently, there are only two authors: Dave Cowden and Jeremy Wright. There is no mention to Adam Urbańczyk. So, we can add Adam to the file or just remove it. Anyone can already see the authors of CadQuery [here](https://github.com/CadQuery/cadquery/graphs/contributors).

Don't get me wrong, authors are import and the three leaders of CadQuery deserve recognition for their work. But I just feel like we should remove the file and move on.

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -1,4 +1,0 @@
-# Core CQ Developers
-
-* [Dave Cowden](https://github.com/dcowden), Creator - Lead Developer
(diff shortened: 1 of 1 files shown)


jmwright:
> But I just feel like we should remove the file and move on.

+1 for removing it and relying on GitHub's automatic contribution tracking instead.

adam-urbanczyk:
+1

jmwright:
@RubenRubens Thanks!

Issue #650, opened 2021-02-20
debugging OCCT errors


mbway:
very often I'm finding that operations that work for some shapes fail when called on other shapes, so to get the result I want I'm trying to create the same shape or similar shapes in different ways until the command works. This is extremely frustrating and time consuming since the only feedback is `OCP.StdFail.StdFail_NotDone: BRep_API: command not done`.

Is there any way to obtain more information about what caused the failure, and perhaps since cadquery also controls the native bindings, perhaps there could be a way to trace what calls are made to OCCT and with which data so that problems can be examined in isolation (although if the API is stateful like cadquery then perhaps this wouldn't be useful)?

adam-urbanczyk:
`cq.Shape.exportBrep` gets you the internal (AFAIK lossless)  format of OCCT. You can use it in `draw.exe` to e.g. submit bugs on the OCCT tracker. We might want to implement `cq.Shape.importBrep`, but for now you can use OCP calls to get there (`OCP.BRepTools.BRepTools.Read_s`).

NB: CQ-editor only displays part of the stack trace, you can run the code in the embedded console or just in (i)python to get the full version. This will tell you which high-level OCCT operation is failing, but that is not always very helpful.

If you find a good way to debug OCCT issues do let us now.

fedorkotov:
@adam-urbanczyk Exporting BREP and stack trace are good but optional logging of all OCCT API calls made by CadQuery would be a even better. Ideally C++ program demonstrating an issue should be easy to reconstruct from such a log. This would help in reporting bugs with relevant examples to OCCT developers. I suspect this would also help you and other CadQuery developers in debugging efforts.

dcowden:
I agree with the idea of a trace. It would be super-helpful because we could ask for that as a standard bug report requirement. That's a great idea @fedorkotov

adam-urbanczyk:
I'm not sure that I understand how would a log of OCCT calls help in debugging when you have a complete python code you can debug. Or are you proposing tracing through the C++ code?

You can easily construct a MWE by dumping the breps at the offending OCCT call and then recreating them via draw. AFAIK this is the preferred format for a OCCT bug reports.

dcowden:
@adam-urbanczyk whats an MWE?

[...]

Issue #281, opened 2020-02-16
Investigate using VTK for web based visualization


adam-urbanczyk:
Current approach is based on x3dom functionality of python-occ. It will probably be not available in the future but it would be good to keep the simple ipython visuaization capabilities. One way to achieve this woulde be to use VTK and some additional tools. Relevant materials

* https://lorensen.github.io/VTKExamples/site/Python/DataManipulation/Cube.py
* https://k3d-jupyter.org/index.html
* https://docs.pyvista.org/index.html

Note that OCCT has (in some versions) VTK related functionalities:

* https://www.opencascade.com/doc/occt-7.0.0/refman/html/toolkit_tkivtk.html --> has it been removed in 7.4 ?

bernhard-42:
I agree searching for an alternative to x3dom is a good idea.
I had severe problems with x3dom when I built my first jupyter extension for cadquery. The biggest one was that with x3dom I regularly ran out of Webgl contexts, and I did not find a way to control the building of webgl contexts. And there is a lack of flexibility from a developer perspective.

That's why I switched over to pythreejs (and pythonocc-core also comes with a pythreejs based jupyter renderer). The overall experience with pythreejs is good. From a documentation perspective, I very often needed to rely on the three.js docs, but since pythreejs is a thin layer on top of threejs, most of the stuff works as expected.

adam-urbanczyk:
Some notes:

getting `vtkPolyData`:
```python
vs = OCP.IVtkOCC.IVtkOCC_Shape(w)
sd = OCP.IVtkVTK.IVtkVTK_ShapeData()
sm = OCP.IVtkOCC.IVtkOCC_ShapeMesher()
sm.Build(vs,sd)

res = sd.getVtkPolyData()
```

So far nice simple visualziation with pyvista+itkwidgets. VTK.js being the lower-level lib
```python
mesh = pv.PolyData(res)
pv.plot_itk(mesh, smooth_shading=True)
```
![image](https://user-images.githubusercontent.com/13981538/114216270-fb7e3b80-9966-11eb-9166-deae6640b74a.png)

adam-urbanczyk:
MWE for future reference:
```python
import OCP
import vtk
import pyvista as pv

shape  = cq.Workplane().box(1,1,1).faces('>Z').hole(0.7).edges().fillet(0.03)

vs = OCP.IVtkOCC.IVtkOCC_Shape( shape.val().wrapped )
[...]

PR #735, opened 2021-04-20
Implement importBrep and vtkPolyData export

adam-urbanczyk:
Related to #650 and #281

- [x] Write to brep
- [x] Stream support
- [x] toVtkPolyData
- [x] _repr_javascript_
- [x] disable keyboard interaction in notebooks
- [x] vtk from assy
- [x] cq_directive based on vtk.js
- [x] tests

### Highlights

Rendering in jupyter notebooks:
image


bernhard-42:
I know it's early times for this implementation, but here are some comments on the parameters and defaults of `toVtkPolyData`: 

Possible changes for the defaults:

- angularTolerance 
  [Opencascade recommends](https://dev.opencascade.org/doc/occt-7.1.0/overview/html/occt_user_guides__modeling_algos.html#occt_modalg_11_2) for the angular deflection a degree angle of 12-20°, which is 0.21-0.35 rad ("_Angular deflection is relatively simple and allows using a default value (12-20 degrees)_"). Since going down to 0.1 (i.e. 5.7°) comes with a significant time penalty (see below) and people might just use the default, one could consider to set the default to 0.2 or even 0.3

- tolerance
  From my understanding, the shape mesher determines the linear deflection according to the shape size and uses the tolerance arg to refine it (see examples below). Since it is not that obvious for users how to select it (and the timings are not so critical) maybe setting a default like 0.001 would be helpful?

Possible additional parameters:

- triangulate: bool, default=False
  if True, apply `vtkTriangleFilter` on the poly data (for us pythreejs users :-) )

- isolines: int, default=0
  Would it be an idea to expose one parameter `isolines` for `u` an `v` so that a user could select wether to get Iso lines (`isolines>0`) or not (`isolines=0`)?

- parallel: bool, default=False
  For larger objects having a possibility to leverage multicore CPUs would be nice.

Personally I found the tolerance=0.001 and angularTolerance of 0.3 was a great compromise of visual quality and tessellation duration.

Some timings:

The large RC example

image

A small example:

image

with

```python
def compute(shape, deviation=0.0001, angular_deviation=0.3, iso_lines=0, info=False):
    vs = OCP.IVtkOCC.IVtkOCC_Shape(shape)
    sd = OCP.IVtkVTK.IVtkVTK_ShapeData()
    sm = OCP.IVtkOCC.IVtkOCC_ShapeMesher(
        theDevCoeff=deviation, 
        theDevAngle=angular_deviation,
        theNbUIsos=iso_lines, 
        theNbVIsos=iso_lines)
    sm.Build(vs,sd)
    return pv.PolyData(sd.getVtkPolyData()), sm
```

adam-urbanczyk:
Will look into this, but the defaults are based on current defaults for other ops. I don't want to change them just because. I also propose to skip the isolines completely.

marcus7070:
554cda6 works in both Jupyter Lab and Notebook for me, nice work @adam-urbanczyk!

cadquery/cq_directive.py
```diff
@@ -84,9 +114,95 @@ def run(self):
...
 
     app.add_directive("cq_plot", cq_directive)
+    app.add_directive("cq_vtk", cq_directive_vtk)
+
+    # download and add vtk.js
+    build_path = Path(app.outdir)
+    out_path = build_path / "_static"
+    out_path.mkdir(parents=True, exist_ok=True)
+
+    request.urlretrieve("https://unpkg.com/vtk.js", out_path / "vtk.js")
```
marcus7070, line 206: Is there any alternative to this? The Nix package manager does not allow network access during build time (because it produces deterministic builds). I'm pretty sure this is the first bit of network access (outside of package managers) in the CQ code base.

I can download the file before building and patch this on the Nix side if there is no reasonable alternative.

adam-urbanczyk: I can bundle the file. BTW: Shouldn't `cq_directive` be moved completely to`doc`?

marcus7070: > I can bundle the file.

Thanks!

> Shouldn't cq_directive be moved completely todoc?

I agree, the Sphinx docs seem to suggest that structure.

adam-urbanczyk:
There are still some js glitches (I'll ask on the vtk.js forum), but I think that you can start reviewing.

marcus7070:
@adam-urbanczyk, sorry to annoy you with unsupported installation questions, but when I build this through Nix I get a single failed test for:
```
> =================================== FAILURES ===================================
> ______________________ TestCadQuery.testBrepImportExport _______________________
> [gw6] linux -- Python 3.8.9 /nix/store/q6gfck5czr67090pwm53xrdyhpg6bx67-python3-3.8.9/bin/python3.8
> self = <tests.test_cadquery.TestCadQuery testMethod=testBrepImportExport>
>     def testBrepImportExport(self):
>         # import/export to file
>         s = Workplane().box(1, 1, 1).val()
>         s.exportBrep("test.brep")
>         si = Shape.importBrep("test.brep")
>         self.assertTrue(si.isValid())
>         self.assertAlmostEqual(si.Volume(), 1)
>         # import/export to BytesIO
>         from io import BytesIO
>         bio = BytesIO()
> >       s.exportBrep(bio)
> tests/test_cadquery.py:4561:
> _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
> self = <cadquery.occ_impl.shapes.Solid object at 0x7fff8e2b7fa0>
> f = <_io.BytesIO object at 0x7fff8e5b7c20>
>     def exportBrep(self, f: Union[str, BytesIO]) -> bool:
>         """
>         Export this shape to a BREP file
>         """
> >       rv = BRepTools.Write_s(self.wrapped, f)
> E       TypeError: Write_s(): incompatible function arguments. The following argument types are supported:
> E           1. (Sh: OCP.TopoDS.TopoDS_Shape, File: str, theProgress: OCP.Message.Message_ProgressRange = <OCP.Message.Message_ProgressRange object at 0x7fffe19a5bf0>) -> bool
> E       
> E       Invoked with: <OCP.TopoDS.TopoDS_Solid object at 0x7fff8e2d5c70>, <_io.BytesIO object at 0x7fff8e5b7c20>
> cadquery/occ_impl/shapes.py:468: TypeError
```

I think you've solved the same thing in https://github.com/CadQuery/OCP/pull/47; did you make some extra changes outside of the OCP repo that I've missed?

I'm using the latest https://github.com/CadQuery/OCP/commit/b058865f1f041941564637addf47c547119d8284, https://github.com/CadQuery/pywrap/commit/f8869e5a47fd3e3e1d31e7ab89b46c158f7487bf and pybind11 v2.6.1. My versions for clang, gcc, jinja2, pyparsing and everything else I can think of all match the logs from the latest OCP CI build.

cadquery/occ_impl/exporters/__init__.py
```diff
@@ -26,9 +27,10 @@ class ExportTypes:
     TJS = "TJS"
     DXF = "DXF"
     VRML = "VRML"
+    VTP = "VTP"
 
 
-ExportLiterals = Literal["STL", "STEP", "AMF", "SVG", "TJS", "DXF", "VRML"]
+ExportLiterals = Literal["STL", "STEP", "AMF", "SVG", "TJS", "DXF", "VRML", "VTP"]
```
jmwright, line 33: The import/export doc will need to be updated for the inclusion of VTP at some point.

cadquery/occ_impl/exporters/assembly.py
```diff
@@ -61,3 +72,52 @@ def exportCAF(assy: AssemblyProtocol, path: str) -> bool:
...
+    with TemporaryDirectory() as tmpdir:
+
+        exporter = vtkJSONSceneExporter()
+        exporter.SetFileName(tmpdir)
+        exporter.SetRenderWindow(renderWindow)
+        exporter.Write()
+        make_archive(path, "zip", tmpdir)
+
+
+def exportVRML(assy: AssemblyProtocol, path: str):
```
jmwright, line 115: Thanks, I will give this a try for the KiCAD generator soon.

cadquery/occ_impl/shapes.py
```diff
@@ -1072,14 +1101,51 @@ def tessellate(
 
         return vertices, triangles
 
-    def _repr_html_(self):
+    def toVtkPolyData(
+        self, tolerance: float, angularTolerance: float = 0.1, normals: bool = True
```
jmwright, line 1105: No default for the tolerance here?

jmwright (approved):
@adam-urbanczyk Thanks for all the work you've put into this!

marcus7070:
This is not completely relevant to this PR, but:

> when I build this through Nix I get a single failed test for:
> 
> ```
> > =================================== FAILURES ===================================
> > ______________________ TestCadQuery.testBrepImportExport _______________________

In my OCP build process I was accidentally giving pywrap the includes from libc++ instead of libstdc++, so when pywrap generated the mangled name for `OCP.BRepTools.BRepTools.Write_s(... iostream ...)` it had a different name to what was in `symbols_mangled_linux.dat` and was being dropped. I corrected my pywrap includes and I now have the correct overloaded methods for `BRepTools.Write_s`.

doc/assy.rst
```diff
@@ -205,7 +205,8 @@ Final result
 
 Below is the complete code including the final solve step.
 
-.. code-block:: python
+.. cq_vtk::
```
marcus7070, line 208: Is there any reason to use cq_vtk over the plain cadquery directive used everywhere else?

marcus7070 (approved):
I would like to have the default camera position as iso rather than [top-down](https://cadquery--735.org.readthedocs.build/en/735/assy.html#final-result), but that can be another PR.

I think we can remove sphinxcadquery from environment.yml? There are also two more lines for sphinxcadquery in conf.py that could also be removed.

The build_docs.sh in the root dir is incorrect now, docs need to be built from the doc dir so the relative link to vslot_2020_1.dxf works.

Thanks so much for all this work @adam-urbanczyk!

cadquery/cq.py
```diff
@@ -4022,15 +4022,17 @@ def offset2D(
 
         return self.newObject(rv)
 
-    def _repr_html_(self) -> Any:
+    def _repr_javascript_(self) -> Any:
```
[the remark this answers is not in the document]
adam-urbanczyk: No clue to be honest. This is the magic method to be used when returning a js based object representation - there is not much I can do about it.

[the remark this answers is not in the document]
adam-urbanczyk: @bernhard-42 would you know?

cadquery/occ_impl/assembly.py
```diff
@@ -155,3 +167,71 @@ def toCAF(
...
+        mapper.SetInputData(data)
+
+        actor = vtkActor()
+        actor.SetMapper(mapper)
+        actor.SetPosition(*trans)
+        actor.SetOrientation(*rot)
+        actor.GetProperty().SetColor(*color[:3])
+        actor.GetProperty().SetOpacity(color[3])
+
+        renderer.AddActor(actor)
```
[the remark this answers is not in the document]
adam-urbanczyk: I'm recursively adding actors to the vtkRenderer instance (and taking into account their global position)

cadquery/occ_impl/exporters/assembly.py
adam-urbanczyk: HTH

cadquery/occ_impl/shapes.py
adam-urbanczyk: Good point - I'll add 1e-3

cadquery/cq.py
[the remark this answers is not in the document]
bernhard-42: @jmwright @adam-urbanczyk Thanks for surfacing that. Jupyter Cadquery monkey patches CadQuery by deleting the  `_repr_html_` methods of `Workplane` and `Shape` and adding its own renderers via `_ipython_display_`
So no problem, for my next version that will support the new version of CadQuery I will also remove the `_repr_javascript_` methods.

cadquery/occ_impl/shapes.py
adam-urbanczyk: Actually I just followed the conventions of mesh and tessellate. I think the rational is - this value is use case dependent and it is a low level API assuming that a potential user will be able to judge himself which tol is needed.

cadquery/occ_impl/exporters/__init__.py
adam-urbanczyk: Adding a note in the docs

cadquery/occ_impl/shapes.py
```diff
@@ -15,6 +15,15 @@
 )
 from typing_extensions import Literal, Protocol
 
+from io import BytesIO
+
+from vtk import (
+    vtkPolyData,
+    vtkTriangleFilter,
+    vtkTriangleMeshPointNormals,
```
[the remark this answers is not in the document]
adam-urbanczyk: Done, thanks

cadquery/occ_impl/shapes.py
```diff
@@ -448,12 +460,29 @@ def exportStep(self, fileName: str) -> IFSelect_ReturnStatus:
...
-    def exportBrep(self, fileName: str) -> bool:
+    def exportBrep(self, f: Union[str, BytesIO]) -> bool:
         """
         Export this shape to a BREP file
         """
 
-        return BRepTools.Write_s(self.wrapped, fileName)
+        rv = BRepTools.Write_s(self.wrapped, f)
+
+        return True if rv is None else rv
```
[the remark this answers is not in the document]
adam-urbanczyk: Actually I just followed the existing convention. Due to the OCP overload `rv` None or True means no error, False means an error. Do you insist that this needs to be addressed in this PR?

cadquery/occ_impl/exporters/assembly.py
```diff
@@ -61,3 +72,52 @@ def exportCAF(assy: AssemblyProtocol, path: str) -> bool:
...
+    Export an assembly to a zipped vtkjs. NB: .zip extensions is added to path.
+    """
+
+    renderer = vtkRenderer()
+    renderWindow = vtkRenderWindow()
+    renderWindow.AddRenderer(renderer)
+    toVTK(assy, renderer)
+
+    renderer.ResetCamera()
+    renderer.SetBackground(1, 1, 1)
```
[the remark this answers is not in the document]
adam-urbanczyk: Thanks!

cadquery/occ_impl/exporters/vtk.py
```diff
@@ -0,0 +1,24 @@
+from vtk import vtkXMLPolyDataWriter
+from ..shapes import Shape
+
+
+def exportVTP(
+    shape: Shape, fname: str, tolerance: float = 0.1, angularTolerance: float = 0.1
```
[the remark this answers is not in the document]
adam-urbanczyk: `toString` and `export` use 0.1, so I'd say yes.

doc/assy.rst
adam-urbanczyk: Good catch, fixed.

cadquery/occ_impl/assembly.py
```diff
@@ -155,3 +167,71 @@ def toCAF(
...
+    if assy.shapes:
+        data = Compound.makeCompound(assy.shapes).toVtkPolyData(tolerance)
+
+        mapper = vtkMapper()
+        mapper.SetInputData(data)
+
+        actor = vtkActor()
+        actor.SetMapper(mapper)
+        actor.SetPosition(*trans)
+        actor.SetOrientation(*rot)
```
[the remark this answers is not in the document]
adam-urbanczyk: That's how I'd interpret it, but when I tried `gp_Extrinsic_ZXY` it did not work and it does seem to work with `gp_Extrinsic_XYZ`:
image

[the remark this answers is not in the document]
marcus7070: Great, thanks for testing it, I only looked at the docs.

cadquery/occ_impl/shapes.py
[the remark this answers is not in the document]
marcus7070: No need. I didn't realise that this is not propagated to the fluent api anywhere, the only way to encounter it is in occ_impl, so the OCP behaviour is more appropriate.

adam-urbanczyk:
Alright, thanks for all the comments. I propose to wait 24h with this and then squash/merge.

marcus7070:
No last second changes @adam-urbanczyk? I'll squash and merge in a bit then.

marcus7070:
That was a huge effort, thank you very much @adam-urbanczyk!

Jojain:
I had not time to ask before merging but :
What this PR is for and what does it allow to do ?
To my knowledge toVTKPolyData allows to convert a cadquery model in data readable by VTK ? So this allows to render cq models in software based on VTK like Paraview ? Or am I  getting it all wrong ?

Thank you in advance.

adam-urbanczyk:
Yes, it allows convert to vtk data structures and save as vtp or vtkjs. Additionally this PR implemented rendering in jupyter and our docs based on vtk.js.

adam-urbanczyk:
Also (de)serialization of cq.Shape objects to BytesIO - should be handy for caching

PR #737, opened 2021-04-20
Improve docs of boolean operations

RubenRubens:
Display `__or__` example as a block of code.


jmwright (approved):
Thanks again @RubenRubens !

Issue #751, opened 2021-04-29
Splitting with arbitrary plane


fedorkotov:
What is the proper way of splitting geometry with an arbitrary plane?

### Is `Workplane.split(..)` bugged or am I using it wrong?
[Documentation suggests](https://cadquery.readthedocs.io/en/latest/classreference.html?highlight=split#cadquery.Workplane.split) that `Workplane.split(..)` should accept `Shape` or `Workplane` 

Even though `Plane` is not a `Shape` split accepts it without error but splits the model along current XY plane instead,
```python
def get_body():
    return cq.Workplane("XY")\
    .rect(1,1)\
    .extrude(1)

result = \
    get_body()\
    .split(
        cq.Plane.named("ZX")
        .rotated((0,-30,0)))
```
![image](https://user-images.githubusercontent.com/1106650/116615906-a3729d80-a944-11eb-8a1c-32c285466b19.png)
(should produce wedge shape, see below)

If I create a `Workplane` from my plane and pass it to `split` it produces `ValueError: Null TopoDS_Shape object` exception
```python
def get_body():
    return cq.Workplane("XY")\
    .rect(1,1)\
    .extrude(1)

cutting_workplane = \
    cq.Workplane(
        inPlane = \
            cq.Plane.named("ZX")
            .rotated((0,-30,0)),
        origin = cq.Vector(0,0,0))

result = \
    get_body()\
    .split(cutting_workplane)
```

### Can I patch `Workplane.plane`?

The following code works and produces expected result but looks ugly. 
Is this Ok thing to do? Is `Workplane.plane` an implementation detail or part of "public" interface of `Workplane`?

```python
def splitTop(wp, plane):
    old_plane = wp.plane
    wp.plane = plane
    wp = wp.split(keepTop=True)
    wp.plane = old_plane
    return wp

def get_body():
    return cq.Workplane("XY")\
    .rect(1,1)\
    .extrude(1)
    
result = splitTop(
    get_body(), 
    cq.Plane.named("ZX")\
    .rotated((0,-30,0)))
```
![image](https://user-images.githubusercontent.com/1106650/116617170-38c26180-a946-11eb-954e-0aff03ee5053.png)

jmwright:
The following works for me.
```python
result = (cq.Workplane().box(10, 10, 10)
                        .faces(">Z").transformed(rotate=(45, 45, 45))
                        .split(keepBottom=True))
```
[...]

Issue #464, opened 2020-09-29
Allow creating named workplanes (like 'XY') through the fluent API


marcus7070:
Often it's easiest to create a workplane with one of the named orientations (like 'XY') and an offset in world coords. The `Workplane.workplane` method can't directly do this. I find myself continually abusing the `Workplane.copyWorkplane` method to create workplanes like so:
```python
part = part.copyWorkplane(cq.Workplane("YZ", origin=(dim_x, dim_y, dim_z)))
```
and I see I'm not the only one who's picked up this habit: https://github.com/CadQuery/cadquery/issues/357#issuecomment-670413077

Would you be interested in a PR that refactored the workplane generating code from `Workplane.__init__` into a new method on the Workplane class like:
```python
def namedWorkplane(inplane='XY', origin=(0, 0, 0)):
```
`Workplane.__init__` would keep the same arguments and just pass them on to `self.namedWorkplane` to set up the first plane.

Now my example above would become:
```python
part = part.namedWorkplane('YZ', origin=(dim_x, dim_y, dim_z))
```

Just as a side note, IMHO the `Workplane.workplane` method is focussed around making workplanes relative to a face. I like it how it is and I don't think it should be overloaded/overcomplicated to do this job as well. 

I'm happy to do this PR, just thought I should get some opinions before I go refactor the code, write tests and docs.

marcus7070:
I realised there is another option. `Workplane.copyWorkplane` could be changed to accept either an `occ_impl.Plane` or `Workplane` object. Then the above example would become
```python
part = part.copyWorkplane(cq.Plane.named('XY', (dim_x, dim_y, dim_z)))
```

Not too sure which version I like better, I would love to hear some opinions!

jmwright:
I guess I'd lean towards the `namedWorkplane` solution, but I rarely feel the need to use `copyWorkplane`, so I may be a skewed sample. I'd be interested in what @adam-urbanczyk thinks about this.

adam-urbanczyk:
Actually I'd vote for extending `workplane` to support this use case (`.workplane('XY')` ) and the one described in #357 (`.workplane((1,1,1)`).  I find this kind of API much cleaner. And there is still #379 waiting to be investigated.

marcus7070:
I'm agreeing with @adam-urbanczyk now, this is a great candidate for multimethods (#379).

jmwright:
It sounds like it's safe to start the PR then @marcus7070

[...]

PR #752, opened 2021-05-01
rework Workplane.split

marcus7070:
Now also accepts Plane as an argument. Workplane split now goes through `Solid.split` instead of `Solid.cut`.

Still to do:
- [ ] tests
- [ ] docs

Related to #751.

### DIFF
--- a/cadquery/cq.py
+++ b/cadquery/cq.py
@@ -264,6 +264,12 @@ class Workplane(object):
 
         return list(all.values())
 
+    def _split(self: T, s: Union[Shape, List[Shape]]):
+
+        solid = self.findSolid()
+        tools = (s,) if isinstance(s, Shape) else s
+        return solid.split(*tools)
+
     @overload
     def split(self: T, keepTop: bool = False, keepBottom: bool = False) -> T:
         ...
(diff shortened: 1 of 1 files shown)


marcus7070:
I think I'll work on https://github.com/CadQuery/cadquery/issues/464#issuecomment-700934881 before I continue this PR. I think it makes more sense to expand the functionality of `Workplane.workplane`. Then we can keep the current method of splitting on a workplane instead of introducing a new method to do almost the same thing (splitting on a Plane).

If there is any use case that is not covered with an improved `Workplane.workplane` then we can have a look at `.split(Plane(...))`.

fedorkotov:
>  If there is any use case that is not covered with an improved Workplane.workplane then we can have a look at .split(Plane(...)).

Such a use case is splitting with a plane specified in world coordinate system (not related to any face or other feature of the model).

adam-urbanczyk:
AFAIR the point of #464 would be to allow that too (e.g. specify origin).

fedorkotov:
@adam-urbanczyk  I thought origin was specified in local coordinate system of current face or other object because of this passage from #464 description
> Just as a side note, IMHO the Workplane.workplane method is focussed around making workplanes relative to a face. I like it how it is and I don't think it should be overloaded/overcomplicated to do this job as well.

 If it is specified in world coordinate system then indeed my case is covered too.

lalebarde:
How to use it please ?

adam-urbanczyk:
#751 is closed, closing this PR.

Issue #788, opened 2021-06-09
Matrix Class is missing the __repr__() special method and therefore can't be displayed


gumyr:
Please consider adding the `__repr__` special method to the Matrix Class.  Something like this:

```python
    def __repr__(self):
        result = ''
        for r in range(4):
            for c in range(4):
                result += str(self.__getitem__((r,c)))
                if c!=3:
                    result += ', '
            result += '\n'
        return result
```
Being able to display the matrix helps with debugging (for example, trying to determine if the rotate methods use radians or degrees).

gumyr:
I'll give it a try

gumyr:
Okay, I have the code (with unittest ready) in a local git branch but I'm not sure of the next steps.  To do this properly do I fork master, branch, update code, push it back to my fork then make a pull request? I only know enough to be dangerous at this point ;)

PR #791, opened 2021-06-16
#788 Add __repr__ and __str__ methods to Matrix

gumyr:
Addressing issue #788


fedorkotov:
I guess for occasional debug output what you did may be good enough. But assembling a long string by consecutive concatenation of large number of small pieces (`s +="something"`) is not very efficient.

See https://waymoot.org/home/python_string/ for example. The article is more than 15 years old but still of interest today.  You may want to re-run tests mentioned there on a modern python.

gumyr:
Thanks for the link, that was interesting. Given that the matrix can only contain 16 values and these methods will likely only be used for debugging I went with simple and easy to understand.

jmwright:
One of the tests had thrown an error that was unrelated to this PR, so I reran it and it passed.

cadquery/occ_impl/geom.py
```diff
@@ -361,6 +361,35 @@ def __getitem__(self, rc: Tuple[int, int]) -> float:
         else:
             raise IndexError("Out of bounds access into 4x4 matrix: {!r}".format(rc))
 
+    def __repr__(self) -> str:
```
adam-urbanczyk, line 364: How about suing f strings and reusing the transposed? Something along those lines:
```python
l = cq.Matrix().transposed_list()
rv=(f'Matrix({l[0::4]},\n'
    f'       {l[1::4]},\n'
    f'       {l[2::4]},\n'
    f'       {l[3::4]})')
```

adam-urbanczyk (changes requested):

cadquery/occ_impl/geom.py
```diff
@@ -361,6 +361,35 @@ def __getitem__(self, rc: Tuple[int, int]) -> float:
...
+                result += str(self.__getitem__((r,c)))
+                if c!=3:
+                    result += ", "
+                elif r!=3:
+                    result += "],\n"
+                else:
+                    result += "]])"
+        return result
+
+    def __str__(self) -> str:
```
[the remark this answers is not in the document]
gumyr: The __repr__ method is supposed to produce a valid Python expression while __str__ is just a readable string. In the unit test (test_cad_objects.py) I do a "str(eval(repr(m)))" which proves out the __repr__ method is this valid python expression. Is this important in practice? I personally don't see a use case for evaluating a Matrix but I guess it's good practice. 

The Matrix __str__ method could produce the same result but I followed the vector formatting: "return "Vector: " + str((self.x, self.y, self.z))". Note that the Vector __repr__ does not produce a valid Python expression.

cadquery/occ_impl/geom.py
gumyr: The transposed_list() method does generate a list but it also transposes the matrix which would have to be undone. To address the comments I've opened another pull request with this code: 
        list = [ [self.__getitem__((r,c)) for c in range(4)] for r in range(4)]
        return "Matrix("+"".join(str(list)).replace("],","],\n       ")+")"
which generates a list much like transposed_list() does and then uses join and some string replacement to generate the same strings as before. The code is more compact but I'll let you decide if it's better.

I also improved the unit test as I realized that using the identity matrix hides accidental transposition of the matrix.

cadquery/occ_impl/geom.py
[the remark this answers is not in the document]
adam-urbanczyk: I don't think that what you state is universally true - try `repr(object())`, it does not result in a python expression at all. I'd say indicate the type in __repr__ and return a list of lists in __str__

cadquery/occ_impl/geom.py
adam-urbanczyk: Does [0::4] not result in selecting a column (a thus a reverse transpose when printing it in every row)?

gumyr: My comment about the use of the transposed_list() method applied to my solution where I do a str(list) which would not compensate for a transposed matrix. I stand behind my pull request for this trivial change.

cadquery/occ_impl/geom.py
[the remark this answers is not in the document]
gumyr: My definitions of __repr__ and __str__ are based on the wisdom of stackoverflow: https://stackoverflow.com/questions/3691101/what-is-the-purpose-of-str-and-repr
I agree that __repr__ doesn't always produce a valid Python expression, that's why I was describing it as good practice. The code I've submitted does produce a valid Python expression and is only two lines long. The __str__ method is almost exactly the same, I just followed the precedence of the Vector __str__ method.

cadquery/occ_impl/geom.py
```diff
@@ -361,6 +361,19 @@ def __getitem__(self, rc: Tuple[int, int]) -> float:
         else:
             raise IndexError("Out of bounds access into 4x4 matrix: {!r}".format(rc))
 
+    def __repr__(self) -> str:
+        """
+        Generate a valid python expression representing this Matrix
+        """
+        list = [ [self.__getitem__((r,c)) for c in range(4)] for r in range(4)]
+        return "Matrix("+"".join(str(list)).replace("],","],\n       ")+")"
```
fedorkotov, line 369: 1. Reusing built in `list` for your own variable name is not a good idea IMHO.
1. Your `list` is a list of lists of numbers. You apply `str(..)` to it. It becomes one string. Then you apply `"".join(..)` to this string. That does not change the string in any way.
1. Then you make string replacements. Why make replacements instead of formatting everything properly in the first place?

Maybe something like this instead (or whatever the Black formater transformes it into)?
```python
matrix_arg = \
  ",\n       ".join(
    (str([self.__getitem__((r,c)) for c in range(4)]) \
     for r in  range(4)))
return f"Matrix({matrix_arg})"   
```

gumyr: Thank you for the feedback. I agree with all of your comments and have made another pull request with the appropriate changes.
Cheers,
Roger

cadquery/occ_impl/geom.py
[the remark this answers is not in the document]
gumyr: Adam, do you still have concerns about the proposed change? The __repr__ method does indicate the type (and just happens to be valid expression) while the __str__ method generates 'Matrix: ' and a bracketed  list of lists.

cadquery/occ_impl/geom.py
adam-urbanczyk: OK enough bike-shedding, just please make it more readable (i.e. get rid of `__getitem__`)  and fix the black failure

cadquery/occ_impl/geom.py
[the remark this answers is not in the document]
adam-urbanczyk: All things considered I'd get rid of `str`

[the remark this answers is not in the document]
gumyr: Would you like me to get rid of the `__str__` method of the Vector class as well?

cadquery/occ_impl/geom.py
```diff
@@ -366,15 +366,15 @@ def __repr__(self) -> str:
         """
         Generate a valid python expression representing this Matrix
         """
-        list = [ [self.__getitem__((r,c)) for c in range(4)] for r in range(4)]
-        return "Matrix("+"".join(str(list)).replace("],","],\n       ")+")"
+        matrix_str = ",\n        ".join(str(self.transposed_list()[i::4]) for i in range(4))
```
fedorkotov, line 369: You are calling transposed_list() in 4 times. Each time you throw away most of the result it returns.
I think it should be called once and result should be stored in temporary variable for later use.
Same with __str__(..) below

cadquery/occ_impl/geom.py
gumyr: I changed out `__getitem__` for `transposed_list`.  All of the checks have passed with this latest revision and there are no conflicts.

cadquery/occ_impl/geom.py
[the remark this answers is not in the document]
gumyr: Adam, I removed the `__str__` method from Matrix. Let me know if you'd like the same done for Vector.

cadquery/occ_impl/geom.py
gumyr: I went to far in an attempt for clarity .. the code only calls `transpose_list` once now. By the way, Adam didn't want the `__str__` method so I removed it. Thanks for your suggestions!

gumyr:
I don't know enough about the build system to debug this.  AppVeyor is reformatting the unit test code (changing white space) and generates the following message:

     def testEdgeWrapperRadius(self):
 
         # get a radius from a simple circle
         e0 = Edge.makeCircle(2.4)
would reformat tests/test_cad_objects.py
Oh no! 💥 💔 💥
2 files would be reformatted, 65 files would be left unchanged.
Command exited with code 1
Build failed

This unittest runs fine in my environment but I don't have AppVeyor.

fedorkotov:
This is [Black Formatter's](https://github.com/psf/black) error message. Install the same version of it that is used in CI/CD scripts (19.10b0 currently, see environment.yml) and format your code with it before committing. I personally don't like Black but using it [is the law](https://github.com/CadQuery/cadquery#contributing-code) here and in many other python projects

gumyr:
Thanks for the help - I didn't know about black but now that I've run it all the checks have passed.  The coverage of geom.py is now at 99.58%.

adam-urbanczyk (approved):

adam-urbanczyk:
@jmwright are you OK with merging this?

jmwright (approved):
Thanks @gumyr

jmwright:
+1 on merging

adam-urbanczyk:
Thanks for the contribution @gumyr !

PR #805, opened 2021-06-27
doc/importexport: fix heading underlines

### LINKED ISSUE #799: Headings in Import/Export of docs incorrect
screenshot2021-06-26-133327

marcus7070:
Will close #799.

This is trivial, I don't think it needs review (but you're welcome to if you have time). I'll merge it tomorrow.


jmwright:
There might be other places in the docs where the headers are broken since I'm pretty sure I just copied this header style from somewhere else.

PR #813, opened 2021-07-05
text() docstring update

### LINKED ISSUE #463: text() fontsize parameter dimensions unclear
The documentation for the `fontsize` parameter of the `text` function says "size of the font" but it's unclear how that size is defined. While the value is dimensionless, what does it refer to? The size of a particular character? Something else? I'm happy to open up a PR to improve the documentation. but I'd need a little clarification before doing so.

lorenzncode:
* Remove following line from docstring (appears to be copy/paste error from extrude())

       extrude always *adds* material to a part.

    * Swap order of :param distance: and :param fontsize: to match argument order

    * Mention fontsize is in model units (fix #463)

    * distance: mention distance is for both extrude/cut

    * Update kind default in docstring as "regular" not "Normal"

    * Add missing fontPath arg description

    * Add examples of usage

      - Using an installed system font

      - Using a font at provided path (when font not installed on system)

        I tested with the free fonts from http://www.gust.org.pl/projects/e-foundry/tex-gyre

      - Cut text into solid

    * Change docstring to fix autodoc-typehints issue where the Parameters section is duplicated (see screenshot)

  If this fix looks good, I can follow up with a similar docstring fix for the duplicate Parameters issue on other methods.

current

docs_text_current_duplicate_parameters

update:

doc_text_update

### DIFF
--- a/cadquery/cq.py
+++ b/cadquery/cq.py
@@ -3930,7 +3930,7 @@ class Workplane(object):
         :param fontPath: path to font file
         :param kind: font type (default: regular)
         :param halign: horizontal alignment
-        :param valign: vertical alignment (default: center)
+        :param valign: vertical alignment
         :return: a CQ object with the resulting solid selected
 
         The returned object is always a CQ object, and depends on whether combine is True, and
--- a/cadquery/cq.py
+++ b/cadquery/cq.py
@@ -3946,7 +3946,7 @@ class Workplane(object):
 
         Specify the font (name), and kind to use an installed system font::
 
-            cq.Workplane().text("CadQuery",5,1,font="Liberation Sans Narrow", kind="italic")
+            cq.Workplane().text("CadQuery", 5, 1, font="Liberation Sans Narrow", kind="italic")
 
         Specify fontPath to use a font from a given file::
 
--- a/cadquery/cq.py
+++ b/cadquery/cq.py
@@ -3928,7 +3928,7 @@ class Workplane(object):
         :param clean: call :py:meth:`clean` afterwards to have a clean shape
         :param font: font name
         :param fontPath: path to font file
-        :param kind: font type (default: regular)
+        :param kind: font type
         :param halign: horizontal alignment
         :param valign: vertical alignment
         :return: a CQ object with the resulting solid selected
--- a/cadquery/cq.py
+++ b/cadquery/cq.py
@@ -3942,7 +3942,7 @@ class Workplane(object):
 
         Examples::
 
-            cq.Workplane().text("CadQuery",5,1)
+            cq.Workplane().text("CadQuery", 5, 1)
 
         Specify the font (name), and kind to use an installed system font::
 
--- a/cadquery/cq.py
+++ b/cadquery/cq.py
@@ -3954,7 +3954,7 @@ class Workplane(object):
 
         Cutting text into a solid::
 
-            cq.Workplane().box(8,8,8).faces(">Z").workplane().text("Z", 5,-1.0)
+            cq.Workplane().box(8, 8, 8).faces(">Z").workplane().text("Z", 5, -1.0)
 
         """
         r = Compound.makeText(
--- a/cadquery/cq.py
+++ b/cadquery/cq.py
@@ -3929,7 +3929,7 @@ class Workplane(object):
         :param font: font name
         :param fontPath: path to font file
         :param kind: font type (default: regular)
-        :param halign: horizontal alignment (default: center)
+        :param halign: horizontal alignment
         :param valign: vertical alignment (default: center)
         :return: a CQ object with the resulting solid selected
 
--- a/cadquery/cq.py
+++ b/cadquery/cq.py
@@ -3950,7 +3950,7 @@ class Workplane(object):
 
         Specify fontPath to use a font from a given file::
 
-            cq.Workplane().text("CadQuery",5,1, fontPath="/opt/fonts/texgyrecursor-bold.otf")
+            cq.Workplane().text("CadQuery", 5, 1, fontPath="/opt/fonts/texgyrecursor-bold.otf")
 
         Cutting text into a solid::
 
--- a/cadquery/cq.py
+++ b/cadquery/cq.py
@@ -3933,7 +3933,7 @@ class Workplane(object):
         :param valign: vertical alignment
         :return: a CQ object with the resulting solid selected
 
-        The returned object is always a CQ object, and depends on whether combine is True, and
+        The returned object is always a Workplane object, and depends on whether combine is True, and
         whether a context solid is already defined:
 
         *  if combine is False, the new value is pushed onto the stack.
--- a/cadquery/cq.py
+++ b/cadquery/cq.py
@@ -3926,7 +3926,7 @@ class Workplane(object):
         :param cut: True to cut the resulting solid from the parent solids if found
         :param combine: True to combine the resulting solid with parent solids if found
         :param clean: call :py:meth:`clean` afterwards to have a clean shape
-        :param font: font name (default: Arial)
+        :param font: font name
         :param fontPath: path to font file
         :param kind: font type (default: regular)
         :param halign: horizontal alignment (default: center)
--- a/cadquery/cq.py
+++ b/cadquery/cq.py
@@ -3919,20 +3919,21 @@ class Workplane(object):
         """
         Create a 3D text
 
-        :param str txt: text to be rendered
-        :param distance: the distance to extrude, normal to the workplane plane
+        :param self:
+        :type self: Workplane
+        :param txt: text to be rendered
+        :param fontsize: size of the font in model units
+        :param distance: the distance to extrude or cut, normal to the workplane plane
         :type distance: float, negative means opposite the normal direction
-        :param float fontsize: size of the font
-        :param boolean cut: True to cut the resulting solid from the parent solids if found.
-        :param boolean combine: True to combine the resulting solid with parent solids if found.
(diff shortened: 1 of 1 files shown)


jmwright:
@marcus7070 Do the examples in this PR fall under the discussions you've been having about keeping examples in the docs up-to-date?

marcus7070:
Hmmmm, that's an interesting thought @jmwright. You could do a basic smoke test by using [pytest's doctest integration](https://docs.pytest.org/en/6.2.x/doctest.html) and changing the docstring to:
```python
 Examples:
>>> cq.Workplane().text("CadQuery",5,1)

Specify the font (name), and kind to use an installed system font:
>>> cq.Workplane().text("CadQuery",5,1,font="Liberation Sans Narrow", kind="italic")

Specify fontPath to use a font from a given file:
>>> cq.Workplane().text("CadQuery",5,1, fontPath="/opt/fonts/texgyrecursor-bold.otf")

Cutting text into a solid:
>>> cq.Workplane().box(8,8,8).faces(">Z").workplane().text("Z", 5,-1.0)
```
But I hate doing double duty (document and test) with docstrings like this because `cq.Workplane().text("CadQuery",5,1, fontPath="/opt/fonts/texgyrecursor-bold.otf")` is a good example but would fail in our test suite. That one isn't too hard to fix, but similar issues would crop up in lots of other docstrings. I'd rather let docstrings do one thing and do it well.

It's also of limited benefit since it's pretty rare a docstring goes out of date with the function, since they're right next to each other.

So my vote is for no tests in docstrings.

cadquery/cq.py
```diff
@@ -3919,20 +3919,21 @@ def text(
         """
         Create a 3D text
 
-        :param str txt: text to be rendered
-        :param distance: the distance to extrude, normal to the workplane plane
+        :param self:
+        :type self: Workplane
```
marcus7070, line 3923:
```suggestion
```
At the moment `self` does not appear in the signature above this (in the HTML docs), and the return type is `~T`. It is less confusing to just leave it off.

cadquery/cq.py
```diff
@@ -3919,20 +3919,21 @@ def text(
...
-        :param str kind: font type (default: Normal)
-        :param str halign: horizontal alignment (default: center)
-        :param str valign: vertical alignment (default: center)
-        :return: a CQ object with the resulting solid selected.
-
-        extrude always *adds* material to a part.
+        :param cut: True to cut the resulting solid from the parent solids if found
+        :param combine: True to combine the resulting solid with parent solids if found
+        :param clean: call :py:meth:`clean` afterwards to have a clean shape
+        :param font: font name (default: Arial)
```
marcus7070, line 3931:
```suggestion
        :param font: font name
```
Defaults are already shown in the signature.

cadquery/cq.py
```diff
@@ -3919,20 +3919,21 @@ def text(
...
+        :param fontPath: path to font file
+        :param kind: font type (default: regular)
```
marcus7070, line 3933:
```suggestion
        :param kind: font type
```

cadquery/cq.py
```diff
@@ -3919,20 +3919,21 @@ def text(
...
+        :param halign: horizontal alignment (default: center)
```
marcus7070, line 3934:
```suggestion
        :param halign: horizontal alignment
```

cadquery/cq.py
```diff
@@ -3919,20 +3919,21 @@ def text(
...
+        :param valign: vertical alignment (default: center)
```
marcus7070, line 3935:
```suggestion
        :param valign: vertical alignment
```

cadquery/cq.py
```diff
@@ -3919,20 +3919,21 @@ def text(
...
+        :return: a CQ object with the resulting solid selected
```
marcus7070, line 3936:
```suggestion
        :return: a Workplane object with the resulting solid selected
```
We used to have a class called `CQ`, but that has been absorbed into the Workplane class.

cadquery/cq.py
```diff
@@ -3919,20 +3919,21 @@ def text(
...
 
         The returned object is always a CQ object, and depends on whether combine is True, and
```
marcus7070, line 3938:
```suggestion
        The returned object is always a Workplane object, and depends on whether combine is True, and
```

cadquery/cq.py
```diff
@@ -3941,6 +3942,22 @@ def text(
         *  if combine is true, the value is combined with the context solid if it exists,
            and the resulting solid becomes the new context solid.
 
+        Examples::
+
+            cq.Workplane().text("CadQuery",5,1)
+
+        Specify the font (name), and kind to use an installed system font::
+
+            cq.Workplane().text("CadQuery",5,1,font="Liberation Sans Narrow", kind="italic")
```
marcus7070, line 3951:
```suggestion
            cq.Workplane().text("CadQuery", 5, 1, font="Liberation Sans Narrow", kind="italic")
```

cadquery/cq.py
```diff
@@ -3941,6 +3942,22 @@ def text(
...
+
+        Specify fontPath to use a font from a given file::
+
+            cq.Workplane().text("CadQuery",5,1, fontPath="/opt/fonts/texgyrecursor-bold.otf")
```
marcus7070, line 3955:
```suggestion
            cq.Workplane().text("CadQuery", 5, 1, fontPath="/opt/fonts/texgyrecursor-bold.otf")
```

cadquery/cq.py
```diff
@@ -3941,6 +3942,22 @@ def text(
...
+
+        Cutting text into a solid::
+
+            cq.Workplane().box(8,8,8).faces(">Z").workplane().text("Z", 5,-1.0)
```
marcus7070, line 3959:
```suggestion
            cq.Workplane().box(8, 8, 8).faces(">Z").workplane().text("Z", 5, -1.0)
```

cadquery/cq.py
```diff
@@ -3941,6 +3942,22 @@ def text(
         *  if combine is true, the value is combined with the context solid if it exists,
            and the resulting solid becomes the new context solid.
 
+        Examples::
+
+            cq.Workplane().text("CadQuery",5,1)
```
marcus7070, line 3947:
```suggestion
            cq.Workplane().text("CadQuery", 5, 1)
```

marcus7070 (changes requested):
The only important change is that `:param self:` one, the rest are trivial stuff but I was suggesting changes anyway so I thought I'd throw them in.

Thanks for doing this @lorenzncode!

cadquery/cq.py
lorenzncode: Thank you for raising this issue regarding `:param self`!  This was my workaround for the issue where the Parameters section is duplicated in the HTML docs.  I will raise this as a separate issue as it affects many other methods (see for example Workplane.box, cboreHole, center, chamfer, circle to name a few).  Perhaps there is another way to solve it.

cadquery/cq.py
lorenzncode: I'll try to remember to run the black formatter next time.

cadquery/cq.py
marcus7070: Ah, ok, I thought the duplicates were due to indentation or something.

Please do raise the issue, a general solution would be better than manually documenting every self.

marcus7070 (approved):
Thanks for doing this @lorenzncode, even if we didn't get everything done it's still an improvement.

cadquery/cq.py
```diff
@@ -3919,28 +3919,43 @@ def text(
...
-        :param str valign: vertical alignment (default: center)
-        :return: a CQ object with the resulting solid selected.
-
-        extrude always *adds* material to a part.
-
-        The returned object is always a CQ object, and depends on whether combine is True, and
+        :param cut: True to cut the resulting solid from the parent solids if found
+        :param combine: True to combine the resulting solid with parent solids if found
+        :param clean: call :py:meth:`clean` afterwards to have a clean shape
+        :param font: font name
```
adam-urbanczyk, line 3929: What is the rationale for removing default values?

marcus7070: They are shown in the signature on HTML docs. 
screenshot2021-07-09-153054
I'm not aware of anywhere that you could see this doc string and not the signature, but let me know if there is an IDE or something where it's a problem.

adam-urbanczyk: OK, I'm not sure that it is better for the reader. I agree though that it'd be nice if they were automatically generated from the method definition.

marcus7070: > it'd be nice if they were automatically generated from the method definition

Here is a way to do that (at least when using Sphinx's inbuilt autodoc-typehints):

```python
def process_docstring_add_defaults(app, what, name, obj, options, lines):
    import inspect
    import re
    if what == "method":
        sig = inspect.signature(obj)
        for idx in range(len(lines)):
            pattern = re.compile(r"^\s*:param\s(\w*):")
            match = pattern.search(lines[idx])
            if match is None:
                continue
            arg = match.group(1)
            if sig.parameters[arg].default != inspect.Parameter.empty:
                lines[idx] += f" (default is {sig.parameters[arg].default})"

def setup(app):
    app.connect("autodoc-process-docstring", process_docstring_add_defaults)
```

But I'm getting off topic for this PR.

adam-urbanczyk: OK, maybe good to open it.

adam-urbanczyk (approved):

adam-urbanczyk:
Thanks @lorenzncode ! Shall I merge @marcus7070 @jmwright ?

jmwright:
+1 to merge

Issue #807, opened 2021-06-27
Travis changed billing


adam-urbanczyk:
Travis changed billing to a point based system, and we consumed the free points really quickly. I'll ask them if they can increase the number of points. Otherwise we'll need to switch to something else (I'd move to appveyor as they support remote log in for debugging).

jmwright:
Addressed in #816

PR #816, opened 2021-07-06
Added MacOS and Linux to Appveyor test matrix

jmwright:
Aimed at addressing #807 

Adds Linux and MacOS to the matrix. I have validated the file with Appveyor's validator, but I may need help figuring out how to get Appveyor to run the new config without merging this into master.


jmwright:
In each Python 3.9 environment I get the following mypy error. It has something to do with this code  `w: Union[Shape, Workplane],`.
Screenshot from 2021-07-06 22-32-00
[Here](https://github.com/CadQuery/cadquery/blob/master/cadquery/occ_impl/exporters/__init__.py#L37) is a quick link. Any ideas?

I can remove Python 3.9 from the matrix, but CadQuery seems to work fine with 3.9 through Miniconda.

jmwright:
The normal solution when Python 3.9 was newer was to upgrade mypy to a newer version, but that was during the 0.7xx days and we're running 0.9xx. I'm going to try some different versions to see if there is a regression in the version we use.

jmwright:
I installed mypy via pip rather than conda and that fixed the problem.

appveyor.yml
```diff
@@ -1,31 +1,67 @@
...
 install:
-    - set "PATH=%MINICONDA_DIRNAME%;%MINICONDA_DIRNAME%\\Scripts;%PATH%"
-    - conda config --set always_yes yes
+    - conda config --set always_yes yes --set changeps1 no
     - conda update -q conda
     - conda env create -f environment.yml
     - activate cadquery
-    - conda install -c conda-forge -c defaults -c cadquery python=%PYTHON_VERSION%
+    - cmd: conda install -c conda-forge -c defaults -c cadquery cadquery=master python=%PYTHON_VERSION%
+    - sh: conda install -c conda-forge -c defaults -c cadquery cadquery=master python=$PYTHON_VERSION
```
marcus7070, line 50: Doesn't this install both `cadquery=master` and `python=3.9`? I don't think we want to install CQ from Conda here.

marcus7070:
> I installed mypy via pip rather than conda and that fixed the problem.

If I put mypy into the environment.yml then Conda managed to get the right version (during `conda-env create ...`). It did not work with `conda install mypy>0.9`. You could add
```bash
sed -i 's/  - pip:/  - mypy>0.9\n  - black=19.10b0\n  - pip:/' environment.yml
```
before the `conda-env create` command.

appveyor.yml
jmwright: My understanding is that it will install the current master branch of CQ and will use Python 3.9 for the environment. The goal is to test the latest CQ with a specific Python version.

jmwright: I think I see your point. The Action already checked out CadQuery's repo so this isn't needed, correct?

appveyor.yml
```diff
@@ -1,31 +1,65 @@
...
 platform:
     - x64
 
 environment:
     matrix:
-        - PYTHON_VERSION: 3.7
+        - APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2019
           MINICONDA_DIRNAME: C:\Miniconda37-x64
-        - PYTHON_VERSION: 3.8
+          PYTHON_VERSION: 3.7
```
adam-urbanczyk, line 10: Do we need to test on so many versions? Same tests (minus mypy/black) are execute with azure. Maybe just a single python version would be enough and would definitely decrease the runtime.

adam-urbanczyk (approved):

appveyor.yml
jmwright: So, Python 3.8?

jmwright: I guess maybe just let conda pick the version? That would simplify things for me.

adam-urbanczyk: Sounds good

appveyor.yml
marcus7070: This workflow is being run on every PR, right (I'm not sure where AppVeyor gets those instructions from)? 

We don't want to accidentally run any tests on the Conda version instead of the PR version. Or if we dropped something essential from `environment.yml` in a PR, this line would still install every dependency from master and tests would still pass.

marcus7070:
> You could add
> 
> ```shell
> sed -i 's/  - pip:/  - mypy>0.9\n  - black=19.10b0\n  - pip:/' environment.yml
> ```
> 
> before the `conda-env create` command.

Oh thank goodness you went away from this @jmwright, it was a bad idea. Changing `environment.yml` here would create a different environment to master. Much better to keep the master and testing environments the same.

jmwright:
So it turns out (I guess I should have known this already) that when you install a specific Python version with `conda` after installing with environment.yml, Miniconda will go through and update/replace the packages for the previous Python version. This requires a whole new solve that can shuffle package versions. In a couple of my tests it errored out with unsolvable conflicts even though the very next run would work fine. I also had one case on AppVeyor where the solve hung and the job timed out after 1 hour. The new version is greatly simplified and hopefully addresses these solve issues by not trying to install a different version of Python after the environment is already built based on the yaml file.

Do we want to wait on any other PRs to be merged before merging this one?

marcus7070 (approved):

marcus7070:
> Do we want to wait on any other PRs to be merged before merging this one?

No, all LGTM.

jmwright:
Ok I'll merge so that we can stop burning Travis points sooner rather than later.

adam-urbanczyk:
Thanks @jmwright !

jmwright:
Ooops, I forgot to hit the Merge button earlier when I said I was going to.

PR #821, opened 2021-07-11
doc/primer: added introspective example

marcus7070:
Please read [the Read The Docs preview](https://cadquery--821.org.readthedocs.build/en/821/primer.html#an-introspective-example) before reading the text diff. The text diff is a bit difficult to read since I used a lot of Sphinx's roles to provide both automatic links to autodocs and formatting. I really like the role formatting since I find it confusing sometimes if text refers to the concept of a workplane or the class `Workplane` (same for solids, faces, edges, planes, objects/`Workplane.objects`).

### DIFF
--- a/doc/primer.rst
+++ b/doc/primer.rst
@@ -257,13 +257,13 @@ Which produces the output:
 
 .. code-block:: none
 
-    Workplane object at 0x7f5c9b822760:
+    Workplane object at 0x2760:
       no parent
-      plane: Plane object at 0x7f5c9b822850:
+      plane: Plane object at 0x2850:
         origin: (0.0, 0.0, 0.0)
         z direction: (0.0, 0.0, 1.0)
       objects: []
-      modelling context: CQContext object at 0x7f5c9b822730:
+      modelling context: CQContext object at 0x2730:
         pendingWires: []
         pendingEdges: []
         tags: {}
@@ -281,13 +281,13 @@ Which produces the output:
 
 .. code-block:: none
 
-    Workplane object at 0x7f5c368eaa90:
-      parent: Workplane object at 0x7f5c9b822760
-      plane: Plane object at 0x7f5c368f3850:
+    Workplane object at 0xaa90:
+      parent: Workplane object at 0x2760
+      plane: Plane object at 0x3850:
         origin: (0.0, 0.0, 0.0)
         z direction: (0.0, 0.0, 1.0)
-      objects: [<cadquery.occ_impl.shapes.Solid object at 0x7f5c9b94bbe0>]
-      modelling context: CQContext object at 0x7f5c9b822730:
+      objects: [<cadquery.occ_impl.shapes.Solid object at 0xbbe0>]
+      modelling context: CQContext object at 0x2730:
         pendingWires: []
         pendingEdges: []
         tags: {}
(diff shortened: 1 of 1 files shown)


marcus7070:
@jojain, @fischman, you also were interested in this when I sent it to Google Groups, do you want to have a look here too?

I tried the `sys.setprofile` method you used @fischman but it wound up being more more code and explanation than this simple print method.

jmwright (approved):
This is great @marcus7070 , thanks for writing it. It would be cool if we could somehow give users the ability to use the patched string representations you created to get more information on their chain and modelling context. Maybe even a way to print out the entire stack so they can trace what's happening. I know that CQ-editor already has this covered pretty well, but being able to print the results of each call in the way you have presented could very illuminating for people who are only using the core library.

marcus7070:
Thanks @jmwright.

> It would be cool if we could somehow give users the ability to use the patched string representations you created to get more information on their chain and modelling context.

I'm actually a big fan of just cutting and pasting that bit of code into a user's CQ script. It's reliable and easy, even if it looks a bit ugly.

> Maybe even a way to print out the entire stack so they can trace what's happening.

That's an interesting idea. It would be a better fit for the fluent api (rather than stopping at every step to call print, which was easier for an example in the docs but is horrible for a user). Sounds like it would be a good fit for the cadquery-plugins repo.

fedorkotov:
@marcus7070 
Don't know if this should be reported here or as a separate bug. When the model from your example is viewed from a certain angle some lines disappear
image

doc/primer.rst
```diff
@@ -165,6 +168,410 @@ iterates on each member of the stack.
...
+      modelling context: CQContext object at 0x7f5c9b822730:
+        pendingWires: []
+        pendingEdges: []
+        tags: {'base': <cadquery.cq.Workplane object at 0x7f5c368eaa90>}
+
+The :attr:`~cadquery.cq.CQContext.tags` attribute of the modelling context is simply a dict
+associating the string name given by the :meth:`~cadquery.Workplane.tag` method to the
+:class:`~cadquery.Workplane`. Methods such as :meth:`~cadquery.Workplane.workplaneFromTagged` and
+selection methods like :meth:`~cadquery.Workplane.edges` can operate on a tagged
+:class:`~cadquery.Workplane`. Note that the :meth:`~cadquery.Workplane.tag` method is unusual for a
```
Jojain, line 328: I don't really understand the "Note",  Maybe adding an example here would make it more clear

Jojain: Thanks @marcus7070 from the questions that have been asked in the discord I feel that a lot of them would be resolved by the reading of this part of the docs !

marcus7070: 3c43519503cb42a74e31acd482d766ef0fa1bb61 is aimed at making this more clear, does it help?

marcus7070:
@fedorkotov 
> Don't know if this should be reported here or as a separate bug.

It's not limited to this PR, you can recreate the same dissapearing line with this example: https://cadquery.readthedocs.io/en/latest/examples.html#mirroring-from-faces

I think it's just how VTK renders edges; if the two adjacent faces have the same brightness then the edge isn't visible. We don't explicitly ask VTK to render edges, so... not a bug, I guess?

There might be something we can do about it though, VTK has a lot of options.

adam-urbanczyk (approved):

adam-urbanczyk:
Looks good @marcus7070 ! Is it ready for merging?

PR #829, opened 2021-07-18
Mixin3D.shell: removed deprecated OCCT method

### LINKED ISSUE #649: Deprecated method used in Mixin3D.shell
From https://old.opencascade.com/doc/occt-7.4.0/refman/html/class_b_rep_offset_a_p_i___make_thick_solid.html#a1b559bffad21d24f792772e87625fecb, the constructor used in `Mixin3D.shell` is deprecated, and needs to be replaced with an empty constructor and `MakeThickSolidByJoin` method (I think?).

https://github.com/CadQuery/cadquery/blob/e4ed8816171e429afc511602049e74beef388640/cadquery/occ_impl/shapes.py#L2023-L2083

It's a long shot, but perhaps this will help with #638, where we struggled to shell a shape that had been transformed.

marcus7070:
Will close #649.

I also shortened the code a bit, there was a few duplicated calls.

### DIFF
--- a/cadquery/occ_impl/shapes.py
+++ b/cadquery/occ_impl/shapes.py
@@ -2351,7 +2351,7 @@ class Mixin3D(object):
 
     def shell(
         self: Any,
-        faceList: Iterable[Face],
+        faceList: Optional[Iterable[Face]],
         thickness: float,
         tolerance: float = 0.0001,
         kind: Literal["arc", "intersection"] = "arc",
@@ -2375,8 +2375,9 @@ class Mixin3D(object):
         occ_faces_list = TopTools_ListOfShape()
         shell_builder = BRepOffsetAPI_MakeThickSolid()
 
-        for f in faceList:
-            occ_faces_list.Append(f.wrapped)
+        if faceList:
+            for f in faceList:
+                occ_faces_list.Append(f.wrapped)
 
         shell_builder.MakeThickSolidByJoin(
             self.wrapped,
--- a/cadquery/occ_impl/shapes.py
+++ b/cadquery/occ_impl/shapes.py
@@ -2357,12 +2357,14 @@ class Mixin3D(object):
         kind: Literal["arc", "intersection"] = "arc",
     ) -> Any:
         """
-            make a shelled solid of given  by removing the list of faces
+        Make a shelled solid of self.
 
-        :param faceList: list of face objects, which must be part of the solid.
-        :param thickness: floating point thickness. positive shells outwards, negative shells inwards
-        :param tolerance: modelling tolerance of the method, default=0.0001
-        :return: a shelled solid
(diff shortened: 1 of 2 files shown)


cadquery/occ_impl/shapes.py
```diff
@@ -2371,45 +2373,38 @@ def shell(
         }
 
         occ_faces_list = TopTools_ListOfShape()
+        shell_builder = BRepOffsetAPI_MakeThickSolid()
 
-        if faceList:
```
jmwright, line 2375: Why did you remove this check to make sure faceList is not None? I see that same check used below.

jmwright (approved):
I just had the one question. Other than that it looks good. Thanks @marcus7070

cadquery/occ_impl/shapes.py
marcus7070: I was going by the type hinting on the signature (`faceList: Iterable[Face],`), so I didn't think `None` was an option.
But you're right, the way that code was originally written it would work with `None` so I should probably change the type hinting to `Optional[Iterable[Face]]` and handle it.

marcus7070: e41b0300138350e50870292fedf63b606587e68a should take care of this.

cadquery/occ_impl/shapes.py
```diff
@@ -2351,18 +2351,20 @@ def chamfer(
 
     def shell(
         self: Any,
-        faceList: Iterable[Face],
+        faceList: Optional[Iterable[Face]],
```
adam-urbanczyk, line 2354: Empty list evaluates to false - not need to make this Optional.

cadquery/occ_impl/shapes.py
```diff
@@ -2371,45 +2373,39 @@ def shell(
...
+            thickness,
+            tolerance,
+            Intersection=True,
+            Join=kind_dict[kind],
+        )
+        shell_builder.Build()
 
-            shell_builder.Build()
-            rv = shell_builder.Shape()
+        if faceList:
```
adam-urbanczyk, line 2392: NB: if [] == if False

cadquery/occ_impl/shapes.py
marcus7070: Indeed, but that wasn't why I added the Optional. The way this function was originally written it would have worked with `faceList=None`. Later it was typehinted as Iterable.

The way I wrote it in my first commit `faceList=None` would have raised a TypeError at `for f in faceList:`. I thought it safer to handle `None` incase there was any user doing it already.

Happy to just go back to my first commit if you are.

adam-urbanczyk: OK, sounds good

adam-urbanczyk (approved):

PR #844, opened 2021-08-15
Doc fixes (add wedge)

revolter:


jmwright:
@revolter Why did you remove the mention of the wedge primitive?

Wedge: https://github.com/CadQuery/cadquery/blob/4c45eb20e3576a2cff74f30c76bc96bf0f58561a/cadquery/cq.py#L3803

revolter:
Because that section has
> See 3-d Operations to learn more.

at the end, which links to https://cadquery.readthedocs.io/en/latest/apireference.html#id1, which doesn't mention wedges nor cylinders.

jmwright:
Ok, so we'll need to add the missing a reference to wedge under _3-d Operations_ at some point.

revolter:
Is it a complicated task? Also, what about cylinder?

jmwright:
> Is it a complicated task?

No, you can just add it below `Workplane.sphere` under the autosummary [here](https://github.com/CadQuery/cadquery/blob/4c45eb20e3576a2cff74f30c76bc96bf0f58561a/doc/apireference.rst#3-d-operations).

> Also, what about cylinder?

I did a search to confirm and there is no analogous method for `Workplane.cylinder`.

revolter:
Like this?

https://cadquery--844.org.readthedocs.build/en/844/primer.html#id1

=>

https://cadquery--844.org.readthedocs.build/en/844/apireference.html#id1

jmwright:
> Like this?

Yes, that looks correct to me.

Looks like we're going to have to correct a mypy issue that's probably caused by the update to OCP 7.5.2. I'll try to get to that today or tomorrow.

jmwright:
@adam-urbanczyk I think it may be the missing OCP 7.5.2 stubs that are keeping mypy from passing in these tests. Could you confirm when you are available?

adam-urbanczyk (approved):

jmwright (approved):
Thanks @revolter !

revolter:
And thank you @adam-urbanczyk for updating my PR 😃

PR #850, opened 2021-08-18
doc/assy: add constraints

### LINKED ISSUE #715: Documentation & examples for each assembly constraint
As we add more constraints I think it makes sense to have a section of the HTML docs showing what each constraint does.

marcus7070:
Will resolve #715.

This is still a WIP, but it's mostly done and I won't be able to get back to it for a few days, so I figured I'd open this draft PR incase anyone wants to check it out.

Still to do:

- [x] Consider a more gentle introduction to constraints (Why you would use them? Define what I mean when I say first object, second object?)
- [x] ~~Introduce dummy geometry~~ #856
- [x] Run my code examples through black
- [x] Is that Plane 3D example really worth it? Should I just replace it with a 3 line code block?
- [x] ~~Explain why rotation has to be locked with PointInPlane, but you can get away without doing it for just Point~~ *Not going to add this, it reads better without it*
- [x] Consider applying better formatting to the list of colors (flexbox or something?)
- [x] Docstring at the top of `gen_colors.py`, it's not obvious what it's used for
- [x] Explain initial locations

### DIFF
--- a/doc/assy.rst
+++ b/doc/assy.rst
@@ -670,8 +670,12 @@ previous example from the two constraints to just one:
     show_object(assy)
 
 
+The result of this code is identical to the above two constraint example.
+
 For the cost function of Plane, please see the Point and Axis sections. The ``param`` argument is
-copied into both constraints and should be left as the default value of ``None``.
+copied into both constraints and should be left as the default value of ``None`` for a "mate" style
+constraint (two surfaces touching) or can be set to ``0`` for a through surface constraint (see
+desciption in the Axis constraint section).
 
 
 PointInPlane
--- a/doc/assy.rst
+++ b/doc/assy.rst
@@ -389,11 +389,7 @@ Objects can be added to an assembly with initial locations supplied, such as:
     assy = cq.Assembly()
     assy.add(
         cone,
-        loc=cq.Location(
-            cq.Vector(0, 0, 0),
-            cq.Vector(1, 0, 0),
-            180,
-        ),
+        loc=cq.Location(cq.Vector(0, 0, 0), cq.Vector(1, 0, 0), 180),
         name="cone0",
         color=cq.Color("green")
     )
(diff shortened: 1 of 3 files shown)


Jojain:
It could be nice to have  a note that says something like :
- Since solve() is computing an optimization, it is sensible to initials parameters. Thus having your shapes positioned roughly were you want them to be can give better results. Especially if all your part are overlapping at the beginning.

(I am not 100% sure this is true though lol)

marcus7070:
> * Since solve() is computing an optimization, it is sensible to initials parameters. Thus having your shapes positioned roughly were you want them to be can give better results/ make the assembly converge. Especially if all your parts are overlapping at the beginning.

Thanks @Jojain. Yeah, I won't word it exactly like that (I think no initial locations is fine for the majority of uses), but if I'm explaining how constraints work then I should also explain initial locations as well.

jmwright:
@marcus7070 Still seeing the following black error in CI.
```
would reformat doc\conf.py
Oh no! \U0001f4a5 \U0001f494 \U0001f4a5
1 file would be reformatted, 67 files would be left unchanged.
Command exited with code 1
```

jmwright (approved):
LGTM, thanks @marcus7070 !

adam-urbanczyk (approved):
Great, looks good! Just one more black run and all should be green.

marcus7070:
> @marcus7070 Still seeing the following black error in CI.
> 
> ```
> would reformat doc\conf.py
> Oh no! \U0001f4a5 \U0001f494 \U0001f4a5
> 1 file would be reformatted, 67 files would be left unchanged.
> Command exited with code 1
> ```

Thanks, I just had to add a comma to a list. Assuming this all passes I'll merge in a few minutes.

marcus7070:
Just a leftover error because this branch doesn't contain #872, so I'll merge.

PR #852, opened 2021-08-20
Fix inconsistent 2D and 3D spellings

revolter:


marcus7070 (approved):
Thanks @revolter!

jmwright (approved):
Thanks @revolter 

@marcus7070 We can merge this if you want, but I would like to keep one of these PRs that's failing on the mypy check open so that we can use it to make sure we have fixed the mypy/OCP-stubs issue.

PR #859, opened 2021-08-25
Added cone 3D primitive to Workplane.

martinbudden:
I've added a cone 3D primitive to the `Workplane`, as per item on the roadmap, https://github.com/CadQuery/cadquery/blob/master/doc/roadmap.rst

This follows the OpenSCAD style, so if a single radius is specified a normal cone is created, whereas if `radius1` and `radius2` is specified then a truncated cone is created.

So `cq.Workplane("XY").cone(40, 10)` creates a cone with height 40 and radius 10
whereas `cq.Workplane("XY").cone(40, radius1=10, radius2=5)` creates a truncated cone with bottom radius 10 and top radius 5.

### DIFF
--- a/cadquery/cq.py
+++ b/cadquery/cq.py
@@ -3816,12 +3816,12 @@ class Workplane(object):
         A truncated cone can be created by specifying parameters radius1 and radius2 instead of radius.
         :param height: The height of the cone
         :type height: float > 0
-        :param radius: The radius of the cone
-        :type radius: float > 0
         :param radius1: The radius of the bottom of the cone
         :type radius1: float > 0
         :param radius2: The radius of the top of the cone
         :type radius2: float > 0
+        :param radius: The radius of the cone
+        :type radius: float > 0
         :param direct: The direction axis for the creation of the cone
         :type direct: A three-tuple
         :param angle: The angle to sweep the cone arc through


jmwright (approved):
Thanks @martinbudden

cadquery/cq.py
```diff
@@ -3800,6 +3800,62 @@ def sphere(
         else:
             return self.union(spheres, clean=clean)
 
+    def cone(
+        self: T,
+        height: float,
+        radius: float = 0,
```
adam-urbanczyk, line 3806: Please remove the redundancy (at most two radius params needed) and defaulting of radius/radius1 :

```
        radius1: float,
        radius2: Optional[float] =None
```

cadquery/cq.py
```diff
@@ -3800,6 +3800,62 @@ def sphere(
...
+        cone is created using the current workplane center.
+
+        If combine is true, the result will be a single object on the stack. If a solid was found
+        in the chain, the result is that solid with all cones produced fused onto it otherwise,
+        the result is the combination of all the produced cones.
+
+        If combine is false, the result will be a list of the cones produced.
+        """
+
+        r1 = radius if radius1 is None or radius1 == 0 else radius1
```
adam-urbanczyk, line 3845:
```suggestion
        r1 = radius1
```

adam-urbanczyk (changes requested):

cadquery/cq.py
martinbudden: There is a reason for this, it to allow the user to specify `cone(height=40, radius=10) ` for a simple cone and `cone(height=40, radius1=10, radius2=5)` for a truncated cone. This is akin to how OpenSCAD handles things.

fedorkotov: This is only convenient if argument names are always specified. But what about `cone(40,10)` for simple cone and `cone(40,10,5)` for truncated cone? `cone(40,10,5)` will give surprising results with three radius arguments.

martinbudden: True. I could solve that problem by ordering the aguments: `cone(height, radius1, radius2, radius)`. Then
```
cone(40,10)
cone(40,10,5)
cone(height=40,radius=10)
cone(height=40,radius1=10,radius2=5)
```
All behave as expected.

adam-urbanczyk: The API of OpenSCAD is not really relevant here (different language; no intended relation with CQ). What is very relevant on the other hand is consistence with existing functions. Taking this into account, I'd propose:
```python
def cone(
        self: T,
        radius1: float,
        height: float,
        radius2: float = 0.0,
       ...
```

The same holds for the recently merged cylinder (radius and height should be interchanged to be similar to e.g. the hole signature). It would be really great, if you could amend it in this PR

martinbudden: > The API of OpenSCAD is not really relevant here (different language; no intended relation with CQ).

While I agree that there is no need to be the same as OpenSCAD, I don't believe other languages are not relevant - taking cues from how other established systems work can often suggest a pattern to use.

> What is very relevant on the other hand is consistence with existing functions

Agreed. The reason for consistency is not for its own sake, it is because consistency helps avoid errors. The parameter ordering `cone(radius1, height, radius2)` is horrible and error-inducing, thus defeating the reason for being consistent. And if we really want to follow other functions, eg `cboreHole` and `cskHole` then we should use `diameter` rather than `radius`. But we already have `sphere`, `fillet`, `ellipse` and `circle` which use `radius`.  So orthogonality is already broken.

Having said all that, I suggest:
```
def cone(
    self: T,
    height: float,
    radius1: float,
    radius2: float = 0.0,
...
```

This:
1. Uses `radius` and so is consistent with the other geometric primitives.
2. Allows a natural parameter ordering for truncated cone.
3. Does not have the contentious `radius` parameter.

adam-urbanczyk:
It is not perfect so let's make it even worse with regard to consistent
param ordering? Interesting reasoning but I don't agree. -1 on merging with
your proposal.

martinbudden:
> It is not perfect so let's make it even worse with regard to consistent param ordering?

No, that's not what I am saying at all.

What I am saying is that there are a number of factors that determine function signature, consistency is one of them and usability is another. Consistency is in fact a subset of usability, and so making something consistent but less usable makes no sense. (I also had a side argument that defining what is consistent is somewhat subjective when the set is already inconsistent.)

adam-urbanczyk:
Our understanding of usability is very different.

fedorkotov:
@adam-urbanczyk please reconsider `radius1`, `height`, `radius2` argument order. IMHO it is really non-obvious and surprising. 
Look at GUIs and APIs of other somewhat similar software
https://docs.blender.org/api/current/bpy.ops.mesh.html#bpy.ops.mesh.primitive_cone_add
https://docs.salome-platform.org/latest/gui/GEOM/create_cone_page.html

Of course CadQuery does not have to follow traditions and conventions of other products but look at 
https://github.com/CadQuery/cadquery/blob/4c45eb20e3576a2cff74f30c76bc96bf0f58561a/cadquery/occ_impl/shapes.py#L2582 and BRepPrimAPI_MakeCone
https://github.com/CadQuery/OCP/blob/e1df3469a39228ce2553aacc3a0238e334d4dd9a/opencascade/BRepPrimAPI_MakeCone.hxx#L49 
Both of them have radii close together not and not separated by height.

adam-urbanczyk:
Radii are separated due to  the 2nd radius being keyword arg with 0 default
value.

fedorkotov:
@martinbudden I think that having a cone primitive creation method in Workplane is better than not having one. And so suggest you follow @adam-urbanczyk  recommendations even if you do not agree so that this can be merged sooner rather than later (or never). If you have very strong opinion on this matter, there is also https://github.com/CadQuery/cadquery-plugins repo. Plugins don't have to conform to the same standards as "built in" features of CadQuery.

adam-urbanczyk:
Another option would be to wait for sketch branch being merged (and implicitly multimethods) and implement two overloads `r,h` and `r1,r2,h`. It will take some time though.

martinbudden:
> Another option would be to wait for sketch branch being merged (and implicitly multimethods) and implement two overloads r,h and r1,r2,h. It will take some time though.

I think I'd rather wait for the implementation of multimethods than add the legacy of a method with the parameter ordering of `r1, h, r2`.

Konubinix:
Hi. I'm not fluent in cadQuery and just wanted to make a cone the same way I make other primitives, with Workpkane().cone(...). IIUC, this PR enables this, right ?

adam-urbanczyk:
Indeed, it is waiting to be reworked so that it provides two overloads `cone(r, h)` and `cone(r1, r2, h)`. Seems though that the OP lost interest in the meantime, so if you want to pick this up you are more than welcome.

Konubinix:
AU ***@***.***> writes:

> Indeed, it is waiting to be reworked so that it provides two overloads `cone(r, h)` and `cone(r1, r2, h)`.

Thank you for the answer. I could manage with ~cq.Workplane("XY", obj=cq.Solid.makeCone(coneinnerdiameter/2, coneouterdiameter/2, coneheight))~

cadquery still looks a bit like magic to me, but the documentation is
pretty nice and the examples help a lot. I'm always amazed to find out
that my (very simple) experiments can be done with this tool. Kudo for
the great work.

-- 
Konubinix
GPG Key    : 7439106A
Fingerprint: 5993 BE7A DA65 E2D9 06CE  5C36 75D2 3CED 7439 106A

PR #875, opened 2021-08-31
until extrude/cutblind

Jojain:
Final PR that would be hopefully totally clean this time.
Sorry for the duplicates, I'm note very familiar with git / github so I made some mistakes along the way.


jmwright:
@Jojain If you run black on this locally does it pass? CI is complaining.

Jojain:
@jmwright 
I ran it before sending the PR.
Maybe not from top of the repo though I'll take a look tonight.

Jojain:
@jmwright is there a specific version of black to use ?
Appveyor was saying he would reformat 6 files, my black run reformatted 4 files and now Appveyor says he would reformat 9 files :roll_eyes:

jmwright:
Yes, the version of black is pinned for CI. Please see [this source](https://github.com/CadQuery/cadquery/blob/master/environment.yml#L14).

Jojain:
It's back to where it was know with only coverage failing. 
Codecov complains about some errors raising not tested, however most of them don't need tests in my opinion as they are just checks on wrong parameters combination.

I'll be adding docs and examples about this once it's merged and there is no API change

adam-urbanczyk:
Thanks, I still see though the unwanted test files being committed. Let me check how to get rid of those.

cadquery/cq.py
```diff
@@ -3441,48 +3480,130 @@ def loft(
...
+                        Face.makeFromWires(ws[0]),
+                        ws,
+                        taper=taper,
+                        upToFace=limitFace2,
+                        additive=additive,
+                    )
+                    thisObj = Compound.makeCompound([thisObj, thisObj2])
+                toFuse = [thisObj]
+            else:
+                thisObj = Solid.extrudeLinear(ws[0], ws[1:], eDir, taper=taper)
```
adam-urbanczyk, line 3600: This line changes the previous behavior - is it intended?

cadquery/occ_impl/shapes.py
```diff
@@ -3016,21 +3097,24 @@ def sweep_multi(
         return cls(builder.Shape())
 
     def dprism(
-        self,
+        self: Union[Shape, "Solid"],
```
adam-urbanczyk, line 3100:
```suggestion
        self,
```

adam-urbanczyk: This cannot be correct - dprism is not a method of Shape.

adam-urbanczyk:
> It's back to where it was know with only coverage failing.
> Codecov complains about some errors raising not tested, however most of them don't need tests in my opinion as they are just checks on wrong parameters combination.

It'd be good to trigger those errors in tests.

cadquery/cq.py
Jojain: Are you referring to the `ws[1:]` being passed in for a tapered extrude ?
I didn't noticed there were this difference in the old behaviour, I will modify that.

cadquery/occ_impl/shapes.py
Jojain: I took another look at that and I would need your help here as I don't know how I can handle things.

I wrote `self: Union[Shape, "Solid"]` because at one point in the operation I retrieve the solid from the workplane with `self.findSolid()` that I then pass as the first argument of `Solid.dprism` however `self.findSolid()` return type is `Union[Compound, "Solid"]` so I was a bit stuck.

Since `BRepFeat_MakeDPrism (const TopoDS_Shape &Sbase, const TopoDS_Face &Pbase, const TopoDS_Face &Skface, const Standard_Real Angle, const Standard_Integer Fuse, const Standard_Boolean Modify)` accepts a Shape it will work with either a compound or a Solid (It should even work with a Face too)

I see some options that I don't find great : 

- Add (or modify the existing one) a dprism static method that would take Compound or Solid as argument
- Add dprism method to Compound
- Remove the possibility of extrude until to extrude to something else than a solid (-> bad idea)
- Tweak the typing in a way that make it works (I don't know how)

I'm up to implement any solution you come up with

adam-urbanczyk:
One more file to go: `tests/test.brep`

Jojain:
Wouldn't it be nice to add those files in the .gitignore ? Each pytest run make them pop again, and I could recommit them by mistake

jmwright:
> Wouldn't it be nice to add those files in the .gitignore ?

The .gitignore on master has these files. It was a recent merge.

cadquery/cq.py
adam-urbanczyk: Yes, that is what I mean. Did you resolve it in the newest commits?

cadquery/occ_impl/shapes.py
adam-urbanczyk: If you are done, I'll try to edit your branch to fix this.

cadquery/cq.py
Jojain: Yes it's now handled like it is on master !

cadquery/occ_impl/shapes.py
Jojain: I am done, thanks

cadquery/cq.py
```diff
@@ -3441,48 +3478,133 @@ def loft(
...
+            if additive:
+                direction = "AlongAxis"
+            else:
+                direction = "Opposite"
+
+            facesList = self.findSolid().facesIntersectedByLine(
+                ws[0].Center(), eDir, direction=direction
+            )
+            if len(facesList) == 0 and both:
+                raise ValueError(
```
adam-urbanczyk, line 3516: Would you be able to provide a simple test vase for this? I think cq.Workplane().rect(1,1).extrude('next') would trigger this error, right?

adam-urbanczyk (approved):

adam-urbanczyk:
@marcus7070 do you plan to take a look?

marcus7070:
Yes, I'll try to get it finished in the next few hours.

cadquery/cq.py
```diff
@@ -3441,48 +3478,133 @@ def loft(
...
+                baseSolid = self.findSolid() if baseSolid is None else thisObj
+                if isinstance(upToFace, int):
+                    facesList = getFacesList(eDir, additive, both=both)
+                    limitFace = facesList[upToFace]
+                    if (
+                        baseSolid.isInside(ws[0].Center())
+                        and additive
+                        and upToFace == 0
+                    ):
+                        upToFace = 1  # extrude until next face outside the solid
```
marcus7070, line 3571: I may be misunderstanding your intention here, but I think you meant:
```suggestion
                    if (
                        baseSolid.isInside(ws[0].Center())
                        and additive
                        and upToFace == 0
                    ):
                        upToFace = 1  # extrude until next face outside the solid
                        
                    limitFace = facesList[upToFace]
```

I would suggest a test is added for this as well.

cadquery/occ_impl/shapes.py
```diff
@@ -1034,6 +1036,85 @@ def intersect(self, *toIntersect: "Shape") -> "Shape":
...
+            else gp_Dir(axis.wrapped)
+        )
+
+        line = gce_MakeLin(oc_point, oc_axis).Value()
+        shape = self.wrapped
+
+        intersectMaker = BRepIntCurveSurface_Inter()
+        intersectMaker.Init(shape, line, tol)
+
+        faces_dist = []  # using a list instead of a dictionnary to be able to sort it
```
marcus7070, line 1073:
```suggestion
        faces_dist = []  # using a list instead of a dictionary to be able to sort it
```

cadquery/cq.py
```diff
@@ -3441,48 +3478,133 @@ def loft(
...
-        eDir = self.plane.zDir.multiply(distance)
+        if upToFace is not None:
+            eDir = self.plane.zDir
+        elif distance is not None:
+            eDir = self.plane.zDir.multiply(distance)
+
+        if additive:
+            direction = "AlongAxis"
+        else:
+            direction = "Opposite"
```
marcus7070, line 3545: Left over code?

```suggestion
```

cadquery/occ_impl/shapes.py
```diff
@@ -1034,6 +1036,85 @@ def intersect(self, *toIntersect: "Shape") -> "Shape":
...
+                    interDir is not None
+                    and interDir.IsOpposite(oc_axis, tol)
+                    and distance > tol
+                ):
+                    faces_dist.append((intersectMaker.Face(), distance))
+
+            elif direction is None:
+                faces_dist.append(
+                    (intersectMaker.Face(), abs(distance))
+                )  # will sort all intersected faces by distance whatever the direction is
```
marcus7070, line 1105: I think you can remove this `direction=None` case, I can't see anywhere you use it.

cadquery/cq.py
```diff
@@ -3441,48 +3478,133 @@ def loft(
...
         # but then cutting it from the main solid fails with BRep_NotDone.
         # the work around is to extrude each and then join the resulting solids, which seems to work
 
         # underlying cad kernel can only handle simple bosses-- we'll aggregate them if there are
         # multiple sets
+        thisObj: Union[Solid, Compound]
 
         toFuse = []
+        taper = 0.0 if taper is None else taper
+        baseSolid = None
```
marcus7070, line 3558:
```suggestion
```

As far as I can tell this is always overridden by line 3562.

cadquery/cq.py
```diff
@@ -3441,48 +3478,133 @@ def loft(
...
 
-        if taper:
-            for ws in wireSets:
-                thisObj = Solid.extrudeLinear(ws[0], [], eDir, taper)
+        for ws in wireSets:
+            if upToFace is not None:
+                baseSolid = self.findSolid() if baseSolid is None else thisObj
```
marcus7070, line 3562:
```suggestion
                baseSolid = self.findSolid()
```
I think `thisObj` is left over code?

marcus7070 (commented):
Thanks @Jojain, this was a big job and a great improvement. I have some minor points.

The multiple types in the docstrings are not working well in the HTML docs, see https://cadquery--875.org.readthedocs.build/en/875/classreference.html#cadquery.Workplane.extrude

I would suggest changing it to have no explicit `:type until:` lines and instead combine all that information into the param line, like:
```rst
:param until: The distance to extrude, normal to the workplane plane. When a float is passed, the extrusion extends this far and a negative value is in the opposite direction to the normal of the plane. The string "next" extrudes until the next face orthogonal to the wire normal. "last" extrudes to the last face. If a object of type Face is passed then the extrusion will extend until this face.
```

cadquery/cq.py
Jojain: No the purpose of the `baseSolid = None` is to handle what the function will do when extruding several wires.

At the first iteration (and if there is only one pending wire) the function will extrude the wire using the solid found in the Workplane with `findSolid`.

If there is several wires, at each iteration the "baseSolid" passed to the `dprism` function is the result of the previous extrusion with the previous wire.

cadquery/cq.py
Jojain: As said earlier, this has a purpose and cannot be removed

cadquery/cq.py
Jojain: I think you misunderstood, this part of the code does : 
- A check if `upToface` is of type `int` or `Face` , if it's type `int` it checks that if an extrusion is required and the extrusion until the next face is required, if the wire is insinde the solid then we want to extrude until the next face **Outside** the solid. Hence the `upToFace = 1`
- If one of these conditions isn't met then we are fine with `upToFace = 0`
- And finally if ``upToFace isn't of type `int` then  it's of type `Face` and we pass the face to extrude to directly

There is indeed no check if `upToFace` is actually of type `Union[int, Face]`. Would you want to add a type check at the beginning and raise an error if that is not the case ? (And add a test associated with it)

I don't know if it's relevant since the allowed type are given in the method signature and it's a private method. A type check is already performed on the public method `extrude`

cadquery/cq.py
Jojain: Not really it's used later, but I'm going to removed the one in the `getFaceList` function as it's not needed and I can directly pass the `direction` var

cadquery/occ_impl/shapes.py
Jojain: The purpose of this is to be more general. It's not used in the PR but if someone wants to get a list of all the faces intersected by a LIne without consideration of the direction then it will be used.
That's why I didn't made this method private. I'm not sure it will be used that way one day though. I'm find with removing it and making the method private, you tell me

cadquery/cq.py
marcus7070: I didn't mean types. 

> if the wire is insinde the solid then we want to extrude until the next face Outside the solid. Hence the upToFace = 1

Then you probably want to set `upToFace = 1` *before* you do `limitFace = facesList[upToFace]`?

cadquery/occ_impl/shapes.py
marcus7070: I don't mind if the method is private or not, but `direction=None` doesn't have any tests. So I would say remove it or test it please.

Jojain:
@marcus7070 
I think I'm done, let me know if there is things I forgot/haven't responded !

marcus7070:
@Jojain can you have a look at a91f623 and 20691bc and make sure you understand and agree with what I've done there? Hopefully the tests make it obvious why I'm making those changes.

marcus7070 (approved):

Jojain:
Thanks @marcus7070 it's now obvious what you meant about limitFace = `facesList[upToFace]` and I had missed it.
That's all good for me. I'll try to bring another PR in the near future to add some doc

marcus7070:
Thank you very much @Jojain, this will be a great addition!

@jmwright, note that we've renamed the argument for `Workplane.extrude` from `distance` to `until` and `Workplane.cutBlind`'s `distanceToCut` also to `until`. I think that needs to be mentioned in the changelog.

PR #882, opened 2021-09-06
Perpendicular is not the same as not parallel

jdthorpe:
The current implementation of "#Z" is the same as "not |Z", which is not the same as perpendicular.


adam-urbanczyk:
Hm, the current implementation looks weird.

adam-urbanczyk (approved):

jmwright (approved):
Thanks @jdthorpe

adam-urbanczyk:
Thanks for the patience and your contribution @jdthorpe !

PR #892, opened 2021-09-27
Adds ability to define polygons with a circumscribed circle

jthacker:
This adds the option to define polygons by a circumscribed circle
in addition to the previous of inscribed circles. Inscribing remains
the default. Circumscribing is useful in certain circumstances
such as when the distance from the center of the polygon to the
midpoint of one of the segments needs to be defined, e.g. when
tiling the polygons along the x-axis.


jmwright:
Does it matter that the polygon orientation shifts between inscribed and circumscribed mode?

Screenshot from 2021-11-01 14-39-48

Screenshot from 2021-11-01 14-40-07

adam-urbanczyk:
Would be good to keep it consistent.

jthacker:
@adam-urbanczyk @jmwright The intention behind doing it this way was to match how the "radius" parameter is used. In both cases, it defines the polygon's extent along the x-axis. For inscribing, it is the standard radius, bisecting the angle of one corner of the polygon (the maximum shape extent). For circumscribing, it is the apothem, orthogonally intersecting the midpoint of one edge. The consistency that is maintained in the current implementation is that the radius or apothem is always drawn alongthe positive x-axis and defines the maximum extent of the shape along that axis. This has the advantage not requiring additional formulae for purposes such as tiling on the x-axis (at least for even sided polygons). An example of the simplicity gained for the tiling purpose:
image

If we go consistent orientation, the distance along the x or y axis' isn't known without using a conversion formula or rotating the polygon by a factor dependent on the number of sides. I can switch to consistent orientation and keeping the radius always along the x-axis if you prefer, I wanted to make sure you were aware of the intent behind this approach first though.

jmwright:
@jthacker Thanks for explaining. I'm ok with leaving it as-is so that the distance is known. The docstring could maybe make a note of the orientation difference and why. What do you think @adam-urbanczyk ?

adam-urbanczyk:
Sounds like a good rationale.

jthacker:
I've made the requested changes to the docstring. It does appear that the doc build step is failing due to a configuration error that appears present on other pull requests as well.

jmwright:
I think the config error is probably related to #906 . Can you pull the latest changes in master to see if it fixes it?

jthacker:
@jmwright Thanks for the pointer. I have updated the MR and all checks pass now.

adam-urbanczyk (approved):

jmwright (approved):
Thanks @jthacker !

PR #912, opened 2021-11-02
Update cq.Workplane.workplane doctring

Jojain:
As a user in the discord pointed it, the workplane method docs could be somehow not clear about what is the Z direction.

I removed the future enchancement that have been added since.
I also removed things that seemed odd and outdated to me feel free to revert the change if it was a mistake

### DIFF
--- a/cadquery/cq.py
+++ b/cadquery/cq.py
@@ -503,7 +503,7 @@ class Workplane(object):
         """
         Creates a new 2D workplane, located relative to the first face on the stack.
 
-        :param offset:  offset for the workplane in it's normal direction . Default
+        :param offset:  offset for the workplane in its normal direction . Default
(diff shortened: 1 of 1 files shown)


adam-urbanczyk (approved):

cadquery/cq.py
```diff
@@ -503,15 +503,15 @@ def workplane(
         """
         Creates a new 2D workplane, located relative to the first face on the stack.
 
-        :param offset:  offset for the work plane in the Z direction. Default
-        :param invert:  invert the Z direction from that of the face.
+        :param offset:  offset for the workplane in it's normal direction . Default
```
jmwright, line 506:
```suggestion
        :param offset:  offset for the workplane in its normal direction . Default
```

jmwright:
Thanks @Jojain !

Issue #819, opened 2021-07-08
docs HTML API reference duplicates Parameter lists


lorenzncode:
See the HTML reference for Workplane.lineTo as an example:

https://cadquery.readthedocs.io/en/latest/classreference.html#cadquery.Workplane.lineTo

![lineTo_current](https://user-images.githubusercontent.com/16394272/124858878-dfe1fb80-df7c-11eb-986a-18c48460f66d.png)

Notice there are two "Parameters:" lists where one list per method is expected.  Here is the portion of the lineTo docstring that documents parameters:

```
    :param float x: the x point, in workplane plane coordinates
    :param float y: the y point, in workplane plane coordinates
    :return: the Workplane object with the current point at the end of the new line
```

sphinx-build is currently configured to use the 3rd party extension "sphinx_autodoc_typehints".  
The docstring param definitions are expected to be of the form:

```
    :param x: the x point, in workplane plane coordinates
```

Modified docstring:

        """
        Make a line from the current point to the provided point

        :param x: the x point, in workplane plane coordinates
        :param y: the y point, in workplane plane coordinates
        :param forConstruction: whether or not the edge is used only for reference
        :return: the Workplane object with the current point at the end of the new line

        see :py:meth:`line` if you want to use relative dimensions to make a line instead.
        """

(I'm not sure if forConstruction should be documented or not - this is an example)
Re-running sphinx-build now yields:

![lineTo2](https://user-images.githubusercontent.com/16394272/124859178-68609c00-df7d-11eb-9a5b-1a69635c1168.png)

The docs are still corrupted and contain a second Parameters: list (single item self).  I find this **second list is created due to sphinx_autodoc_typehints** option set in conf.py:
**```always_document_param_types = True```**

Commenting this out or setting to False yields:

![lineTo3](https://user-images.githubusercontent.com/16394272/124859258-9514b380-df7d-11eb-871c-2efcdac466ac.png)

Many of the API reference methods have this issue with multiple Parameters lists. Not all of them (such as intersect) because in those cases there is no other text description in the docstring after the first parameter list.  The extension creates the stub for self, and other args and they become part of the first list.

One option to clean up the API reference is to disable ```always_document_param_types``` and continue using the sphinx_autodoc_typehints extension along with making docstring updates.

[...]

PR #916, opened 2021-11-06
Fix doc issue where parameters listed twice in description

lorenzncode:
Fix doc issue where parameters could be listed twice in the method description

* Remove sphinx extension sphinx_autodoc_typehints

* Update sphinx pin to version 4.2.0

* Show typehints in both signature and description


lorenzncode:
To resolve #819.

adam-urbanczyk:
Looks good @lorenzncode . One minor point - would it be easy to move self to the top?

tmp

lorenzncode:
Yes, it's easy to move self to the top.  I'll update the PR shortly.

lorenzncode:
I see one open issue to debug - the methods with Literal in the signature are rendered with strikethrough:

image

I also clobbered the existing setup() function that adds tables.css.

lorenzncode:
With docutils 0.16, the result is as expected:
image



Prior result with strikethrough text was using docutils  0.17.1.  HTML is ```<s><span class="pre">'XY'</span></s>```

Perhaps related to https://docutils.sourceforge.io/HISTORY.html

`Use HTML text-level tags <small>, <s>, <q>, <dfn>, <var>, <samp>, <kbd>, <i>, <b>, <u>, <mark>, and <bdi> if a unique, matching class value is found in inline and literal elements. Use <ins> and <del> if a unique matching class value is found in inline, literal, or container elements. Use <small> for generated code line numbers.`

lorenzncode:
The html (strikethrough) issue was reported by another user as sphinx-doc/sphinx#9872.  I will update this PR after sphinx 4.3.1 is released.

adam-urbanczyk (approved):

adam-urbanczyk:
Shall I merge or wait for 4.3.1 @lorenzncode ?

lorenzncode:
I'd propose to wait for sphinx 4.3.1.  I expect the release soon based on comments in the GH issues and the sphinx 4.3.1 release milestone tracker.  I'll watch for the release then update this PR to pin to sphinx 4.3.1.

lorenzncode:
Sphinx 4.3.1 is out!  The docs look good to me now (other than the known docstring issues mentioned in #922 to be fixed separately).

jmwright:
Looks like there's [a glibc error](https://dev.azure.com/cadquery/conda-packages/_build/results?buildId=2002&view=logs&j=469eeba9-e84c-576c-2174-efb2c0654ac2&t=a31fa24d-b17a-5ff4-3b81-3d95ffba7ac2&l=554) with Python 3.9 in Azure's Linux runs.

adam-urbanczyk:
Yup, will look into this

lorenzncode:
- [ ] Update sphinx pin to version 4.4.0
- [ ] Test with new _autodoc_typehints_format_ setting; compare 'fully-qualified' vs 'short'

I'd like to complete the tasks above before this is merged.

lorenzncode:
This could be a good time for this PR as it will resolve the current Read the Docs build failures (for example see PR #1018).

jmwright:
@lorenzncode Sounds good. It has two approvals, so I think it is ready to merge. Do you want to do the honors?

jmwright:
Thanks for fixing this @lorenzncode

Issue #909, opened 2021-10-30
Add precision on the 3 layers API of CQ


Jojain:
As it has been discussed on discord, it is often that people when they first learn cadquery dont understand the 3 layers API of CQ : 
1. Fluent API related to Workplane class
2. Topological API related to Shapes and its subclasses
3. OCCT API 

The docs needs to be update to explain more in depth these details and to provide users the tools to understand how to leverage the maximum of cadquery.

PR #917, opened 2021-11-06
added docs on the CQ API layers

Jojain:
As invoked with https://github.com/CadQuery/cadquery/issues/909

This PR aims to bring more documentation about the different CadQuery APIs and how one can take advantage of them.

This has been written with my own understanding prism so it is rather subjective.  I'm not sure these concepts have been properly named before so this is an attempt to clarify things for eveybody

### DIFF
--- a/doc/primer.rst
+++ b/doc/primer.rst
@@ -246,7 +246,7 @@ For example we can create a circular face like so ::
   In CadQuery (and OCCT) all the topological classes are shapes, the :class:`~cadquery.Shape` class is the most abstract topological class. 
   The topological class inherits :class:`~cadquery.Mixin3D` or :class:`~cadquery.Mixin1D` which provide aditional methods that are shared between the classes that inherits them.
 
-The direct API as its name suggest doesn't provide a parent/children data structure, instead each method call directly returns an object of the specified topological type.
+The direct API as its name suggests doesn't provide a parent/children data structure, instead each method call directly returns an object of the specified topological type.
 It is more verbose than the fluent API and more tedious to work with, but as it offer more flexibility (you can work with faces, which is something you can't do in the fluent API)
 it is sometimes more convenient than the fluent API.
 
--- a/doc/primer.rst
+++ b/doc/primer.rst
@@ -167,30 +167,29 @@ iterates on each member of the stack.
 
 This is really useful to remember  when you author your own plugins. :py:meth:`cadquery.cq.Workplane.each` is useful for this purpose.
 
-CadQuery different APIs
+CadQuery API layers
 ---------------------------
 
-Once you start to dive a bit more in the depth of CadQuery you may find yourself a bit confused between the different types of objects CadQuery is jungling with.
-This chapter aims to give an explanation on this topic and will give you the right knowledge to make you able to leverage all the CadQuery possibilities. 
+Once you start to dive a bit more into CadQuery, you may find yourself a bit confused juggling between different types of objects the CadQuery APIs can return.
+This chapter aims to give an explanation on this topic and to provide background on the underlying implementation and kernel layers so you can leverage more of CadQuery functionality.
 
 CadQuery is composed of 3 different API, which are implemented on top of each other.
 
-1. The fluent API 
-2. The direct API 
+1. The Fluent API 
+2. The Direct API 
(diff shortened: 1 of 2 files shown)


doc/primer.rst
```diff
@@ -248,19 +246,19 @@ For example we can create a circular face like so ::
   In CadQuery (and OCCT) all the topological classes are shapes, the :class:`~cadquery.Shape` class is the most abstract topological class. 
   The topological class inherits :class:`~cadquery.Mixin3D` or :class:`~cadquery.Mixin1D` which provide aditional methods that are shared between the classes that inherits them.
 
-The direct API as it's name suggest doesn't provide a parent/children data structure, each method call return directly an object of the specified topological type.
+The direct API as its name suggest doesn't provide a parent/children data structure, instead each method call directly returns an object of the specified topological type.
```
jmwright, line 249:
```suggestion
The direct API as its name suggests doesn't provide a parent/children data structure, instead each method call directly returns an object of the specified topological type.
```

doc/primer.rst
```diff
@@ -248,19 +246,19 @@ For example we can create a circular face like so ::
...
 It is more verbose than the fluent API and more tedious to work with, but as it offer more flexibility (you can work with faces, which is something you can't do in the fluent API)
 it is sometimes more convenient than the fluent API.
 
 The OCCT API
 ~~~~~~~~~~~~~~~~~~~~~~
 
-Finally we are discussing about the OCCT API. The OCCT API is the lowest level of CadQuery. The direct API is build upon the OCCT API, the OCCT API in CadQuery is available through OCP.
-OCP are the Python bindings of the OCCT C++ library CadQuery uses. This means you have access to all the OCCT C++ libraries in Python and in CadQuery.
+Finally we are discussing about the OCCT API. The OCCT API is the lowest level of CadQuery. The direct API is built upon the OCCT API, where the OCCT API in CadQuery is available through OCP.
+OCP are the Python bindings of the OCCT C++ libraries CadQuery uses. This means you have access to all the OCCT C++ libraries in Python and in CadQuery.
```
jmwright, line 257: I don't think that all the OCCT C++ classes/methods are wrapped.

jmwright (approved):
Thanks @Jojain

doc/primer.rst
adam-urbanczyk: Most of the classes...

Jojain: I thought they all were, maybe there is technical exceptions but since you have access to the classes of OCAF and 3D visualisation I assume it's safe to say it like that ?

marcus7070:
```suggestion
OCP are the Python bindings of the OCCT C++ libraries CadQuery uses. This means you have access to (almost) all the OCCT C++ libraries in Python and in CadQuery.
```

doc/primer.rst
```diff
@@ -167,6 +167,190 @@ iterates on each member of the stack.
...
+.. note::
+  The general way of importing a specific class of the OCCT API is ::
+
+    from OCP.thePackageName import theClassName
+
+  For example if you want to use the class `BRepPrimAPI_MakeBox <https://dev.opencascade.org/doc/refman/html/class_b_rep_prim_a_p_i___make_box.html>`_.
+  You will go by the following ::
+
+    from OCP.BRepPrimAPI import BRepPrimAPI_MakeBox
+
```
marcus7070, line 273: I think this is just general Python knowledge, not really specific to CQ or OCP.

```suggestion
```

marcus7070 (approved):

doc/primer.rst
adam-urbanczyk: What do you think about this suggestion @Jojain ?

Jojain: Well the intent here is to specifically talk about how to import binded c++ classes coming from OCP. I wrote this example because I struggled at first knowing how to import a specific class from OCP.
I my opinion this is much needed, since I couldn't find this information anywhere else.

adam-urbanczyk: OK, clear

adam-urbanczyk (approved):

adam-urbanczyk:
Thanks for the contribution @Jojain !

PR #918, opened 2021-11-07
use vtkmodules import instead of vtk

### LINKED ISSUE #915: vtk - import requires unnecessary dependencies
Currently, when cadquery imports vtk python modules it does something like:
`from vtk import vtkPolyData`

This actually causes vtk to load all vtk modules. In turn, this tries to load all of the shared libraries for every vtk module. If any of those are not present, it will cause cadquery to fail with an ImportError exception. For example:
```
>>> import cadquery
  File "<stdin>", line 1, in <module>
  File "/usr/lib/python3.9/site-packages/cadquery/__init__.py", line 3, in <module>
    from .occ_impl.shapes import (
  File "/usr/lib/python3.9/site-packages/cadquery/occ_impl/shapes.py", line 20, in <module>
    from vtk import (
  File "/usr/lib/python3.9/site-packages/vtk.py", line 30, in <module>
    all_m = importlib.import_module('vtkmodules.all')
  File "/usr/lib/python3.9/importlib/__init__.py", line 127, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "/usr/lib/python3.9/site-packages/vtkmodules/all.py", line 84, in <module>
    from .vtkIOODBC import *
ImportError: libodbc.so.2: cannot open shared object file: No such file or directory
```

Instead, we could use the newer "vtkmodules". For example:
`from vtkmodules.vtkCommonDataModel import vtkPolyData`

With this syntax, only the necessary vtk modules are loaded and everything succeeds even if you don't have the shared libraries for other (unused) vtk functionality.

Currently, AUR packages for Arch Linux are affected by this issue. Unless you manually install all the optional dependencies for vtk (which shouldn't be necessary), cadquery will not run.

darrellenns:
fixes #915

### DIFF
--- a/cadquery/occ_impl/assembly.py
+++ b/cadquery/occ_impl/assembly.py
@@ -10,7 +10,11 @@ from OCP.TDF import TDF_Label
 from OCP.TopLoc import TopLoc_Location
 from OCP.Quantity import Quantity_ColorRGBA
 
-from vtk import vtkActor, vtkPolyDataMapper as vtkMapper, vtkRenderer
+from vtkmodules.vtkRenderingCore import (
+    vtkActor,
+    vtkPolyDataMapper as vtkMapper,
+    vtkRenderer,
+)
 
 from .geom import Location
 from .shapes import Shape, Compound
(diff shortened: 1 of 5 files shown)


jmwright (approved):
Thanks @darrellenns

adam-urbanczyk (approved):

PR #923, opened 2021-11-24
Handle kwargs+multimethods in a backward compatibile way

adam-urbanczyk:


jmwright (approved):
Thanks @adam-urbanczyk !

marcus7070:
@adam-urbanczyk take my last commit as a suggestion, do you approve?

Since specifying positional arguments as keyword arguments is supported by vanilla python, I do think we need to make it explicit to the users that it might cause problems.

marcus7070 (approved):

Issue #926, opened 2021-12-01
Initialize assy metadata to {}


adam-urbanczyk:
And add a kwarg to the constructor.

adam-urbanczyk:
Closed by #928

PR #928, opened 2021-12-03
#926 Initialize assy metadata

gumyr:
With this change `metadata` is a full instance variable of the `Assembly` class with a default of `None`. The `_copy()` method was changed to copy metadata ensuring that metadata added in a sub-assembly is maintained.

A new test verifies that the metadata is present and editable in both the base and sub-assembly.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -18,7 +18,7 @@ jobs:
 - template: conda-build.yml@templates
   parameters:
     name: Linux_38
-    vmImage: 'ubuntu-20.04'
+    vmImage: 'ubuntu-18.04'
     py_maj: 3
     py_min: 8
     conda_bld: 3.21.6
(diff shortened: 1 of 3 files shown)


gumyr:
As implemented here the initialization is to None as this is slightly more memory efficient than an empty dict {}. However, this makes working with metadata a little more complex as elements can't be added until it's created which requires an if / then. Adam probably anticipated this - what do you think of initialization to {}?

gumyr:
Does anyone know what the problem is here? The code is working in my build.

jmwright:
> Does anyone know what the problem is here? The code is working in my build.

The glibc error in the Linux builds is effecting another PR and will be addressed. You can ignore the codecov failure. It is over-sensitive sometimes.

adam-urbanczyk:
Not related to this PR @gumyr (which BTW looks good). I'm trying to debug this (CI) issue in #926.

adam-urbanczyk (approved):

jmwright (approved):
Thanks @gumyr !

adam-urbanczyk:
Merging, thanks for the contribution @gumyr !

jmwright:
All green now with the 18.04 fix and it has been approved, so I'm going to merge.

PR #952, opened 2022-01-06
add __repr__ to plane

### LINKED ISSUE #739: Better __repr__ on Plane
From [the Python docs](https://docs.python.org/3/reference/datamodel.html#object.__repr__):
> object.__repr__(self)
> ... If at all possible, this should look like a valid Python expression that could be used to recreate an object with the same value (given an appropriate environment).

Currently:
```
>>> repr(cq.Plane((1, 2, 3), (1, 0, 0), (0, 0, 1)))
'<cadquery.occ_impl.geom.Plane object at 0x7f91265ff9d0>'
```

I would suggest `Plane(origin=(1, 2, 3), xDir=(1, 0, 0), normal=(0, 0, 1))`.

Jojain:
Resolve #739, didn't added test it as it is really trivial

### DIFF
--- a/cadquery/occ_impl/geom.py
+++ b/cadquery/occ_impl/geom.py
@@ -584,7 +584,6 @@ class Plane(object):
 
     def __repr__(self):
         return f"Plane(origin={str(self.origin.toTuple())}, xDir={str(self.xDir.toTuple())}, normal={str(self.zDir.toTuple())})"
-         
 
     @property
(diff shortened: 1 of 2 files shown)


jmwright (approved):
Looks good to me, thanks @Jojain

adam-urbanczyk (approved):

adam-urbanczyk:
LGTM, thanks @Jojain !

Issue #832, opened 2021-07-21
Adding revolveCut and sweepCut


Jojain:
Hello,

As far as I know currently we can't use `revolve` and `sweep` to remove material with the fluent API.
I think it would be nice to have it, and to have in general for each matter creation operation the opposite matter removal operation.

This could be implemented either as a param in the discussed methods as `additive = False` for example.
Or to implement additionals methods as it has been done with `extrude` and `cutBlind`, let's say `revolveCut` and `sweepCut`.

PS : Personal opinion here : I think it would be nice to have coherent way to have material addition/removal operations. So either having a standard method with a param to remove material or 2 methods but not mix these two ways in order to keep consistency.

adam-urbanczyk:
There is already an issue to track this: https://github.com/CadQuery/cadquery/issues/30. The proposal would be to extend combine with other options.

Jojain:
Sorry, I'm closing this and continuing discussion on #30

PR #954, opened 2022-01-07
added combine="cut" option for 3D operations

Jojain:
Aims to target #30 I added the possibility to use combine="cut" for 3D operations to remove material from the context solid.
It's essentially a shorcut of calling `cut()` later.

I didn't added it for `extrude` as it has already `cutBlind` that does the same thing, I didn't added it to `text` since it has a `cut` kwarg. I think it should be changed to be consistent with all the other methods but since it will probably not be agreed on I didn't changed it.

As I already said in #832, I still think everything should be made consistent and that we should remove/deprecate things that aren't consistent. I know it would require breaking change but not doing it will likely confuse more and more new users.


Jojain:
I've added test to target the lines marked as non tested but it keeps failing

cadquery/cq.py
```diff
@@ -2968,7 +2968,7 @@ def twistExtrude(
 
         :param distance: the distance to extrude normal to the workplane
         :param angle: angle (in degrees) to rotate through the extrusion
-        :param boolean combine: True to combine the resulting solid with parent solids if found.
+        :param boolean or string combine: True to combine the resulting solid with parent solids if found, "cut" to remove the resulting solid with the parent solids if found.
```
jmwright, line 2971:
```suggestion
        :param boolean or string combine: True to combine the resulting solid with parent solids if found, "cut" to remove the resulting solid from the parent solids if found.
```

cadquery/cq.py
```diff
@@ -3155,7 +3159,7 @@ def sweep(
 
         :param path: A wire along which the pending wires will be swept
         :param boolean multiSection: False to create multiple swept from wires on the chain along path. True to create only one solid swept along path with shape following the list of wires on the chain
-        :param boolean combine: True to combine the resulting solid with parent solids if found.
+        :param boolean or string combine: True to combine the resulting solid with parent solids if found, "cut" to remove the resulting solid with the parent solids if found.
```
jmwright, line 3162:
```suggestion
        :param boolean or string combine: True to combine the resulting solid with parent solids if found, "cut" to remove the resulting solid from the parent solids if found.
```

cadquery/cq.py
```diff
@@ -2993,7 +2993,9 @@ def twistExtrude(
 
         r = Compound.makeCompound(shapes).fuse()
 
-        if combine:
+        if isinstance(combine, str) and combine == "cut":
```
jmwright, line 2996: It seems like this section of code ends up being repeated quite a bit, but I'm not sure that there's an quick/efficient way to avoid the code duplication. Probably best just to leave it for now unless someone has an idea.

jmwright (approved):
Thanks @Jojain !

Jojain:
> > I've added test to target the lines marked as non tested but it keeps failing
> 
> I think it's fine. We have to sometimes ignore codecov when it complains about minor things.

Well ok, just let me know if there is something I missed, I'm not really used to add tests so I'm not always sure if it's done right or not

cadquery/cq.py
Jojain: I'm not sure to understand what we are talking about. The plan would be to move what to `_combineWithBase` ?

adam-urbanczyk: There is a lot of repetitive code like:
```python
        if isinstance(combine, str) and combine == "cut":
            newS = self._cutFromBase(r)
        elif isinstance(combine, bool) and combine:
            newS = self._combineWithBase(r)
        else:
            newS = self.newObject([r])
```

it could be refactored to:

```python
return self._combineWithBase(r, combine)
```

(ofc you'd need to update `_combineWithBase` accordingly).

Jojain: Ok I think I've refactored `_combineWithBase` like you thought about it.
I'm proposing to add a breaking change to `text` that has both `cut` and `combine` kwarg. In my opinion we should get rid of the `cut` kwarg and keep the `combine` consistent with other methods. It would be fairly easy for user to make the change but would result in errors in model that still use it

Let me know if there is an agreement on this or not so I can make the change (or not) @adam-urbanczyk @jmwright @marcus7070

adam-urbanczyk: Long term it is fine, but please use this for now: https://github.com/CadQuery/cadquery/blob/ea66ded172c2d6458a1209723546fb8778aa3060/cadquery/utils.py#L8.

Jojain:
As discussed on discord I'm not sure it make sense to add the option for `each` for me it would result in wrong use of the method.

I've not added it to `extrude` since there is already `cutBlind` and the `extrude` method is already very complex, it would however make sense to still add it (but requires a bit more work) and deprecate `cutBlind`, what are your thoughts about it ?

Finally github says there is conflicts but there isnt, I cannot solve it so a lead dev will have to resolve the so called conflicts

jmwright:
> I've not added it to extrude since there is already cutBlind and the extrude method is already very complex, it would however make sense to still add it (but requires a bit more work) and deprecate cutBlind, what are your thoughts about it ?

My vote would probably be to keep `cutBlind()`, but re-implement it to use the new combine/cut functionality. That way we don't have to break existing scripts, but we're leveraging the latest best practices within the codebase. Thoughts?

Jojain:
> My vote would probably be to keep cutBlind(), but re-implement it to use the new combine/cut functionality. That way we don't have to break existing scripts, but we're leveraging the latest best practices within the codebase. Thoughts?

Well I don't think it makes sense to have combine/cut with `cutBlind` since what it does is explicitly in the name, we can maybe make `extrude` call `cutBlind` if it set with `combine="cut"` and keep it, just telling user that best practice are to use `extrude` with `combine="cut"`  instead of `cutBlind`. That a good compromise IMHO

adam-urbanczyk:
> 

To make everything self-consistent (make the code shorter and get rid of special cases) it would make sense to add the option to every method that allows to use combine. For now forget about `cutBlind` etc.

Jojain:
Hopefully this is done.

In the end I didn't added combine option to `each` because I don't think it fits the method well, and actually having it only in `eachpoint` may lead users to better choose which method they actually need for what they want to do. 

I still don't know about the merging issues I let you check that, all test are green on my side.

fedorkotov:
Please don't forget to update `changes.md` file. It's very helpful when trying to get up to speed with CadQuery after a long pause.

cadquery/cq.py
```diff
@@ -3181,18 +3176,45 @@ def sweep(
...
+            # dispatch on the mode
+            if mode in ("cut", "s"):
+                newS = self._cutFromBase(obj)
+            elif mode in (True, "a"):
+                newS = self._fuseWithBase(obj)
+
         else:
-            newS = self.newObject([r])
+            # do not combine branch
+            newS = self.newObject(obj if not isinstance(obj, Shape) else [obj])
```
adam-urbanczyk, line 3208: I'm not sure that that is the optimal way. We might want to rewrite the failing test

cadquery/cq.py
```diff
@@ -61,6 +62,7 @@
 
 CQObject = Union[Vector, Location, Shape, Sketch]
 VectorLike = Union[Tuple[float, float], Tuple[float, float, float], Vector]
+CombineMode = Union[bool, Literal["cut", "a", "s"]]  # a : additive, s: subtractive
```
adam-urbanczyk, line 65: I added also modes used in Sketch. Note that it is better to use Literal for typing.

adam-urbanczyk (approved):

adam-urbanczyk:
Thanks for the contribution @Jojain ! I did some rework, let me know what you think.

cadquery/cq.py
```diff
@@ -3041,46 +3036,40 @@ def extrude(
...
-            # Handle `until` multiple values
-            if isinstance(until, str) and until in ("next", "last") and combine:
-                if until == "next":
-                    faceIndex = 0
-                elif until == "last":
-                    faceIndex = -1
 
-                r = self._extrude(None, both=both, taper=taper, upToFace=faceIndex)
+        # Handle `until` multiple values
+        if isinstance(until, str) and until in ("next", "last") and combine:
```
Jojain, line 3041: If combine is  "cut" the conditional test will be True but the executed code won't be what's expected. 
That's why I added an explicit condition to check if `combine="cut"` and if this True we call the `cutBlind` method.
I think we should keep my first test

cadquery/cq.py
```diff
@@ -3189,28 +3176,39 @@ def sweep(
...
+        self: T,
+        obj: Union[Shape, Iterable[Shape]],
+        mode: CombineMode = True,
+        clean: bool = False,
     ) -> T:
         """
         Combines the provided object with the base solid, if one can be found.
         :param obj: The object to be combined with the context solid
-        :param combine_mode: The mode to combine with the base solid (True, False or "cut")
+        :param mode: The mode to combine with the base solid (True, False or "cut")
```
Jojain, line 3190: We should maybe document the fact that we can pass `cut="a"` and `cut="s"` here and in all the method that use combine?

cadquery/cq.py
```diff
@@ -2392,7 +2393,7 @@ def each(
         self: T,
         callback: Callable[[CQObject], Shape],
         useLocalCoordinates: bool = False,
-        combine: Union[bool, str] = True,
+        combine: CombineMode = True,
```
Jojain, line 2396: I already discussed my opinion on adding `combine` arg for this method so I won't talk about it again but in any case if we decide to add it we should probably write somewhere that the new default behaviour is to `combine` with the base which wasnt the case before. Same remark goes for `eachpoint`

Jojain (commented):
My comment aside I'm good with what you came up with. I think the code simplification is great.
It would be useful to document the change in the docs as they introduce a "new concept" which is that every protusion method can also be used to remove material.
I think we should lead newcomers in the docs to use the `combine` argument so everything would feel more smooth.

cadquery/cq.py
adam-urbanczyk: Good point. Do you think it would make sense to set it to False? I'm kind of hesitating about it.

Jojain: Setting it to False would make things transparent but we would loose consistency with other methods. 
I don't think each and eachpoint are that used so a breaking change here would probably not be a big deal.

cadquery/cq.py
adam-urbanczyk: Thanks, missed that. I think it is now fixed.

cadquery/cq.py
adam-urbanczyk: OK, I added a note in changes.md

adam-urbanczyk:
@jmwright @Jojain I'm done with my adjustments. Please take a final look.

jmwright (approved):

jmwright:
@Jojain This looks good to me. Do you think it is ready to merge?

Jojain:
@jmwright  @adam-urbanczyk  It's all good for me

jmwright:
Thanks!

Issue #108, opened 2019-02-22
Text placement on a 2D curve


adam-urbanczyk:
Support text on a 2D curve

Continuation of #48

gumyr:
Is this the basics of what is intended by this feature?
```python
from math import degrees, atan2
import cadquery as cq


def _faceOnWire(self, path: cq.Wire) -> cq.Face:
    """Reposition a face from alignment to the x-axis to the provided path"""
    path_length = path.Length()

    bbox = self.BoundingBox()
    face_bottom_center = cq.Vector((bbox.xmin + bbox.xmax) / 2, 0, 0)
    relative_position_on_wire = face_bottom_center.x / path_length
    wire_tangent = path.tangentAt(relative_position_on_wire)
    wire_angle = degrees(atan2(wire_tangent.y, wire_tangent.x))
    wire_position = path.positionAt(relative_position_on_wire)

    return self.rotate(
        face_bottom_center, face_bottom_center + cq.Vector(0, 0, 1), wire_angle
    ).translate(wire_position - face_bottom_center)


cq.Face.faceOnWire = _faceOnWire


def textOnWire(txt: str, fontsize: float, distance: float, path: cq.Wire) -> cq.Solid:
    """Create 3D text with a baseline following the given path"""
    linear_faces = (
        cq.Workplane("XY")
        .text(
            txt=txt,
            fontsize=fontsize,
            distance=distance,
            halign="left",
            valign="bottom",
        )
        .faces("<Z")
        .vals()
    )
    faces_on_path = [f.faceOnWire(path) for f in linear_faces]
    return cq.Compound.makeCompound(
        [cq.Solid.extrudeLinear(f, cq.Vector(0, 0, 1)) for f in faces_on_path]
    )


path = cq.Edge.makeThreePointArc(
    cq.Vector(0, 0, 0), cq.Vector(50, 30, 0), cq.Vector(100, 0, 0)
)
text_on_path = textOnWire(
    txt="The quick brown fox jumped over the lazy dog",
    fontsize=5,
    distance=1,
    path=path,
)
if "show_object" in locals():
    show_object(text_on_path, name="text_on_path")
    show_object(path, name="path")
```
which generates this:
![image](https://user-images.githubusercontent.com/85638518/148412102-dbfc2e59-17e1-404a-bae8-1fddfb60e5b6.png)

If so, I don't see how this is related to #109

adam-urbanczyk:
[...]

PR #955, opened 2022-01-10
 Text placement on a 2D curve #108 

gumyr:
This submission enables the creation of 3D text on a predefined Edge or Wire, e.g.:
```python
fox = (
    cq.Workplane("XZ")
    .threePointArc((50, 30), (100, 0))
    .textOnPath(
        txt="The quick brown fox jumped over the lazy dog",
        fontsize=5,
        distance=1,
        start=0.1,
    )
)
```
The implementation follows the general design in Inkscape where the bottom center of each character is aligned to the path and appears as follows:
image
Any plane can be used to create the text - testing was done with a random plane normal. A float `start` value allows the user to "slide" the text along the path to get the alignment they like (even sliding off the end of the path).

As is, the code "prefers" an Edge over a Wire:
```python
    if self.ctx.pendingEdges:
        path = self.ctx.pendingEdges.pop()
    else:
        path = self.ctx.pendingWires.pop()
```
If there is a way to get the last Edge or Wire put on the stack, I think that would be an improvement.

### DIFF
--- a/cadquery/cq.py
+++ b/cadquery/cq.py
@@ -52,10 +52,7 @@ from .occ_impl.exporters.svg import getSVG, exportSVG
 
 from .utils import deprecate
 
-from .selectors import (
-    Selector,
-    StringSyntaxSelector,
-)
+from .selectors import Selector, StringSyntaxSelector, DirectionMinMaxSelector
 
 from .sketch import Sketch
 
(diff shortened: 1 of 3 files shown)


adam-urbanczyk:
I'm afraid I'll insist on implementing this in the occ_impl layer.

gumyr:
Fair enough.  I hadn't really planned on doing this feature at all so if you don't want this version I'll just release it in cq_warehouse. Going forward, if you have absolute requirements like this it would be helpful to put them into the issue.

I was surprised by the black formatting issues.  At least one of the formatting changes was the removal of trailing white space - could it be that the file was never formatted with black?

PR #962, opened 2022-01-18
Fix arc length cost

adam-urbanczyk:


jmwright:
@lorenzncode I requested your review of this PR since you have been closely involved. You can decline if you would rather not.

jmwright (approved):
Thanks @adam-urbanczyk

adam-urbanczyk:
Yes, please do have a look @lorenzncode . The other issue/PR might be more complicated.

lorenzncode (approved):
I checked out the branch and ran a few manual tests.  I did not find any problem with arc length cost.


There may be an issue related to bounds.

```
s = (
    cq.Sketch()
    .segment((0.3, 0), (0, 0), "seg1")
    .arc((0.0, 0.0), (-0.25, 0.5), (0.0, 1.0), "arc1")
    .constrain("seg1", "Fixed", None)
    .constrain("seg1", "arc1", "Coincident", None)
    .constrain("arc1", "Length", 60)
    .solve()
)
```

The above which omits Radius and ArcAngle constraints, and with relatively large Length constraint value returns an "entity" with arc angle > 2*pi.

```
	[ 5.01675398  6.6817405   8.35544592 -2.49756986  7.18094529] (CIRCLE)

```

The resulting Edge length is ~45 when expected value is 60.

I tried setting an upper bound of 2*pi to arc angle (where currently only the lower bound for radius is set).  The result was an Edge with length 60 and entity:

```
	[16.70179153 19.90442625 25.9833798  -2.44346111  2.30916842] (CIRCLE)

```

I think the bounds issue can be handled separately from this PR.  Looking forward to see how the sketch solver progresses!

adam-urbanczyk:
@lorenzncode I tried adding `ub` and `lb` but I get failures for some test cases. I think this will be a longer story (different solver, different parametrization, ...). In the meantime I'd propose to merge this to fix the obvious bug. @jmwright if you agree, feel free to go ahead with merging.

Issue #959, opened 2022-01-15
Sketch Angle, ArcAngle constraints are reversed


lorenzncode:
I find ArcAngle results are in the reversed sense.

For example, the results of the following Sketch with or without solve is expected to be the same.

```
s = (
    cq.Sketch()
    .arc((0,0), 1, 30, 120, "arc1")
    .constrain("arc1", "FixedPoint", 0)
    .constrain("arc1", "ArcAngle", 120)
    #.solve()
)
```
Without solve:
![image](https://user-images.githubusercontent.com/16394272/149629459-3cb31f9f-3a30-4b91-8335-8ad088cbbe10.png)

With solve (not expected):
![image](https://user-images.githubusercontent.com/16394272/149629470-ef1d1f60-6332-4a8f-85f0-a36c678528d9.png)

Angle constraint is likewise reversed.  I'll plan to open a PR shortly.

adam-urbanczyk:
Thanks for checking! Constraint itself looks OK, but It might be the translation from the solver DOFs to cq.Edge.

PR #964, opened 2022-01-18
Sketch solver update

lorenzncode:
* Arc length fix
* Angle fix

I see #962 already addresses the arc length bug.  I hope this PR is helpful regarding #959.

### DIFF
--- a/cadquery/occ_impl/sketch_solver.py
+++ b/cadquery/occ_impl/sketch_solver.py
@@ -62,12 +62,12 @@ def invalid_args(*t):
 
 def arc_first(x):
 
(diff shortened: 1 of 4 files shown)


jmwright (approved):
Thanks @lorenzncode

adam-urbanczyk:
Thanks @lorenzncode , this ended up in limbo I guess. I'm not sure if I agree with the direction - better change the convention when creating the arc or when passing the param. Additionally, I'm currently thinking about rewriting the constraint solver from scratch based on the issues we have seen.

lorenzncode:
A larger rewrite sounds good.  Closing.

I'd be in favor of keeping the Sketch arc rotation convention as it is currently because that seems to be the common convention across `Sketch`, `Workplane`, `Assembly`.

PR #971, opened 2022-01-26
simple solution test

Jojain:
Keep selected objects after workplane creation, it breaks quite a lof of tests, we need to investigate if the failing test can be rewritten or if it's really a breaking change

### DIFF
--- a/cadquery/cq.py
+++ b/cadquery/cq.py
@@ -667,6 +667,7 @@ class Workplane(object):
         plane = Plane(offsetCenter, xDir, normal)
         s = self.__class__(plane)
         s.parent = self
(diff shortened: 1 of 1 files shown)


adam-urbanczyk:
Thanks, it might be trickier in terms of backward-compatibility than I expected.

adam-urbanczyk:
I think it'd be better to add a kwarg to work plane (say `reset:bool = False`) to not mess with breaking changes.

Jojain:
Yeah that's probably easier to deal with. Do we want to reset the selection totally calling `findSolid` or we just put the last selection state on the stack?

PR #975, opened 2022-02-01
Additional constraints

adam-urbanczyk:
- [x] simplification of the assy solver
- [x] validation
- [x] PointOnLine
- [x] FixedPoint
- [x] FixedAxis
- [x] FixedRotation
- [x] docs


adam-urbanczyk:
I did not add the new constraints yet, but @marcus7070 @lorenzncode could you take a look to see if the reworked version is more readable?

cadquery/occ_impl/solver.py
```diff
@@ -20,36 +56,243 @@
...
+            gp_Dir: self._getAxis,
+            gp_Pln: self._getPln,
+        }
+
+        for a, t in zip(args, tcast(Tuple[Type[ConstraintMarker], ...], marker_types)):
+            arg_check[t](a)
+            try:
+                arg_check[t](a)
+            except ValueError:
+                raise ValueError(f"Unsupported entitiy {a} for constraint {kind}.")
```
marcus7070, line 128:
```suggestion
                raise ValueError(f"Unsupported entity {a} for constraint {kind}.")
```

cadquery/occ_impl/solver.py
```diff
@@ -1,17 +1,53 @@
...
+    "Point": (2, (gp_Pnt, gp_Pnt), Real, None),
+    "Axis": (2, (gp_Dir, gp_Dir), Real, radians),
+    "PointInPlane": (2, (gp_Pnt, gp_Pln), Real, radians),
+    "FixedPoint": (1, (gp_Pnt,), Tuple[Real, Real, Real], None),
+}
+
+# translation table for compound constraints {name : (name, ...)}
+CompoundConstraints: Dict[ConstraintKind, Tuple[ConstraintKind, ...]] = {
+    "Plane": ("Axis", "Point")
+}
```
marcus7070, line 47: Compound constraints sound great. Efficient re-use of code and abstraction.

cadquery/occ_impl/solver.py
```diff
@@ -20,36 +56,243 @@
...
+        # specify kinds of the simple constraint
+        if self.kind in CompoundConstraints:
+            kinds = CompoundConstraints[self.kind]
+            params = (self.param,) * len(kinds)
+        else:
+            kinds = (self.kind,)
+            params = (self.param,)
+
+        # builds the tuple and return
+        return tuple((m, k, p) for m, k, p in zip(markers, kinds, params))
```
marcus7070, line 225: Absolutely more readable now, thanks @adam-urbanczyk.

tests/test_assembly.py
```diff
@@ -445,9 +445,8 @@ def make_constraint(shape0):
         return cq.Constraint(ids, (shape0, shape0), sublocs, "PointInPlane", 0)
 
     def fail_this(shape0):
-        c0 = make_constraint(shape0)
         with pytest.raises(ValueError):
-            c0._getPln(c0.args[0])
+            make_constraint(shape0)
```
marcus7070, line 449: That's much neater now, thanks.

cadquery/occ_impl/solver.py
```diff
@@ -20,36 +62,262 @@
...
+        elif self.kind == "FixedAxis":
+            markers = [(self._getAxis(args[0]),)]
+
+        else:
+            raise ValueError(f"Unknown constraint kind {self.kind}")
+
+        # specify kinds of the simple constraint
+        if self.kind in CompoundConstraints:
+            kinds = CompoundConstraints[self.kind]
+            params = (self.param,) * len(kinds)
```
lorenzncode, line 232: Readability does look to be much improved.  I'm still getting familiar with the assy solver.

Regarding `CompoundConstraints` (Plane), here the same `param` value is passed to both the Axis and Point simple constraints.

Sharing the same `param` value across the simple constraints (at least for Plane) didn't make sense to me.  It could be I'm missing something.  I thought I'd explore passing two `param` values to `CompoundConstraints` (3.) or for Plane, fix the Point `param` and allow the Axis `param` angle to vary (2.). 


1. Single `param` value passed to all simple constraints

   This is the current implementation.  Say I specify a Plane constraint with `param=pi/2`.

   This solves for Axis angle of pi/2, and Point **distance** pi/2.  It doesn't make sense to me because the Axis param value is an angle and the Point param is a distance.

   If a single `param` is passed to both Axis, and Point, then any value other than 0, or None (both already treated as 0 for Point param) probably should be disallowed.

   When I try Plane with value other than 0, or None, I get very weird results because the solver optimizes for Point distance set to the angle in radians. 


2. Pass `param` to Axis, fix Point `param` to 0

   For more flexibility of the Plane constraint, perhaps allow for passing angle param value of 0 to pi.  The Point param value would always be fixed to zero. 

   `params = (self.param, 0)`

   Plane could then be used for any combined Axis, Point case unless non-zero Point param was required.


3. Pass two separate `param` values to the individual simple constraints

   Perhaps some backwards compatible way could be designed to pass a tuple of param values to Plane.

   Say a Plane constraint is specified with `param=(pi,100)`.

   This might offer the most flexibility for Plane without requiring to enter two separate Axis and Point constraints.

adam-urbanczyk: Good point, I think I'd go for 3.

adam-urbanczyk: After some thinking I implemented 2 actually.

cadquery/occ_impl/solver.py
```diff
@@ -1,17 +1,99 @@
...
+        Tuple[int, Real],
+        lambda x: (x[0], radians(x[1])),
+    ),
+}
+
+# translation table for compound constraints {name : (name, ...), converter}
+CompoundConstraints: Dict[
+    ConstraintKind, Tuple[Tuple[ConstraintKind, ...], Callable[[Any], Tuple[Any, ...]]]
+] = {
+    "Plane": (("Axis", "Point"), lambda x: (x, 0)),
```
lorenzncode, line 87:
```suggestion
    "Plane": (("Axis", "Point"), lambda x: (radians(x) if x is not None else None, 0)),
```
This is to apply the conversion func to Axis param when part of the Plane compound constraint.  Will look more tomorrow!

doc/assy.rst
```diff
@@ -710,6 +710,176 @@ Where:
...
+    # effectively b2 will be constrained to be on the intersection of the two edges
+
+    assy.solve()
+    show_object(assy)
+
+
+FixedRotation
+=============
+
+FixedRotation fixes the rotation of the given argument to be equal to the given point specified via the parameter of the constraint. This constraint locks all rotational degrees of freedom of the argument.
```
marcus7070, line 805:
```suggestion
FixedRotation fixes the rotation of the given argument to be equal to the value specified via the parameter of the constraint. The argument is rotated about it's origin firstly by the Z angle, then Y and finally X.

This constraint locks all rotational degrees of freedom of the argument.
```
At least I think that's correct, please do check closely @adam-urbanczyk.

cadquery/occ_impl/solver.py
```diff
@@ -20,36 +102,317 @@
...
+        # convert to marker objects
+        if self.kind == "Axis":
+            markers = [(self._getAxis(args[0]), self._getAxis(args[1]),)]
+
+        elif self.kind == "Point":
+            markers = [(self._getPnt(args[0]), self._getPnt(args[1]))]
+
+        elif self.kind == "FixedPoint":
+            markers = [(self._getPnt(args[0]),)]
+
```
marcus7070, line 272: You've got `FixedPoint` in here twice.
```suggestion
```

cadquery/occ_impl/solver.py
```diff
@@ -20,36 +102,317 @@
...
+            markers = [(self._getAxis(args[0]),)]
+
+        elif self.kind == "FixedRotation":
+            markers = [(None,), (None,), (None,)]
+
+        elif self.kind == "FixedRotationAxis":
+            markers = [(None,)]
+
+        else:
+            raise ValueError(f"Unknown constraint kind {self.kind}")
```
marcus7070, line 301: Just incase this idea didn't occur to you when you were writing this, I think you could replace all the `if ... elif` with a dict containing lambdas, like:
```python
marker_lookup = {
    ...
    "FixedAxis": lambda s, a: [(s._getAxis(a[0]),)],
    ...
}
try:
    markers = marker_dict[self.kind](self, args)
except KeyError:
    raise ValueError(f"Unknown constraint kind {self.kind}")
```
But I can't figure out how to get that past mypy, so I'll just leave the suggestion with you.

marcus7070 (approved):
This looks fantastic @adam-urbanczyk, thanks so much for doing it!

cadquery/assembly.py
```diff
@@ -434,32 +353,40 @@ def solve(self) -> "Assembly":
...
                     ents[name] = i
-                    if name == self.name:
-                        lock_ix = i
+                    if c.kind == "Fixed" or name == self.name:
+                        locked.append(i)
                     i += 1
 
+        # Lock the first occuring entity if needed.
+        if not locked:
+            locked.append(0)
```
lorenzncode, line 371:
```suggestion
            unary_objects = [c.objects[0] for c in self.constraints if instance_of(c.kind, UnaryConstraintKind)]
            binary_objects = [c.objects[0] for c in self.constraints if instance_of(c.kind, BinaryConstraintKind)]
            for b in binary_objects: 
                if b not in unary_objects:
                    locked.append(ents[b])
                    break
```
When there are no explicit "Fixed" constraints, the current logic will lock the first occurring entity.  

Now with the new unary constraints this can be a problem if the first entity is part of a unary constraint.  I think it's important to avoid locking an entity that is part of a unary constraint as it will guarantee non-convergence of the optimizer. 

Something like the above changes might be implemented to lock only the first object of a binary constraint that is also not part of any unary constraint.

lorenzncode (changes requested):

cadquery/occ_impl/solver.py
adam-urbanczyk: Yes, it'll involve some funky typing... Let's leave as-is

adam-urbanczyk:
Ok, I think that this is it @jmwright @lorenzncode @marcus7070 .  The only thin I don't really like is the automatic fixed constraint handling. Maybe in the future (not in this PR) we should require an explicit `Fixed` constraint being specified.

cadquery/occ_impl/solver.py
```diff
@@ -20,36 +102,314 @@
...
+
+        if isinstance(arg, Face):
+            rv = gp_Pln(self._getPnt(arg), arg.normalAt().toDir())
+        elif isinstance(arg, (Edge, Wire)):
+            normal = arg.normal()
+            origin = arg.Center()
+            plane = Plane(origin, normal=normal)
+            rv = plane.toPln()
+        else:
+            raise ValueError(f"Can not construct a plane for {arg}.")
```
lorenzncode, line 220:
```suggestion
            raise ValueError(f"Cannot construct a plane for {arg}.")
```

cadquery/occ_impl/solver.py
```diff
@@ -20,36 +102,314 @@
...
+
+        return center.toPnt()
+
+    def _getLin(self, arg: Shape) -> gp_Lin:
+
+        if isinstance(arg, (Edge, Wire)):
+            center = arg.Center()
+            tangent = arg.tangentAt()
+        else:
+            raise ValueError(f"Can not construct a plane for {arg}.")
```
lorenzncode, line 244:
```suggestion
            raise ValueError(f"Cannot construct a plane for {arg}.")
```

doc/assy.rst
```diff
@@ -643,7 +643,7 @@ The result of this code is identical to the above two constraint example.
 For the cost function of Plane, please see the Point and Axis sections. The ``param`` argument is
 copied into both constraints and should be left as the default value of ``None`` for a "mate" style
```
lorenzncode, line 644:
```suggestion
For the cost function of Plane, please see the Point and Axis sections. The ``param`` argument is applied to Axis and should be left as the default value for a "mate" style
```

lorenzncode (changes requested):
I've added a few minor suggestions for docs, error messages.

doc/assy.rst
```diff
@@ -710,6 +710,178 @@ Where:
...
+    assy.constrain('b2','FixedRotation',(45,0,45))
+
+    assy.solve()
+    show_object(assy)
+
+
+FixedAxis
+=========
+
+FixedAxis fixes the orientation of the given argument's normal or tangent to be equal to the orientation  of the vector specified via the parameter of the constraint. This constraint locks two rotational degrees of freedom of the argument.
```
jmwright, line 848:
```suggestion
FixedAxis fixes the orientation of the given argument's normal or tangent to be equal to the orientation of the vector specified via the parameter of the constraint. This constraint locks two rotational degrees of freedom of the argument.
```

jmwright (approved):
Looks good, thanks for all the work @adam-urbanczyk 

I'm relying on @marcus7070 and @lorenzncode quite a bit for this review since it seems that they understand the subtleties of the constraint system better than I do.

adam-urbanczyk:
Great, I'll merge tomorrow.

PR #992, opened 2022-02-11
Remove the hdf dep

adam-urbanczyk:
Remove the pinned hdf dep


adam-urbanczyk:
Trivial change - merging

whophil:
Do you know what changed which allowed this pin to be removed? I have this same `hdf5` pin in some other projects.

Issue #565, opened 2021-01-06
Topology traversal and topology based selectors


adam-urbanczyk:
E.g. find all wires connected to a vertex

Look here for reference: https://github.com/tpaviot/pythonocc-utils/blob/master/OCCUtils/Topology.py

michaelgale:
In a way, I use custom Selectors (in cqkit) to achieve this result, i.e. topology traversal.  Using the example of finding all wires connected to a vertex:

```python
import cad query as cq
from cqkit import *

# make a cube
r = cq.Workplane("XY").rect(1, 2).extrude(5)
# make a selector based on a vertex on the top corner of cube
vs = SharedVerticesWithObjectSelector(obj=Vertex.makeVertex(0.5, 1.0, 5.0))
# select wires connected to vertex
w = r.wires(vs).vals()
# w contains the three wires connected to vertex as expected
```

The `Selector` classes are easily CQ's best feature.  I can't imagine being able to build some of my complex models using other traditional CAD tools.  The Selectors with their pythonic arithmetic, logical and programmatic composition allow for very powerful topology navigation and selection.

michaelgale:
As a followup to the above example, you can just as easily reuse the selector to find all shared shape objects:
```python
# select faces common with vertex
f = r.faces(vs).vals()
# and edges
e = r.edges(vs).vals()

# create another selector for the diagonally opposite vertex
vs2 = SharedVerticesWithObjectSelector(obj=Vertex.makeVertex(-0.5, -1.0, 5.0))
# now select the single wire at the top
wtop = r.wires(AndSelector(vs, vs2)).vals()
# or all wires except bottom "floor"
w = r.wires(vs + vs).vals()
```

Issue #371, opened 2020-05-29
Advancde selectors


adam-urbanczyk:
Placeholder issue to discuss about extending selectors

- selectors on tags
- selectors based on history
- selectors for topology traversal

michaelgale:
I'll kick things off with a few notes about how I'm going to refactor my own CQ-Kit Selector classes to make them more general purpose and to arrange them into a more logical taxonomy about what they do.  

## 1. Selection by Geometric Property

`HasCoordinateSelector(coordinate, min_points=1)`
`HasXCoordinateSelector(HasCoordinateSelector)`
`HasYCoordinateSelector(HasCoordinateSelector)`
`HasZCoordinateSelector(HasCoordinateSelector)`
- returns objects which have any or all of the specified X,Y,Z coordinate.  `min_points` means that the object must have at least this many of its vertices with the desired coordinate. For edges, this means either the start or end point if `min_points=1` or both start and end points if `min_points=2`, etc.

`LengthSelector(lengths, tolerance=DEFAULT_TOL)`
`EdgeLengthSelector(LengthSelector)`
`WireLengthSelector(LengthSelector)`
`RadiusSelector(LengthSelector)`
`DiameterSelector(LengthSelector)`
- returns edges or wires satisfying one or more length criteria

`AreaSelector(areas, tolerance=DEFAULT_TOL)`
- returns faces or wires with an area (centre of mass) criteria

`EdgeCountSelector(counts)`
- returns solids, shells, faces or wires with specified number of edges <=, ==, >=

`VertexCountSelector(counts)`
- returns solids, shells, faces or wires with specified number of vertices <=, ==, >=

Criteria for these selectors can be a single value, string constraint (e.g. "<=5.3"), or list of values/strings (e.g. [4.0, "<2.5", 7.5])

## 2. Selection by Orientation

`VerticalSelector(with_heights=None, tolerance=DEFAULT_TOL)`
`VerticalEdgeSelector(VerticalSelector)`
`VerticalWireSelector(VerticalSelector)`
`VerticalFaceSelector(VerticalSelector)`

- returns objects standing "up" i.e. "almost and exactly" |Z with optional height criteria (convenience by saving one step of logical AND with LengthSelector)

`FlatSelector(at_heights=None, tolerance=DEFAULT_TOL)`
`FlatEdgeSelector(FlatSelector)`
`FlatWireSelector(FlatSelector)`
`FlatFaceSelector(FlatSelector)`

[...]

PR #1013, opened 2022-02-23
Topology traversal selection

Jojain:
Hello I've made a minimal working example of a topology traversal capability for CQ. It aims to answer #565  ( and eventually #371 ) 
Since the design has not been discussed that much in the issues I propose the following design. It it currently limited in capability but if we have an agreement on the design I will add all the needed cases (considering objects connected not necessarily by vertices)

The proposed design would allow to do such thing :
`box = cq.Workplane().box(10, 10, 10).faces(">Z").connected("Edge")`
Which gives the following result : 
image

Or : 
`box = cq.Workplane().box(10, 10, 10).faces(">Z").connected("Face", include_child_shape=True)`
image

Let me know if you have comments or recommandation

### DIFF
--- a/cadquery/cq.py
+++ b/cadquery/cq.py
@@ -37,11 +37,14 @@ from typing import (
 from typing_extensions import Literal
 from inspect import Parameter, Signature
 
+from cadquery.occ_impl.topo_explorer import ConnectedShapesExplorer
+
 
 from .occ_impl.geom import Vector, Plane, Location
-from .occ_impl.shapes import (
+from cadquery.occ_impl.shapes import (
     Shape,
     Edge,
+    Shapes,
     Wire,
     Face,
     Solid,
@@ -1054,6 +1057,27 @@ class Workplane(object):
         a union, cut, split, or fillet.  Compounds can contain multiple edges, wires, or solids.
         """
         return self._selectObjects("Compounds", selector, tag)
+
+    def connected(self, shape_type: Shapes, include_child_shape=False):
+        """
+        Returns a new Workplane with shapes of type `shape_type` that are connected to the previous Workplane objects
+        Example:
+        code::
+            box = Workplane().box(10,10,10).faces(">Z").connected("Edge") 
+            print(len(box.val()))
+            >>> 4 # It selects the 4 edges orthogonal to the face
+        
+        Note that by default it doesn't include searched shapes from the selected child shape, if you wanted
+        the 4 edges orthogonal to the face as well as the 4 edges from the face set `include_child_shape` to True
+        """
+
+        if len(self.objects) > 1:
+            raise NotImplementedError("Only one object workplane supported yet")
+        explorer = ConnectedShapesExplorer(self.findSolid(), self.val())
+
+        objs = explorer.search(shape_type, include_child_shape)
+
+        return self.newObject(objs)
 
     def toSvg(self, opts: Any = None) -> str:
         """
--- a/cadquery/occ_impl/topo_explorer.py
+++ b/cadquery/occ_impl/topo_explorer.py
@@ -1,0 +1,105 @@
+from typing import Optional
+from OCP.TopExp import TopExp_Explorer
+
+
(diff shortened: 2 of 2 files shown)


adam-urbanczyk:
Thanks @Jojain - let me digest this. I must say I planned to implement this in a completely different way - by refactoring of the selectors.

Jojain:
No problem, there isn't much code and I think it's self explaining enough but if you want me to explain feel free to ask.

The main point of this is that it doesn't restrain selections to be thinning down the numbers of objects on the stack. 
With this you can select the top Z face and with chaining operations of selection you can get the bottom Z face.

This is particularly useful on complex model where it is a lot easier for the human brain to think of a succession of simple selection rather than coming up with only one selection call that would target the desired shape.

I would also be happy to know how you would want to refactor selectors to see if I can be of any help !

snoyer:
> The main point of this is that it doesn't restrain selections to be thinning down the numbers of objects on the stack. With this you can select the top Z face and with chaining operations of selection you can get the bottom Z face.

Do selectors _have to_ thin down the number of objects, though? Right now they technically do as they are implemented as filters, but is that a design principle or just an implementation limitation/oversight?

Currently `thing.edges(selector)` tells the selector "I've listed the edges of thing for you, look through them and decide which ones to keep".
If we were willing to make it mean "here's thing, find any of its edges as you see fit" then this `connected()` feature could be implemented with these new selectors.
`box.faces(">Z").connected("Edge")` would become `box.faces(">Z").edges(Connected())` for example.

Here's a dirty monkey-patched proof-of-concept:

```python
from abc import ABC, abstractmethod
from dataclasses import dataclass
from typing import Any, Iterable, Optional, Union

import cadquery as cq
from cadquery import Selector, Shape, Workplane
from topo_explorer import ConnectedShapesExplorer


# new selector API

@dataclass(frozen=True)
class SelectionContext(): #TODO better name?
    workplane: Workplane
    objType: Any
    tag: Optional[str]

    def collectProperty(self, objType:Optional[Any]=None):
        cq_obj = self.workplane._getTagged(self.tag) if self.tag else self.workplane
        return cq_obj._collectProperty(objType if objType else self.objType)

    #TODO additional useful stuff would go here to keep the actual NewSelector class simple/stable

class NewSelector(ABC):
    @abstractmethod
    def select(self, ctx: SelectionContext) -> Iterable[Shape]:
        pass


# monkey-patching it in

_old_selectObjects = Workplane._selectObjects

def _new_selectObjects(
        self: Workplane,
        objType: Any,
        selector: Optional[Union[Selector, str, NewSelector]] = None,
        tag: Optional[str] = None,
    ) -> Workplane:
        if isinstance(selector, NewSelector):
            ctx = SelectionContext(workplane=self, objType=objType, tag=tag)
            return self.newObject(selector.select(ctx))
        else:
            return _old_selectObjects(self, objType, selector, tag)

Workplane._selectObjects = _new_selectObjects


# example usage

class Connected(NewSelector):
    def __init__(self, include_source: bool=False) -> None:
        self.include_child_shape = include_source
    
    def select(self, ctx: SelectionContext) -> Iterable[Shape]:
        types = {
            'Edges': 'Edge',
            'Faces': 'Face',
            #TODO add the rest or change `topo_explorer.ENUM_MAPPING` to match
        }
        solid = ctx.workplane.findSolid()
        for object in ctx.workplane.objects:
            explorer = ConnectedShapesExplorer(solid, object)
            yield from explorer.search(types[ctx.objType], self.include_child_shape)


boxes = cq.Workplane().box(10, 10, 10) - cq.Workplane().box(2, 10, 10)
edges = boxes.faces("<Z").faces('<X').edges(Connected())
faces = boxes.faces("<Z").faces('>X').faces(Connected(include_source=True))
```

I don't use cadquery enough to know if this is the one true way to refactor selectors but something along those lines would be an improvement for topology-based use cases.

adam-urbanczyk:
Thanks for your work. In the end this was implemented as `siblings` and `ancestors` methods.

PR #1016, opened 2022-02-27
polarArray start angle, and rotation fix (#1006)

### LINKED ISSUE #1006: Workplane polarArray incorrect startAngle and angle (according to its docstring and convention)
From the `polarArray` docstring:
```
        :param startAngle: Starting angle (degrees) of array. 0 degrees is
            situated along local X-axis.
        :param angle: The angle (degrees) to fill with elements. A positive
            value will fill in the counter-clockwise direction.
```

Would you be opposed to correcting `polarArray` behavior to follow the docstring?

There are no doc examples and few other references of usage.  I think the issue wasn't noticed because usage was typically with `angle=360` e.g. #118.

lorenzncode:
To resolve #1006.

### DIFF
--- a/cadquery/cq.py
+++ b/cadquery/cq.py
@@ -63,6 +63,7 @@ from .sketch import Sketch
 CQObject = Union[Vector, Location, Shape, Sketch]
 VectorLike = Union[Tuple[float, float], Tuple[float, float, float], Vector]
 CombineMode = Union[bool, Literal["cut", "a", "s"]]  # a : additive, s: subtractive
+TOL = 1e-6
 
 T = TypeVar("T", bound="Workplane")
 """A type variable used to make the return type of a method the same as the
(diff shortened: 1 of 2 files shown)


jmwright (approved):
Looks good, thanks @lorenzncode

cadquery/cq.py
```diff
@@ -1486,57 +1486,47 @@ def polarArray(
...
-            loc = Location(Vector(x, y))
-
-        locs = [loc]
+        if count < 1:
+            raise ValueError(f"At least 1 element required, requested {count}")
 
         # Calculate angle between elements
         if fill:
-            if angle % 360 == 0:
+            if abs(math.remainder(angle, 360)) < 1e-6:
```
adam-urbanczyk, line 1508: Could you make this 1e-6 into a module constant, say `TOL`?

cadquery/cq.py
```diff
@@ -1486,57 +1486,47 @@ def polarArray(
...
-            y = radius * math.cos(phi)
-
-            if rotate:
-                loc = Location(Vector(x, y), Vector(0, 0, 1), -phi_deg)
             else:
-                loc = Location(Vector(x, y))
+                # Inclusive start and end
+                angle = angle / (count - 1) if count > 1 else startAngle
+
+        def locs():
```
adam-urbanczyk, line 1514: I'd propose to keep the original linear flow here (i.e. no nested functions).

adam-urbanczyk (changes requested):

cadquery/cq.py
lorenzncode: Module constant added.  Perhaps in the future tolerance could be a constant at the package level as there is a similar comparison in `Sketch.parray`.  Then maybe a different constant is needed for angle comparison vs others.

cadquery/cq.py
lorenzncode: Got it.  I thought I'd experiment with a generator but probably no benefit as `pushPoints` builds a list.

I'm having fun with CQ and certainly fortunate that I'm still able to do so.  Thanks for taking the time to review.

adam-urbanczyk (approved):

adam-urbanczyk:
Thanks @lorenzncode !

PR #1023, opened 2022-03-05
Fix screw holes in parametric enclosure example

prplz:
Updates parametric enclosure example for centerOption changes (#532)


prplz:
I just realized this could also be fixed by skipping workplane entirely: `cutlip.faces(">Z").rect(POSTWIDTH, POSTLENGTH, forConstruction=True).vertices()`

jmwright:
@prplz I think that skipping the `workplane()` call is too likely to cause confusion example. Was this the only `workplane` call that was causing problems, is that why you added `centerOption` to it but none of the others?

prplz:
This particular example needs it because the lid is moved off center, causing the screw holes to not get added without my change

jmwright (approved):
Ok, thanks for the contribution.

jmwright:
Simple change, merging with only one review.

PR #1029, opened 2022-03-15
Apply DIR_SCALING to the fixed rotation cost

adam-urbanczyk:


jmwright (approved):
Thanks @adam-urbanczyk I think this looks good.

lorenzncode (commented):
I tested the PR changes.  I confirm the solver converges when testing the new pytest case.

I tried adding another long beam ("three") to the frame.

```
frame = (
    cq.Assembly()
    .add(make_beam(length=800, size=60, draft_angle=45), name="one")
    .add(make_beam(length=200, size=60, draft_angle=45), name="two")
    .add(make_beam(length=800, size=60, draft_angle=45), name="three")
    .constrain("one", "Fixed")

    .constrain("two", "FixedRotation", (0, 0, 90))
    .constrain("one@faces@<Y", "two@faces@>Y", "Point")

    .constrain("three", "FixedRotation", (0, 180, 0))
)
```

To complete the frame beam, "three" should be rotated 180 degrees.  The try with `(0, 180, 0)` was not successful, although changing to (180, 0, 180) did work.

Using `FixedRotation` and `Point` constraints, I wasn't quite able to complete the frame.

```
frame = (
    cq.Assembly()
    .add(make_beam(length=800, size=60, draft_angle=45), name="one")
    .add(make_beam(length=200, size=60, draft_angle=45), name="two")
    .add(make_beam(length=800, size=60, draft_angle=45), name="three")
    .add(make_beam(length=200, size=60, draft_angle=45), name="four")
    .constrain("one", "Fixed")

    .constrain("two", "FixedRotation", (0, 0, 90))
    .constrain("one@faces@<Y", "two@faces@>Y", "Point")

    .constrain("three", "FixedRotation", (180, 0, 180))
    #.constrain("three@faces@>Y", "four@faces@>Y", "Point")
    #.constrain("three@faces@<Y", "two@faces@<Y", "Point")

    .constrain("four", "FixedRotation", (0, 0, -90))
    .constrain("one@faces@>Y", "four@faces@<Y", "Point")
)
```

frame

I tested another case with two `FixedRotationAxis` constraints.

```
assy = (
    cq.Assembly()
    .add(b1, name="b1")
    .add(b2, name="b2")
    .constrain("b1", "Fixed")
    .constrain("b2", "FixedRotationAxis", (0, x))
    .constrain("b2", "FixedRotationAxis", (1, y))
)
```

Solving for various values of rotation and plotting the result shows that like the frame example the solver does not converge for values beyond +-90 rotation in axis 1.

df_dir_scaling_m180_210_30


In the plot, cost is represented by the colormap where green is cost < 1e-10 (constraints are met or "pass"), red is cost > 1e-6 (constraints not met or "fail"), yellow is in range of 1e-10 to 1e-6 or "near".


When axis 1 rotation == 90, any value of axis 0 rotation (other than zero) fails to converge.

Here is another plot with step size of 1 degree or 361*361 calls to `assy.solve()`.

df_dir_scaling_m180_181_1

I think that for the `FixedRotation` constraint to be more generally useful, the solver probably would need to handle more combinations of rotation over the axes.  I'm trying to understand the current limitations so far.  I don't have a good idea for improvement.

adam-urbanczyk:
Thanks @lorenzncode , probably the cost function as implemented now is not continuous.

adam-urbanczyk:
Not a robust solution, closing.

Issue #1031, opened 2022-03-15
Assembly solver FixedRotation does not converge


lorenzncode:
Here is an example case where FixedRotation/FixedRotationAxis does not converge.

```
import cadquery as cq

b1 = cq.Workplane().box(1, 2, 3)
b2 = cq.Workplane().box(1, 2, 3)

assy = (
    cq.Assembly()
    .add(b1, name="b1")
    .add(b2, name="b2")
    .constrain("b1", "Fixed")
    .constrain("b2", "FixedRotationAxis", (0, 0))
    .constrain("b2", "FixedRotationAxis", (1, 90))
    .constrain("b2", "FixedRotationAxis", (2, 90))
)

assy.solve()

show_object(assy)
```

lorenzncode:
I'd like to propose the following potential solution to the `FixedRotation` constraint issue. 

- Similar to the `Fixed` constraint, apply `fixed_cost` to `FixedRotation`. 
- Again similar to `Fixed`, modify the transform that is passed to the cost function. Instead of identity transform, fix the rotation part.

lorenzncode:
Duplicate of #1027.

Issue #1025, opened 2022-03-10
Simple square bar frame assembly doesn't line up


spike77453:
If I'm not mistaken, this should produce a simple square bar frame:
```
import cadquery as cq

def make_beam(size, length, draft_angle):
  beam = (
    cq.Workplane("XZ")
    .rect(size, size)
    .extrude(-length)
  )

  if draft_angle != 0:
    beam = (
      beam
      .faces(">Y").workplane().transformed(offset=cq.Vector(size/2,0,0),rotate=cq.Vector(0,-draft_angle,0))
      .split(keepBottom=True)
      .faces("<Y").workplane().transformed(rotate=cq.Vector(0,draft_angle,0))
      .split(keepBottom=True)
    )

  return beam


y_beam = make_beam(length=800, size=60, draft_angle=45)
x_beam = make_beam(length=700, size=60, draft_angle=45)

frame = (
  cq.Assembly()
  .add(x_beam, name='X_front')
  .add(x_beam, name='X_back')
  .add(y_beam, name='Y_left')
  .add(y_beam, name='Y_right')

    # Constrain beams to make frame
  .constrain('Y_left@faces@<Z', 'X_front@faces@<Z', 'Axis', param=0)
  .constrain('Y_left@faces@<Y', 'X_front@faces@>Y', 'Plane')

  .constrain('Y_left@faces@<Z', 'X_back@faces@<Z', 'Axis', param=0)
  .constrain('Y_left@faces@>Y', 'X_back@faces@<Y', 'Plane')

  .constrain('Y_left@faces@<Z', 'Y_right@faces@<Z', 'Axis', param=0)
  .constrain('X_front@faces@<Y', 'Y_right@faces@>Y', 'Plane')
)
frame.solve()

if "show_object" in locals():
  show_object(frame, 'frame')
```

However, the result looks like this:
![frame_1](https://user-images.githubusercontent.com/25122064/157643111-89783a3f-487c-4606-ac27-30ce4700cc2d.png)

Changing

```
.constrain('X_front@faces@<Y', 'Y_right@faces@>Y', 'Plane')
```
to
```
.constrain('X_back@faces@>Y', 'Y_right@faces@<Y', 'Plane')
```
(which just moves the constraint from the front right corner to the back right corner) fixes the issue.

Including 
```
.constrain('X_back@faces@>Y', 'Y_right@faces@<Y', 'Point')
```
(which should overconstrain the assembly) also works:
[...]

PR #1034, opened 2022-03-23
Change FixedRotation constraint to fixed_cost

### LINKED ISSUE #1027: FixedRotation constraint doesn't fix rotation
Considering the following example:
```
import cadquery as cq

def make_beam(size, length, draft_angle = 0):
  beam = (
    cq.Workplane('YX')
    .sketch()
    .trapezoid(length, size, draft_angle)
    .finalize()
    .extrude(size)
  )
  return beam

frame = (
  cq.Assembly()
  .add(make_beam(length=800, size=60, draft_angle=45), name='one')
  .add(make_beam(length=200, size=60, draft_angle=45), name='two')

  .constrain('one', 'FixedRotation',(0,0,0))
  .constrain('two', 'FixedRotation',(0,0,90))
  .constrain('one@faces@<Y', 'two@faces@>Y', 'Point')
)
frame.solve()

if "show_object" in locals():
  show_object(frame, 'frame')
```
I would expect the first two constraints to fix the rotation of both beams and the third constraint to translate beam 'two' in place.   However:

test

If I understand https://cadquery.readthedocs.io/en/latest/assy.html#fixedrotation correctly, there should only be translatory degrees of freedom left after using a 'FixedRotation' constraint so I'm a bit puzzled by the result.

lorenzncode:
This updates the `FixedRotation` constraint to potentially resolve #1027, #1031.

I'll mark as a draft pending more testing and discussion on this approach.

### DIFF
--- a/cadquery/assembly.py
+++ b/cadquery/assembly.py
@@ -358,6 +358,7 @@ class Assembly(object):
 
         i = 0
         locked = []
+        locked_rotation = {}
         for c in self.constraints:
             for name in c.objects:
                 if name not in ents:
@@ -365,6 +366,8 @@ class Assembly(object):
                     i += 1
                 if c.kind == "Fixed" or name == self.name:
                     locked.append(ents[name])
+                if c.kind == "FixedRotation":
+                    locked_rotation[ents[name]] = c.param
 
         # Lock the first occuring entity if needed.
         if not locked:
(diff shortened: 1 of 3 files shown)


lorenzncode:
Closing - as was mentioned in #1025 more fundamental changes to the solver are in progress.

PR #1037, opened 2022-03-30
Add a dockerfile

roipoussiere:
I created a dockerfile for cadquery ([here on docker hub](https://hub.docker.com/r/cadquery/cadquery)). This is an easy way to try cq without having to install Conda.

Does it make sense for you to add it to this repo?

## Usage

Let's suppose we have the following script:

```py
import cadquery as cq
box = cq.Workplane("XY").box(1, 2, 3)
print(box.toSvg())
```

You can generate the svg with:

    docker run -i --rm cadquery/cadquery python < ./box.py > box.svg

If you need more control to your project (installing Python dependencies, use external files, etc), you can create a docker volume:

    docker run --rm -it -v $(pwd)/examples:/home/cq cadquery/cadquery

It will open a shell prompt on which you can work.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,0 +1,19 @@
+# Usage:
+#     docker run --rm -i cadquery python < ./box.py > box.svg
+#     docker run --rm -it -v $(pwd)/examples:/home/cq cadquery
+
+FROM continuumio/miniconda3:4.10.3
+
+RUN adduser --disabled-password --gecos "Default user" --uid 1000 cq && \
+    apt-get update -y && \
+    apt-get install --no-install-recommends -y libgl1-mesa-glx libglu1-mesa && \
+    apt-get clean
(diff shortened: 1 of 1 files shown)


jmwright:
@roipoussiere Can you merge master into your branch? There was a very recent change that should fix the AppVeyor failure.

adam-urbanczyk:
@roipoussiere  could you add some kind of simple test?

roipoussiere:
What kind?

If you CI runs on docker, it's not trivial to test the dockerfile because it would require to run docker on docker.

adam-urbanczyk:
Simple smoke test would be fine - build the image and verify that it runs.

roipoussiere:
Now we have the `ocp-vtk` wheel, a cadquery docker image seem not really necessary, so I'm going to close this.

PR #1066, opened 2022-04-27
Added the ability for the user to adjust the tolerance and angular tolerance of VRML tessellation

jmwright:
With the KiCAD 3D model generator the generated VRML files were 6MB, which was too large compared to the ~200KB original files. I added the ability for the user to adjust the `tolerance` and `angularTolerance` so that I could decrease the file size. I don't think I'll be able to keep the same quality at ~200MB, but I can get vast improvements in file size with these changes.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,3 +19,9 @@ simple.step
 simple.stp
 simple.xml
 test.brep
+nested.caf
+nested.glb
+nested.stp
+nested.wrl
+nested.xml
+nested.zip
--- a/cadquery/assembly.py
+++ b/cadquery/assembly.py
@@ -442,7 +442,7 @@ class Assembly(object):
         elif exportType == "XML":
             exportCAF(self, path)
         elif exportType == "VRML":
-            exportVRML(self, path)
+            exportVRML(self, path, tolerance, angularTolerance)
         elif exportType == "GLTF":
             exportGLTF(self, path, True, tolerance, angularTolerance)
         elif exportType == "VTKJS":
--- a/cadquery/occ_impl/assembly.py
+++ b/cadquery/occ_impl/assembly.py
@@ -179,7 +179,7 @@ def toVTK(
     loc: Location = Location(),
     color: Tuple[float, float, float, float] = (1.0, 1.0, 1.0, 1.0),
     tolerance: float = 1e-3,
-    angularTolerance: float = 0.1
+    angularTolerance: float = 0.1,
 ) -> vtkRenderer:
 
     loc = loc * assy.loc
@@ -189,7 +189,9 @@ def toVTK(
         color = assy.color.toTuple()
 
     if assy.shapes:
-        data = Compound.makeCompound(assy.shapes).toVtkPolyData(tolerance, angularTolerance)
+        data = Compound.makeCompound(assy.shapes).toVtkPolyData(
+            tolerance, angularTolerance
+        )
 
         mapper = vtkMapper()
         mapper.SetInputData(data)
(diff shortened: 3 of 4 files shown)


jmwright:
I think this is ready for review now. @adam-urbanczyk and @lorenzncode I requested reviews from you two, but if either of you don't have time to review let me know and I'll request from someone else.

adam-urbanczyk (approved):

.gitignore
```diff
@@ -19,3 +19,9 @@ simple.step
 simple.stp
 simple.xml
 test.brep
+nested.caf
```
lorenzncode, line 22: I think this is fine.  Just to add I was thinking any new tests for exporters should be implemented with pytest to make use of tmp_path_factor.  Then we wouldn't have this problem.

For example:
https://github.com/lorenzncode/cadquery/blob/e53bdc0951e67da23cccf31f5b334022b23af654/tests/test_exporters.py#L220-L227

lorenzncode (changes requested):
Let's expose tolerance and angular tolerance to `Assembly.save` as well.
https://github.com/CadQuery/cadquery/blob/2eeabc8bd911fe5e1ea5ec80743fb1c712bb8d4e/cadquery/assembly.py#L445
`            exportVRML(self, path, tolerance, angularTolerance)`

.gitignore
jmwright: Should we create an issue for converting the existing tests as well?

jmwright: I'm not sure I understand what you mean about exposing the tolerances to `Assembly.save`. They [seem to be already exposed](https://github.com/CadQuery/cadquery/blob/9e9b452e66f7925f7c15689f3e07da47b13f2454/cadquery/assembly.py#L422), unless you meant somewhere above or below that level in code.

lorenzncode: > Should we create an issue for converting the existing tests as well?

Yes, makes sense.  I can follow up and file a separate issue.


> I'm not sure I understand what you mean about exposing the tolerances to Assembly.save. 

You are correct.  The tolerances are already exposed.  I meant to say that the tolerances are not currently passed on to `exportVRML`.

https://github.com/CadQuery/cadquery/blob/a1f46463ce3f5eab25ff2c777918705ff03912cb/cadquery/assembly.py#L444-L445

The request is to pass `tolerance`, and `angularTolerance` to the `exportVRML` call here.  Sorry that wasn't clear.

jmwright: @lorenzncode I've made that change now. The only oddity I see with exposing that there is that the default `tolerance` and `angularTolerance` values in the `Assembly.save` method signature are set up for GLTF, which are 0.1 and 0.1. This is different than the defaults for `exportVRML`, which are 1e-3 and 0.1 respectively. I don't think it's a huge problem, and I think it's better to have those settings exposed, but in some situations the user might get lower resolution than expected, and than was intended by the original dev who implemented `Assembly.save`, which I believe is @adam-urbanczyk .

lorenzncode: Given `exportVRML` accepts tolerance settings now, and `Assembly.save` provides these parameters I think the user would expect the tolerance settings to be considered for all formats that accept them.

> This is different than the defaults for exportVRML, which are 1e-3 and 0.1 respectively.

I see, agree better to pass the tolerance settings.  Perhaps GLTF and VRML tolerances could be changed to be consistent if necessary.

lorenzncode (approved):

jmwright:
@lorenzncode Fixed CI, so I cna merge this now. Thanks for the reviews.

Issue #1082, opened 2022-05-11
Additional Option for STEP Export


jmwright:
The STEP files generated by our default export process were too large for the KiCAD generator update project, with the files being about 2x larger. Setting the following mode fixed the issue, and seems to be the mode they used in the past (although they used the C++ form).

```python
Interface_Static.SetIVal_s('write.surfacecurve.mode', 0)
```

I added it just above [this line](https://github.com/CadQuery/cadquery/blob/c32c08eb4b6537a6116dd5e5ff2360c7ab038fce/cadquery/occ_impl/exporters/assembly.py#L40) in the source in my local copy.

So my question is this. How should this option be added and exposed in the code? Is this an edge case, or will this option be of use to the broader user base? Please let me know if you have any thoughts @adam-urbanczyk @lorenzncode @dcowden @marcus7070

michaelgale:
This is main reason I added alternative STEP export to [CQ-Kit](https://github.com/michaelgale/cq-kit)--the STEP files with the P-Curves were simply too big.  Furthermore, the floating point residue round off of coordinate values also consumed needless bytes resulting in bigger files.  More details are in the [StepFileExporter](https://github.com/michaelgale/cq-kit/blob/f5820c67030f14ae8e387872ae859ad023983c8e/cqkit/cq_files.py#L198) class docstring.

adam-urbanczyk:
Sounds like a very specific use case. BTW: why would KiCAD use STEP and not the internal format of OCCT (`.brep`) is not clear to me.

jmwright:
@michaelgale @lorenzncode @adam-urbanczyk Please have a look at the PR [here](https://github.com/CadQuery/cadquery/pull/1083) and let me know your thoughts. It only accepts two parameters right now, but the code is set up to allow more without cluttering the method signature later. If you have a suggestion on how to reduce the code duplication let me know.

@adam-urbanczyk I'm not sure of the reasons why the KiCAD team uses STEP, but they've been using it for a long time and so I would expect a fair amount of technical debt that will keep them headed down that path for the foreseeable future.

adam-urbanczyk:
> @michaelgale @lorenzncode @adam-urbanczyk Please have a look at the PR [here](https://github.com/CadQuery/cadquery/pull/1083) and let me know your thoughts. It only accepts two parameters right now, but the code is set up to allow more without cluttering the method signature later. If you have a suggestion on how to reduce the code duplication let me know.
> 
[...]

PR #1083, opened 2022-05-13
Added quality controls to STEP export for shape and assembly

jmwright:
This is in support of the KiCAD model generator, per #1082 

What is accepted via kwargs can be added to over time, but for now there are just 2 parameters that influence file size.

### DIFF
--- a/cadquery/assembly.py
+++ b/cadquery/assembly.py
@@ -420,6 +420,7 @@ class Assembly(object):
         exportType: Optional[ExportLiterals] = None,
         tolerance: float = 0.1,
         angularTolerance: float = 0.1,
+        **kwargs,
     ) -> "Assembly":
         """
         save as STEP or OCCT native XML file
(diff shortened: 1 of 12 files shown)


jmwright:
@lorenzncode I'm adding tests for the single object and assembly STEP exports, and setting them up to use the temporary directory method you've been using. `test_assembly.test_step_export` worked fine when I added `tmp_path_factory` as an argument, but when I tried the same thing with `test_exporters.testSTEPOptions` I got this error:
```
TypeError: TestExporters.testSTEPOptions() missing 1 required positional argument: 'tmp_path_factory'
```
Am I missing a decorator or something?

lorenzncode:
@jmwright I'd suggest to move the new tests outside of the `TestExporters` class to define as a `pytest` test rather than `unittest`.  `TestExporters` is a subclass of `unittest.TestCase`.

lorenzncode (approved):
I've tested this branch and confirm file size decrease with write_pcurves False.  I also confirm the uncertainty value change with change in write.precision.mode.

I made some minor changes:

* exporters.export: expose STEP options
* Use default OCCT write.precision.mode value 0
* Doc update

Regarding docstrings, my preference for these export options, would be to defer to OCCT documentation for the most part.  I've modified the docstring to provide short descriptions only, rather than try to reinterpret the meanings especially for write.precision.mode.

jmwright:
@lorenzncode Thanks, these changes look good to me.

jmwright:
@adam-urbanczyk Any objections to me merging this?

PR #1089, opened 2022-05-24
Updating the readme for new pip install instructions

jmwright:
This updates the readme to let users know there are two installation methods now (conda and pip), and lists the limitations of the pip installation. It also slips in a small tweak that was needed in setup.py. PyPi would not accept the long_description in the markdown format without an explicit setting to enable it.

@fpq473 you have been closely involved in getting the pip installs working, so please have a look if you get time.

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,7 +44,37 @@ It is currently possible to use CadQuery for your own projects in 3 different wa
     * Linux [installation video](https://youtu.be/sjLTePOq8bQ)
     * Windows [installation video](https://youtu.be/3Tg_RJhqZRg)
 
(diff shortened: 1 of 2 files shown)


fpq473:
Text looks correct and clear to me.

Minor: PyPi => PyPI; PyPi makes me think of a RaspberryPi ;)

jmwright:
> PyPi => PyPI

Fixed, thanks.

lorenzncode (approved):

jmwright:
@lorenzncode and @fpq473 have reviewed this and it's only a change to the readme, so I'm going to go ahead and merge.

PR #1100, opened 2022-06-08
stop `eachpoint(..., useLocalCoordinates=False)` from manipulating the object's location

### LINKED ISSUE #1099: `eachpoint(..., useLocalCoordinates=False)` manipulates the point's location when it shouldn't
Here's what the docs say `eachpoint()` should do:
https://github.com/CadQuery/cadquery/blob/a5fadebe72b2a46881c3847cca748b8cf716ccd0/cadquery/cq.py#L2440-L2441

When `each()` is called with `useLocalCoordinates=False`, it does no manipulation of the object before passing it on to the callback function:
https://github.com/CadQuery/cadquery/blob/a5fadebe72b2a46881c3847cca748b8cf716ccd0/cadquery/cq.py#L2423

but when `eachpoint()` is called with `useLocalCoordinates=False`, it manipulates/transforms (`p * loc`) the object's location before passing it on to the callback function:
https://github.com/CadQuery/cadquery/blob/a5fadebe72b2a46881c3847cca748b8cf716ccd0/cadquery/cq.py#L2479

I guess it probably shouldn't do that.

greyltc:
fixes https://github.com/CadQuery/cadquery/issues/1099

### DIFF
--- a/cadquery/cq.py
+++ b/cadquery/cq.py
@@ -2476,7 +2476,7 @@ class Workplane(object):
         if useLocalCoordinates:
             res = [callback(p).move(loc) for p in pnts]
         else:
-            res = [callback(p * loc) for p in pnts]
+            res = [callback(p) for p in pnts]
 
         for r in res:
             if isinstance(r, Wire) and not r.forConstruction:
--- a/tests/test_cadquery.py
+++ b/tests/test_cadquery.py
@@ -5119,7 +5119,7 @@ class TestCadQuery(BaseTest):
                     Face.makePlane(1, 1),
                 ]
             )
-            .eachpoint(lambda l: Face.makePlane(1, 1).locate(l))
+            .eachpoint(lambda l: Face.makePlane(1, 1).locate(l), True)
         )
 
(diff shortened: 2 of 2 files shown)


greyltc:
Even better would be a change that turns `eachpoint()` into a thin wrapper around `each()` with only the logic for converting the objects to points.

jmwright:
@greyltc The eachpoint test seems to be failing on this PR.

greyltc:
Sorry, maybe I'm looking in the wrong place, but I can't see the details of the failure. All I can see is
```
Check failure on line 1059 in Build log

@azure-pipelinesazure-pipelines
/ CadQuery.cadquery (Linux_3_10)
Build log #L1059
Bash exited with code '1'.
```
I clicked on "View more details on Azure Pipelines" there but that takes me to a page that says "build not found". This PR makes a fundamental change to the way `eachpoint` works with `useLocalCoordinates=False` so I assume the test pass criteria needs to be rewritten to match that, but it would be nice to see the verbose check failure output to make sure.

Issue #965, opened 2022-01-19
[EASY QUESTION] Exporting an assembly model as .stl format


traceforward:
I can not export an assembly model which I created with cq.Assembly() with exporters.exportSTL. How can I do that. Please help.
Thanks in advance.

adam-urbanczyk:
Use `toCompound` and then export.

Jojain:
@adam-urbanczyk Since this is not the first time someone ask this, shouldn't we consider the `exportStl` to automatically make a compound of the assembly ?
I'm pretty sure it's trivial to do, and would be more transparent for users

adam-urbanczyk:
Maybe, but I really wonder what is the use of such an STL.

jmwright:
This issue has already been addressed, but I'm going to post [this response](https://gitlab.com/kicad/libraries/kicad-packages3D-generator/-/issues/5#note_1226473401) here for archival purposes. It's a good explanation from someone in the KiCAD community of why you would want to fuse and print an assembly.

PR #1101, opened 2022-06-08
Add STL to Assemblys export formats

qwelyt:
Turn the assembly into a compound and use Shapes export method to export
the assembly. This way the export function does not need to be
duplicated, and it is what users have been doing manually to get their
objects.

Should solve https://github.com/CadQuery/cadquery/issues/965

### DIFF
--- a/cadquery/assembly.py
+++ b/cadquery/assembly.py
@@ -27,7 +27,7 @@ from .selectors import _expression_grammar as _selector_grammar
 # type definitions
 AssemblyObjects = Union[Shape, Workplane, None]
 ConstraintKinds = Literal["Plane", "Point", "Axis", "PointInPlane"]
-ExportLiterals = Literal["STEP", "XML", "GLTF", "VTKJS", "VRML"]
+ExportLiterals = Literal["STEP", "XML", "GLTF", "VTKJS", "VRML", "STL"]
 
 PATH_DELIM = "/"
 
@@ -455,7 +455,7 @@ class Assembly(object):
 
         if exportType is None:
             t = path.split(".")[-1].upper()
-            if t in ("STEP", "XML", "VRML", "VTKJS", "GLTF"):
+            if t in ("STEP", "XML", "VRML", "VTKJS", "GLTF", "STL"):
                 exportType = cast(ExportLiterals, t)
             else:
                 raise ValueError("Unknown extension, specify export type explicitly")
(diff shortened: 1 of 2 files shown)


SDI8:
Shouldn't all of the "shape-formats" be supported then (and not just STL)?
This situation is very similar to what is happening with Workplanes in [`exporters.export`](https://github.com/CadQuery/cadquery/blob/a5fadebe72b2a46881c3847cca748b8cf716ccd0/cadquery/occ_impl/exporters/__init__.py#L59), where any Workplane gets converted to a compound first. 
You could achieve similar functionality to your proposal by allowing Assemblies in[ `exporters.export`](https://github.com/CadQuery/cadquery/blob/a5fadebe72b2a46881c3847cca748b8cf716ccd0/cadquery/occ_impl/exporters/__init__.py#L36) - although I'm not in a position to foresee all issues that that could entail.

jmwright:
@lorenzncode @adam-urbanczyk Any thoughts on this?

adam-urbanczyk:
STL has been discussed already. It does not support all the features of assy (color), but OK popular request and whatnot. -1 for other formats though.

lorenzncode:
I wasn't sure if the discussion in #965 had concluded.  Regarding the use of such an STL, it may be that the user wants to access the solver functionality for a particular part.  A user on Discord mentioned this approach, and asked "Am I misusing Assembly?".

adam-urbanczyk (approved):

cadquery/assembly.py
```diff
@@ -468,6 +468,8 @@ def save(
             exportGLTF(self, path, True, tolerance, angularTolerance)
         elif exportType == "VTKJS":
             exportVTKJS(self, path)
+        elif exportType == "STL":
+            self.toCompound().exportStl(path, tolerance, angularTolerance)
```
lorenzncode, line 472: Add support for ascii?  ` assy.save("out.stl", ascii=True)`

```py 
            ascii = False
            if "ascii" in kwargs and kwargs["ascii"]:
                ascii = True
            self.toCompound().exportStl(path, tolerance, angularTolerance, ascii)
```

lorenzncode: ascii support could be revisited in a separate PR.

lorenzncode (approved):

adam-urbanczyk:
Should I merge @jmwright @lorenzncode ?

jmwright:
+1

lorenzncode:
+1

adam-urbanczyk:
Thanks @qwelyt !

PR #1116, opened 2022-07-06
fix docs for interpPlate to use correct camelCase for pushPoints

jdegenstein:
interpPlate references pushpoints which is the incorrect case and does not exist. I also reduced the linewidth of the docstring since it was >>88 chars in length.


jmwright:
Thanks @jdegenstein !

PR #1106, opened 2022-06-17
Add option to export binary STL

lorenzncode:
* exporters.export add "ascii" (or "ASCII") options dict
* Shape.exportStl add "ascii" bool param
* fix corrupt docstrings/typos in shapes.py

PR #1117, opened 2022-07-07
Fix exportStl tolerance default value docs

sethfischer:


jmwright:
Thanks @sethfischer 

@lorenzncode Feel free to merge if this looks good to you.

lorenzncode:
Thanks for raising this.  Since there is already another PR in-progress (#1106) that modifies this method let's make the change there. The fix is now included in #1106. It will be easier to include the fix there and avoid a potential merge conflict.

Issue #1125, opened 2022-07-26
Docs type field is redundant (in some cases)


lorenzncode:
The code now includes type hints.

The docstring specifying say ```:type myparam: str``` is redundant and could be removed.

In some cases, such as,  ```Workplane.vertices``` the field elaborates on the valid types so adds value and must not be removed.

Issue #922, opened 2021-11-20
docs -  invalid reStructuredText in docs and docstrings


lorenzncode:
I'm opening this issue to track reStructedText problems in the docs/docstrings.

1. Warnings reported during build (see log below)
2. Other invalid reST in docstrings
    
See e.g. Workplane first, item, last, end where an empty line must be inserted for proper formatting.

    
    def end(self, n: int = 1) -> "Workplane":
        """
        Return the nth parent of this CQ element
        :param n: number of ancestor to return (default: 1)
        :rtype: a CQ object
        :raises: ValueError if there are no more parents in the chain.

        For example::

            CQ(obj).faces("+Z").vertices().end()

        will return the same as::

            CQ(obj).faces("+Z")
        """
   

This docstring results in:

![image](https://user-images.githubusercontent.com/16394272/142740284-16e95cec-4ad2-4d63-8fbe-bf2ed647b092.png)


With (pending) changes for #819, adding empty line to docstring, (and removing unnecessary rtype):

    
        """
        Return the nth parent of this CQ element

        :param n: number of ancestor to return (default: 1)
        :raises: ValueError if there are no more parents in the chain.

        For example::

            CQ(obj).faces("+Z").vertices().end()

        will return the same as::

            CQ(obj).faces("+Z")
        """
    

Now we have Parameters:, Raises:, and Return Type. 

![image](https://user-images.githubusercontent.com/16394272/142740440-00f540e2-6403-41b9-993e-7d222d6ffdaa.png)


 

Building the docs results in 16 warnings:

```
cadquery/occ_impl/shapes.py:docstring of cadquery.occ_impl.shapes.Shape.facesIntersectedByLine:7: WARNING: Field list ends without a blank line; unexpected unindent.
cadquery/occ_impl/shapes.py:docstring of cadquery.occ_impl.shapes.Wire.assembleEdges:6: WARNING: Unexpected indentation.
cadquery/cq.py:docstring of cadquery.cq.Workplane.ellipse:17: WARNING: Unexpected indentation.
cadquery/cq.py:docstring of cadquery.cq.Workplane.extrude:21: WARNING: Bullet list ends without a blank line; unexpected unindent.
cadquery/cq.py:docstring of cadquery.cq.Workplane.extrude:22: WARNING: Block quote ends without a blank line; unexpected unindent.
cadquery/cq.py:docstring of cadquery.cq.Workplane.interpPlate:11: WARNING: Unexpected indentation.
[...]

PR #1126, opened 2022-07-26
Fix docstring corruption (#922 #1125)

lorenzncode:
Fix docstring corruption (invalid reStructuredText which also breaks HTML docs) in shapes.py and some cleanup.


lorenzncode:
Closing, changes added to #1251.

PR #1140, opened 2022-08-30
Update pip related commands

adam-urbanczyk:
I'd also consider mentioning conda first - it is much better supported. Actually I'm surprised with the current layout.


jmwright (approved):
Thanks @adam-urbanczyk

PR #1151, opened 2022-09-12
Improve style of quick start docs

jdegenstein:
Remove line-break slashes and replace with parentheses per psf/black formatter guidelines. This also makes the quick start page more consistent with the other examples that are already in the documentation.


adam-urbanczyk:
Looks good, thanks @jdegenstein! 

(doc fix - merging without review)

PR #1156, opened 2022-09-16
OCCT7.6

adam-urbanczyk:
By popular request - CQ with OCCT 7.6


lorenzncode (approved):
LGTM!  Since this is the first OCP update since pip support was added, the procedure is probably not very well defined and automated.

For pip support, should OCP update PRs like this include a change to setup.py to pin say cadquery-ocp=7.6?  The new cadquery-ocp package would have to be prepared ahead of time.

After the new cadquery-ocp is uploaded, pip installs will start to break since there is no version pin currently.  Adding a version pin should improve the situation in the future.

Anyway, without some careful coordination there will likely be some pip install downtime/breakage.

jmwright:
The OCP releases on conda-forge went from 7.6.2-alpha to 7.6.3-alpha, with no releases in-between. Is 7.6.2 going to have a production release, or will it always be pre-release? Is 7.6.3 really alpha quality, or is the "alpha" just denoting that it's a first release?

https://anaconda.org/CadQuery/ocp/files

@lorenzncode 
> After the new cadquery-ocp is uploaded, pip installs will start to break since there is no version pin currently. Adding a version pin should improve the situation in the future.

I can start updating the ocp-build-system repo to 7.6.3 now to try to get it ready to publish to PyPi. It's always going to be hard to prevent builds from breaking during a small window of time though.

adam-urbanczyk:
> The OCP releases on conda-forge went from 7.6.2-alpha to 7.6.3-alpha, with no releases in-between. Is 7.6.2 going to have a production release, or will it always be pre-release? Is 7.6.3 really alpha quality, or is the "alpha" just denoting that it's a first release?
> 
> https://anaconda.org/CadQuery/ocp/files
> 
> @lorenzncode
> 
> > After the new cadquery-ocp is uploaded, pip installs will start to break since there is no version pin currently. Adding a version pin should improve the situation in the future.
> 
> I can start updating the ocp-build-system repo to 7.6.3 now to try to get it ready to publish to PyPi. It's always going to be hard to prevent builds from breaking during a small window of time though.

Unless you need it for something, I propose to forget about 7.6.2. I'll update setup.py to reflect the reality better.

adam-urbanczyk:
So what do you think @jmwright ?

jmwright:
> So what do you think @jmwright ?

The PR itself is fine, but we have run into a GLIBC version problem when building the Python wheels. The wheel builds use conda for the environment, but wheels don't bundle the system libraries like conda seems to.

https://github.com/CadQuery/ocp-build-system/issues/1

adam-urbanczyk:
Do you want to block this PR on that issue? OCP was compiled with gcc 8 AFAICT, I do not compile other libraries myself.

jmwright:
> Do you want to block this PR on that issue?

Not really, but we should at least take the version pin back out of setup.py since it will point to a non-existent package version until we figure out how to fix the wheel builds.

adam-urbanczyk:
> > Do you want to block this PR on that issue?
> 
> Not really, but we should at least take the version pin back out of setup.py since it will point to a non-existent package version until we figure out how to fix the wheel builds.

7.6 and 7.5 are not compatible. If you point to 7.5 in setup.py , you'll get errors when importing/running. IMO it is better to fail at install than at runtime.

fpq473:
I might suggest adding a 7.5.3 pin to setup.py right before committing these changes (which would update the pin).  That way, there is a commit that works in pip, regardless of what wheels have been built.

jmwright (approved):
With #1160 I think this is ready to merge @adam-urbanczyk

fpq473:
Please see #1161 for a fix to the version pin.

adam-urbanczyk:
After #1161 master is green, merging.

PR #1160, opened 2022-09-29
Pin cq-ocp to 7.5

adam-urbanczyk:
See #1156


jmwright (approved):
Thanks @adam-urbanczyk

Issue #1158, opened 2022-09-22
Wrong face orientation when exporting assemblies to VRML


mojca:
I'm not yet "proficient in interpreting VRML files", but it seems that the assemblies exported into VRML end up with wrong face orientation.

I first reported the issue to KiCad:
* https://gitlab.com/kicad/code/kicad/-/issues/12400

and prepared the files with a minimum example here:
* https://github.com/mojca/vrml-backface-culling-issue

```python
import cadquery as cq

box = cq.Workplane().box(2.0, 4.0, 1.0)
assy = cq.Assembly(box, loc=cq.Location((0,0,0)), color=cq.Color("black"))

cq.exporters.export(box, 'box-ok.wrl', cq.exporters.ExportTypes.VRML)
assy.save('box-has-back-face-culling-issues.wrl', exportType='VRML')
```

@jmwright

adam-urbanczyk:
Currently VRML export is based on VTK and it uses VTK-based normal calculation. In 7.6 it is possible to use the normals from OCCT. This line would have to be modified.

https://github.com/CadQuery/cadquery/blob/259cc7b4e47cf8da695085ab507963b57c60e046/cadquery/occ_impl/assembly.py#L194

So long story short - let's wait till 7.6 is merged and try to fix it there.

mojca:
If I understand correctly: someone needs to finalize support for Open CASCADE version 7.6.0 in CadQuery and then you'll be able to test an alternative approach to export to VRML and see if it works better? (It's not totally unambiguous what version 7.6 refers to.) Thank you for checking this out.

adam-urbanczyk:
Indeed, [OCCT/OCP] 7.6. See: https://github.com/CadQuery/cadquery/pull/1156

adam-urbanczyk:
@mojca could you check if #1167 solves this issue?

mojca:
What is the best way to install CadQuery in a proper way, so that I will be able to test that change? (I would prefer to stick with `pip` rather than `conda`, if possible. Or I can manually edit some files if needed.)

adam-urbanczyk:
This or clone the repo and use `pip install -e .`. You might want to ignore the deps if you already installed them.

mojca:
Thank you, that seems to work fine (both the installation and rendering a complex model).

mojca:
This leaves two remaining "issues" related to KiCad:
[...]

PR #1167, opened 2022-10-04
Do not recompute normals when converting to vtk

adam-urbanczyk:


mojca:
Thank you very much, this seems to have solved the problem reported in #1158.

lorenzncode (approved):
'box-has-back-face-culling-issues.wrl'  before fix (here in blender with WeightedNormal):
image

with this PR:
image

jmwright (approved):
Thanks @adam-urbanczyk Looks good.

PR #1172, opened 2022-10-07
Possibly better edge traversal

### LINKED ISSUE #1169: optimise files exported 2d files
Here is a basic script that produce a rounded box:

```py
import cadquery as cq

result = ( cq
    .Workplane('XY')
    .box(50, 50, 1)
    .edges('|Z')
    .fillet(10)
)

exporters.export(result, 'test.svg', opt={
    'width': 300,
    'height': 300,
    'marginLeft': 10,
    'marginTop': 10,
    'showAxes': False,
    'projectionDir': (0, 0, 1),
    'strokeWidth': 0.1,
    'strokeColor': (0, 0, 0),
    'hiddenColor': (0, 0, 255),
    'showHidden': False
})
```

Only one path is necessary for 2d export, but if we look at the source, we can see that many paths are created:

```xml
       <g  stroke="rgb(0,0,0)" fill="none">
			<path d="..." />
			<path d="..." />
			<path d="..." />
			<path d="..." />
			<path d="..." />
			<path d="..." />
			<path d="..." />
			<path d="..." />
       </g>
```

This can be very annoying when using a cnc or laser cutter, because the tool will make a lot of additional useless movements and wille take much more time to cut the piece.

Here is a simulation of what a machine will do on the exported svg (svg moves are replaced with lines):
image

adam-urbanczyk:
Maybe closes #1169

### DIFF
--- a/cadquery/occ_impl/exporters/dxf.py
+++ b/cadquery/occ_impl/exporters/dxf.py
@@ -120,9 +120,10 @@ def exportDXF(w: Workplane, fname: str):
     dxf = ezdxf.new()
     msp = dxf.modelspace()
 
-    for e in shape.Edges():
+    for w in shape.Wires():
+        for e in w.Edges():
 
-        conv = DXF_CONVERTERS.get(e.geomType(), _dxf_spline)
-        conv(e, msp, plane)
+            conv = DXF_CONVERTERS.get(e.geomType(), _dxf_spline)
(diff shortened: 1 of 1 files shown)


roipoussiere:
I just tested and I can't see any difference. Here is a before-after:
image

adam-urbanczyk:
Not solving the issue - closing.

PR #1198, opened 2022-11-24
Option to control verbosity of assembly solver

### LINKED ISSUE #1191: Provide option to control verbosity of assembly solver
Let's open another issue for this. TBH I was thinking more in the direction of removing it in the near feature (once the solver is more established.

_Originally posted by @adam-urbanczyk in https://github.com/CadQuery/cadquery/pull/1063#discussion_r874414165_

sethfischer:
Proposed method signature for `cadquery.assembly.Assembly.solve`:

```python
def solve(self, verbosity: int = 0) -> "Assembly":
```

When verbosity == 0 there is no output. Values of 1 and greater have increasing levels of verbosity printed to stdout.

Closes #1191

### DIFF
--- a/cadquery/assembly.py
+++ b/cadquery/assembly.py
@@ -346,7 +346,7 @@ class Assembly(object):
 
         return self
 
-    def solve(self) -> "Assembly":
+    def solve(self, verbosity: int = 0) -> "Assembly":
         """
         Solve the constraints.
         """
@@ -412,7 +412,7 @@ class Assembly(object):
         solver = ConstraintSolver(locs, constraints, locked=locked, scale=scale)
 
         # solve
-        locs_new, self._solve_result = solver.solve()
+        locs_new, self._solve_result = solver.solve(verbosity)
 
         # update positions
 
--- a/cadquery/occ_impl/solver.py
+++ b/cadquery/occ_impl/solver.py
@@ -659,7 +659,7 @@ class ConstraintSolver(object):
 
     def solve(self, verbosity: int = 0) -> Tuple[List[Location], Dict[str, Any]]:
 
-        suppress_banner = ("yes" if verbosity == 0 else "no")
+        suppress_banner = "yes" if verbosity == 0 else "no"
 
         opti = self.opti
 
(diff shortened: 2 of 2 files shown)


Jojain:
I would suggest adding an Enum rather than just a raw int. It's easier to use.

sethfischer:
> I would suggest adding an Enum rather than just a raw int. It's easier to use.

I'm not able to find a description of the verbosity levels (0-5).

From the [casadi documentation](https://casadi.sourceforge.net/v2.0.0/api/html/d6/d07/classcasadi_1_1NlpSolver.html):

|Id |Type|Default|Description|
|---|---|---|---|
|print_level|OT_INTEGER|5|Output verbosity level. (see IPOPT documentation)|

From the [IPOPT documentation](https://coin-or.github.io/Ipopt/hsl__ma97s_8h_source.html#l00082):

```cpp
int print_level;          /* <0 for no printing, 0 for basic, >1 for most */
```
is the best I can find. I expect the above is intended to be:

```cpp
int print_level;          /* <0 for no printing, 1 for basic, >1 for most */
```
However, 1 and 5 (default) do not result in the same verbosity.

adam-urbanczyk:
No worries, int is fine given the API of IPOPT and actually IMO easier to use (not necessarily more beautiful).

adam-urbanczyk (approved):

adam-urbanczyk:
NB: RTD failure fixed in master

lorenzncode (approved):
I tested the branch and did not find any problem.  verbosity values in range 0 to 12 are accepted.

If a test is needed, perhaps something like the following could be added later:
```py
@pytest.mark.parametrize(
    "verbosity, output", [(0, None), (3, "Optimal Solution Found"),],
)
def test_verbosity(verbosity, output, simple_assy2, capsys):

    assy = simple_assy2

    assy.constrain("b1", "b2", "Point", 1)

    assy.solve(verbosity=verbosity)

    if verbosity == 0:
        assert len(capsys.readouterr().out) == 0
    else:
        hasmatch = False
        for line in capsys.readouterr().out.split("\n"):
            if line.find(output) > 0:
                hasmatch = True
                break
        assert hasmatch

```

adam-urbanczyk:
Thanks @sethfischer !

PR #1201, opened 2022-12-01
show deprecation warning only when required

hiaselhans:
Show the deprecation warning (used in workplane.text) only when the argument is actually used in the function call.


adam-urbanczyk (approved):

jmwright (approved):
Thanks @hiaselhans

PR #1176, opened 2022-10-09
Fix typo

tkoyama010:
Fix a small typo.

PR #1202, opened 2022-12-02
Resolve ezdxf.math.BSpline mypy issue

lorenzncode:
- Fix mypy issue Argument 4 to "BSpline" has incompatible type "Optional[List[float]]"; expected "Iterable[float]"
- Exclude ezdxf version 1.0 until it can be tested


adam-urbanczyk:
This is fixed in master: https://github.com/mozman/ezdxf/blob/f1906eca3ca1211002d89b3990fc08e109c940d4/src/ezdxf/math/bspline.py#L862

I added a workaround in #1176 

Background: https://mypy-lang.blogspot.com/2022/11/mypy-0990-released.html

PR #1207, opened 2022-12-19
Added --pre option to pip install line to make it explicit

jmwright:
Not having this option caused a problem on a Windows install because the `--pre` wasn't assumed by pip when installing from git.


jmwright:
Simple readme fix. Merging.

Issue #1159, opened 2022-09-27
FixedPoint constraint's example is a copy / paste of PointOnLine's


xxx:
Hi,

I'm working through the process of grokking constraints, and noticed that the example at https://cadquery.readthedocs.io/en/latest/assy.html#fixedpoint is a copy / paste of the one above it.

thanks!

PR #1157, opened 2022-09-18
Assembly export: do not add leaf component when shapes is empty (#993)

lorenzncode:

PR #1210, opened 2022-12-23
Update assembly constraint docs

lorenzncode:
* Correct FixedPoint constraint example
* Simplify Location usage in examples and format with black for consistency

### DIFF
--- a/doc/assy.rst
+++ b/doc/assy.rst
@@ -370,8 +370,6 @@ STEP can be loaded in all CAD tool, e.g. in FreeCAD and the XML be used in other
     door.save('door.step')
     door.save('door.xml')
     
-In the case of STEP colors are preserved but not transparency.
-
 ..  image:: _static/door_assy_freecad.png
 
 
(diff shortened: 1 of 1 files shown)


lorenzncode:
Fixes #1159.

I removed a note on STEP and transparency.  I find that STEP export includes transparency data so the note may be out of date.

jmwright:
I created a test assembly with transparency and exported to STEP. Then I imported that into the online Autodesk viewer because the version of FreeCAD I'm running doesn't seem to respect the color info. Seems to confirm that transparency works.

Screenshot from 2022-12-23 16-37-08

jmwright (approved):
Thanks @lorenzncode !

lorenzncode:
@jmwright Thanks for confirming transparency is working for you!  Regarding colors, PR #1157  attempts to resolve an issue with STEP export and color handling.  I've tested STEP export with the PR branch and several viewers, including FreeCAD and Autodesk, and did not find any problem.  You might try exporting the STEP again with the PR 1157 branch to see if it resolves the color issue when viewing in FreeCAD.

jmwright:
@lorenzncode I just tested in FreeCAD 0.20.1 and the color and alpha worked, but I had to turn off the STEP import option `Enable STEP compound merge` which is apparently on by default. If that setting is not turned off I get two default grey merged blocks.

Screenshot from 2022-12-23 18-00-57

adam-urbanczyk (approved):

PR #1211, opened 2022-12-23
fix incorrect coordinate system of glTF exports

uki-dev:
correctly exports glTF in right-handed +Y up coordinate space as defined in the [glTF specification](https://registry.khronos.org/glTF/specs/2.0/glTF-2.0.html#coordinate-system-and-units)

before:
before

after:
after

### DIFF
--- a/cadquery/occ_impl/assembly.py
+++ b/cadquery/occ_impl/assembly.py
@@ -83,6 +83,10 @@ class AssemblyProtocol(Protocol):
     def loc(self) -> Location:
         ...
 
+    @loc.setter
+    def loc(self, value: Location) -> None:
+        ...
+
     @property
     def name(self) -> str:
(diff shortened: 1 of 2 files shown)


uki-dev:
`AssemblyProtocol.loc` is seemingly read-only , are there any other recommendations on how i could address this ?

adam-urbanczyk (approved):

jmwright (approved):
Thanks for the contribution @uki-dev !

lorenzncode (approved):
I tested the change - importing to Blender, CAD Assistant, Mayo, and PlayCanvas.  

All displayed the model in the desired orientation by default except for Mayo.  For Mayo, the GLTF Import option "System Coordinates Converter" can be changed from the default of "Undefined".

PR #1217, opened 2023-01-05
Fix VTK HTML template text-align value

sethfischer:
Remove "s" character causing invalid `text-align` value.


jmwright (approved):
Thanks @sethfischer !

adam-urbanczyk (approved):

PR #1226, opened 2023-01-12
Limit DXF BSpline Degree

### LINKED ISSUE #1225: DXF export of filleted splines is missing fillet arcs
When manually constructing splines, extruding, filleting the result, and then exporting them, the resultant DXF file does not render correctly in LibreCAD (or various online DXF processors like the one provided by SendCutSend or Ponoko).

```python
import cadquery as cq

s = cq.Sketch()
s = s.spline([
    (0, 0),
    (10, 10),
    (10, 0)
])
s = s.close()
s = s.assemble()

w = cq.Workplane("XY")
w = w.placeSketch(s)
w = w.extrude(5)
w = w.edges("|Z").fillet(2)

show_object(w)

cq.exporters.export(w, 'test.dxf')

```

Output:

image

bobtwinkles:
Most DXF renderers and processing tools can't handle BSplines with degree >= 3 (order >= 4). For maximum compatibility, we should approximate such BSplines using degree-3 splines. This change uses the OpenCascade facilities to do so, though ezdxf.math also provides some spline approximation facilities that could be used. Using the OpenCascade approach allows us to match FreeCAD's parameters which are presumably tuned on a diversity of real-world designs.

Fixes #1225

### DIFF
--- a/cadquery/occ_impl/exporters/dxf.py
+++ b/cadquery/occ_impl/exporters/dxf.py
@@ -119,7 +119,7 @@ def exportDXF(
     :param w: Workplane to be exported.
     :param fname: Output filename.
     :param approx: Approximation strategy. None means no approximation is applied.
-    "spline" results in all splines being approximated as cubic splines."arc" results
+    "spline" results in all splines being approximated as cubic splines. "arc" results
     in all curves being approximated as arcs and straight segments.
     :param tolerance: Approximation tolerance.
 
--- a/cadquery/occ_impl/exporters/dxf.py
+++ b/cadquery/occ_impl/exporters/dxf.py
@@ -1,31 +1,26 @@
-from ...cq import Workplane, Plane
+from ...cq import Workplane, Plane, Face
 from ...units import RAD2DEG
 from ..shapes import Edge
 from .utils import toCompound
 
 from OCP.gp import gp_Dir
-from OCP.GeomConvert import GeomConvert, GeomConvert_ApproxCurve
-from OCP.GeomAbs import GeomAbs_C0
+from OCP.GeomConvert import GeomConvert
 
-from typing import Dict, Any
+from typing import Optional, Literal
 
 import ezdxf
 
 CURVE_TOLERANCE = 1e-9
 
 
-def _dxf_line(
-    e: Edge, msp: ezdxf.layouts.Modelspace, plane: Plane
-):
+def _dxf_line(e: Edge, msp: ezdxf.layouts.Modelspace, plane: Plane):
 
     msp.add_line(
         e.startPoint().toTuple(), e.endPoint().toTuple(),
     )
 
 
-def _dxf_circle(
-    e: Edge, msp: ezdxf.layouts.Modelspace, plane: Plane
-):
+def _dxf_circle(e: Edge, msp: ezdxf.layouts.Modelspace, plane: Plane):
 
     geom = e._geomAdaptor()
     circ = geom.Circle()
@@ -53,9 +48,7 @@ def _dxf_circle(
         msp.add_arc((c.X(), c.Y(), c.Z()), r, a1, a2)
 
 
-def _dxf_ellipse(
-    e: Edge, msp: ezdxf.layouts.Modelspace, plane: Plane
-):
+def _dxf_ellipse(e: Edge, msp: ezdxf.layouts.Modelspace, plane: Plane):
 
     geom = e._geomAdaptor()
     ellipse = geom.Ellipse()
@@ -76,9 +69,7 @@ def _dxf_ellipse(
     )
 
 
-def _dxf_spline(
-    e: Edge, msp: ezdxf.layouts.Modelspace, plane: Plane
-):
+def _dxf_spline(e: Edge, msp: ezdxf.layouts.Modelspace, plane: Plane):
 
     adaptor = e._geomAdaptor()
     curve = GeomConvert.CurveToBSplineCurve_s(adaptor.Curve().Curve())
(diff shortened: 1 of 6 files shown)


bobtwinkles:
This ends up having a pretty severe impact on the accuracy of the exported DXFs, degrading the output quality by a few orders of magnitude. We could probably recover that by reducing the allowed approximation error in `GeomConvert_ApproxCurve` but it's unclear if the accuracy is critically important.

jmwright:
DXF is also used to generate CNC gcode, and the current behavior is non-approximated, so I don't know if this approximation should be the default.

bobtwinkles:
The tolerance specified is for 1 micrometer in a mm scale drawing, or 1 thou on an inch-scale drawing. I suppose the later is probably worth worrying about. If you have suggestions on how we should plumb through specifying whether or not to approximate, I'd be happy to implement that change.

jmwright:
@bobtwinkles I'd like to hear from others about this too, but my first thought was to leave the default at un-approximated and add an option somewhere to approximate splines during the export. However, that forces the CQ dev to know about the option and use it correctly, and end users of the DXF will be stuck if the dev chooses incorrectly.

I guess I don't really understand the scope of the problem. How many applications are not able import our DXF splines now versus how many will have that problem after approximation? Do we know if FreeCAD approximates splines during their DXF export? If they do it, we'd probably be safe to as well.

adam-urbanczyk:
Thanks for the PR @bobtwinkles . I agree with @jmwright , this has to be an option (e.g. `approximate : Optional[Literal['spline', 'arc']] = None`. I want to be able to use the current exact approach. 

You had to modify the tests significantly, so there is something wrong in the approximation. 1e-6 error cannot result in up to 50% change of the area. -1 for merging until this is fixed/understood.

BTW: you can look into `ShapeCustom_BSplineRestriction`.

bobtwinkles:
> However, that forces the CQ dev to know about the option and use it correctly, and end users of the DXF will be stuck if the dev chooses incorrectly.

Right, this is my concern as well. If I do implement an option I'll try to drop a note in the docs to point out this potential footgun which will hopefully help avoid the problem.

>  How many applications are not able import our DXF splines now versus how many will have that problem after approximation? Do we know if FreeCAD approximates splines during their DXF export?

Nothing I have to hand (LibreCAD and the Ponoko/SendCutSend online processors) can import the unapproximated DXFs correctly (except round-tripping to other OpenCascade-based software like cadquery and FreeCAD). My implementation is basically a direct port of [the FreeCAD 0.20 DXF export logic](https://github.com/FreeCAD/FreeCAD/blob/60e3084d424c35c50c857de4a17da25fd0ee5900/src/Mod/TechDraw/App/TechDrawExport.cpp#L640-L710). 

> You had to modify the tests significantly, so there is something wrong in the approximation. 1e-6 error cannot result in up to 50% change of the area. -1 for merging until this is fixed/understood.

I don't think the test changes indicate there's a 50% change in area? The changes are loosening the acceptable order of magnitude difference by 2-3 notches, not changing the value being compared with. All cases still require 1E-3 agreement or better, which matches the tolerance of the approximation (1E-3). Or really it's better, since most of the tests check area while the approximation is a tolerance on the wires.

> BTW: you can look into ShapeCustom_BSplineRestriction.

Looks like that [uses ApproxCurve internally](https://git.dev.opencascade.org/gitweb/?p=occt.git;a=blob;f=src/ShapeCustom/ShapeCustom_BSplineRestriction.cxx;h=ebab162cbe8abb8d22cf0804447e475bbd2cfc80;hb=HEAD#l1016), so I don't know if we would see any benefit from using the more generic interface.

adam-urbanczyk:
My bad, I confused the tolerance for the test value. I think you are making the problem bigger than it is, random online dxf viewer does display your shape correctly: https://sharecad.org/.

Regarding the function - I'd prefer that the conversion is handled at the `Shape` (a.k.a. topological level) so that you don't have to modify the individual geometry handling routines. Next user will want arc spline approximation...

bobtwinkles:
> Regarding the function - I'd prefer that the conversion is handled at the Shape (a.k.a. topological level) so that you don't have to modify the individual geometry handling routines.

Ah, I see. Do you mean we should add a method to `Workplane` or `Sketch` to wrap the current shape in a `ShapeCustom_BSplineRestriction`? That wouldn't be too bad to implement but would be a poor user experience IMO. `ShapeCustom_BSplineRestriction` will convert *all* curves of the underlying shape to BSplines including conics. The important case there being circles, which (again referring to the online fabrication quoting services I'm familiar with) will break "tappable hole" detection. That can itself be worked around by doing the approximation before forming any circular features, but that forces users to think about low-level representation details instead of the natural order of expression of their model. In the case of the part I was actually working on, I'd need to restructure a few dozen operations to make that work. For this reason I'd prefer doing this at the export layer, not while building the shape.

bobtwinkles:
I've updated this PR with plumbng to make the approximation optional, and some documentation about why it might be required. I ended up reusing the `opt` infrastructure that the SVG and VRML exporters use, though I didn't reuse the linear/angular specification because they don't mean exactly the same thing as they do for those exporters. Maybe it should though.

adam-urbanczyk (approved):

lorenzncode (commented):
I started testing `toSplines` and found a case where the degree was not restricted to the specified value (3).


```py
pts = [(0, 0), (10, 10), (20, -10)]
wp = cq.Workplane().splineApprox(pts, 1e-3, 3, 6)
```
This resulted in degree = 5.  After `toSplines` (with default args or with reduced tol) the degree was not reduced.

Original edge on left, with translated and restricted version on right (with display of poles).

image

Changing pts to the following resulted in degree = 6.
```pts = [(0, 0), (10, 10), (20, -10), (30, 10)]```

With `toSpline`, it resulted in degree = 5.
image

**Comment by adam-urbanczyk** [comment]
@lorenzncode  thanks! Using C0 or G2 seems to solve the issue.

File: cadquery/occ_impl/exporters/dxf.py [hunk 1]
```diff
@@ -105,12 +107,21 @@ def _dxf_spline(e: Edge, msp: ezdxf.layouts.Modelspace, plane: Plane):
...
+    tolerance: float = 1e-3,
+):
     """
     Export Workplane content to DXF. Works with 2D sections.
 
     :param w: Workplane to be exported.
-    :param fname: output filename.
+    :param fname: Output filename.
+    :param approx: Approximation strategy. None means no approximation is applied.
+    "spline" results in all splines being approximated as cubic splines."arc" results
```
jmwright, line 122:
```suggestion
    "spline" results in all splines being approximated as cubic splines. "arc" results
```

cadquery/occ_impl/shapes.py
```diff
@@ -2544,6 +2576,15 @@ def project(self, other: "Face", d: VectorLike) -> "Face":
 
         return self.constructOn(other, outer_p, *inner_p)
 
+    def toArcs(self, tolerance: float = 1e-3) -> "Face":
+        """
+        Approximate a planar faces arcs and straight line segments.
```
jmwright, line 2581:
```suggestion
        Approximate planar face arcs and straight line segments.
```

jmwright (approved):
Thanks @bobtwinkles !

doc/importexport.rst
```diff
@@ -231,6 +231,28 @@ optimum values that will produce an acceptable mesh.
 
     exporters.export(result, '/path/to/file/mesh.vrml', tolerance=0.01, angularTolerance=0.1)
 
+Exporting DXF
+##############
+
+By default, the DXF exporter will output splines exactly as they are represented by the OpenCascade kernel. Unfortunately some software can not handle higher-order splines, so if some of curves and arcs appear to be missing CadQuery can be instructed to approximate splines during the export process. This approximation is controlled by the following options:
```
lorenzncode, line 237:
```suggestion
By default, the DXF exporter will output splines exactly as they are represented by the OpenCascade kernel. Unfortunately some software cannot handle higher-order splines resulting in missing curves after DXF import. To resolve this, specify an approximation strategy controlled by the following options:
```

doc/importexport.rst
```diff
@@ -231,6 +231,28 @@ optimum values that will produce an acceptable mesh.
...
+* ``tolerance``: Acceptable error of the approximation, in the DXF's coordinate system. Defaults to 0.001 (1 thou for inch-scale drawings, 1 µm for mm-scale drawings).
+
+.. code-block:: python
+
+    import cadquery as cq
+    from cadquery import exporters
+
+    result = cq.Workplane().box(10, 10, 10)
+
+    exporters.exportDXF(
```
lorenzncode, line 249:
```suggestion

    cq.exporters.exportDXF(
```

box is not a good example here.  I think an example of creating a Workplane result object can be omitted  - just show the usage with approx.

lorenzncode (approved):
I tested the approx="arc" and did not find any problem.   Added minor doc suggestions.  Thanks @adam-urbanczyk and @bobtwinkles!

adam-urbanczyk:
Merging, thank you @bobtwinkles !

PR #1232, opened 2023-01-14
Update cadquery_cheatsheet.html

pbjacob:
'+Z' may select multiple faces, so its 'Objects Returned' should be '0..many'. Actually, the third example in this table, namely -X selector, is already correct result, which should be the same with +Z.


jmwright (approved):
Thanks for the contribution @pbjacob

jmwright:
Straight forward cheatsheet fix. Merging.

Issue #1228, opened 2023-01-13
Cannot create dev environment. OCP version


YHRen:
following the instruction `conda env create -n cq-dev -f environment.yml` and encountered this error:

```
Pip subprocess error:
ERROR: Ignored the following versions that require a different python version: 0.7 Requires-Python ~=2.7, >=3.4
ERROR: Could not find a version that satisfies the requirement cadquery-ocp<7.7,>=7.6 (from cadquery) (from versions: 7.5.3, 7.6.3a0)
ERROR: No matching distribution found for cadquery-ocp<7.7,>=7.6

failed

CondaEnvException: Pip failed
```

adam-urbanczyk:
Which conda version are you using? `pip` reqs of cq should be ignored:

https://github.com/CadQuery/cadquery/blob/369fcd95280631b1ae61ef6502ee49216ecd5cd4/environment.yml#L28

YHRen:
@adam-urbanczyk 

```
ᐅ conda -V
conda 22.11.1
```

```
ᐅ conda info

     active environment : base
            shell level : 1
          conda version : 22.11.1
    conda-build version : 3.22.0
         python version : 3.9.13.final.0
       virtual packages : __archspec=1=x86_64
                          __cuda=11.4=0
                          __glibc=2.31=0
                          __linux=5.4.0=0
                          __unix=0=0
  conda av metadata url : None
           channel URLs : https://repo.anaconda.com/pkgs/main/linux-64
                          https://repo.anaconda.com/pkgs/main/noarch
                          https://repo.anaconda.com/pkgs/r/linux-64
                          https://repo.anaconda.com/pkgs/r/noarch
               platform : linux-64
             user-agent : conda/22.11.1 requests/2.28.1 CPython/3.9.13 Linux/5.4.0-137-generic ubuntu/20.04.5 glibc/2.31
                UID:GID : 1000:1000
             netrc file : None
           offline mode : False
```

after commented out line 28, the error is the same. 

```
ᐅ conda env create -n cq-dev -f environment.yml
Collecting package metadata (repodata.json): done
Solving environment: done

Downloading and Extracting Packages

Preparing transaction: done
Verifying transaction: done
Executing transaction: done
Installing pip dependencies: \ Ran pip subprocess with arguments:
[...]

PR #1233, opened 2023-01-14
environment.yml cleanup

adam-urbanczyk:
Do not install deps when pip installing cadquery from a conda/mamba env

Solves #1228


jmwright (approved):
Looks good, thanks @adam-urbanczyk

Issue #1090, opened 2022-05-24
extrude in subtractive mode ignores both parameter


lorenzncode:
```
import cadquery as cq

w, l, h = 50, 20, 15
lsnap = 8
rsnap = 0.5

s = cq.Sketch().circle(rsnap)

b = (
    cq.Workplane()
    .box(w, l, h)
    .tag("box")
    .faces("<Y")
    .vertices("<X and >Z")
    .workplane(centerOption="CenterOfMass", offset=-l/2)
    .center(0, -2.0)
    .placeSketch(s)
    .extrude(lsnap, combine="a", both=True)
    .center(0, -3.0)
    .placeSketch(s)
    # subtractive, both is ignored
    .extrude(lsnap, combine="s", both=True)
)
```

YHRen:
encountered the same problem. `both` in extrude is ignored after `placeSketch`.

YHRen:
when `combine="s"`, it defers to "cutBlind" which ignores "both".

PR #1234, opened 2023-01-15
fix bug: extrude both direction and cut #1090

YHRen:
This RP tries to fix a bug #1090 that when `both=True` and `combine="cut"` used in `extrude`, the `both=True` is ignored. The simple fix is to add a `both` argument to `cutBlind`, which is called by `extrude` when `combine=cut` is used. A corresponding test is also added.

However, there is a subtle numerical issue (I believe) caused by flipping the norm direction:
https://github.com/CadQuery/cadquery/blob/a2c1f7c4a6176ef431c1dd84170d2caebef71e8e/cadquery/cq.py#L3648

This will result a fathom face between the two extruded objects, as shown below:
 
extrude_both_cut

This shouldn't be a problem for exporting to stl, but it is troubling nonetheless. Please feel free to make changes.

### DIFF
--- a/cadquery/cq.py
+++ b/cadquery/cq.py
@@ -3037,7 +3037,7 @@ class Workplane(object):
 
         # If subtractive mode is requested, use cutBlind
         if combine in ("cut", "s"):
-            return self.cutBlind(until, clean, taper)
+            return self.cutBlind(until, clean, both, taper)
 
         # Handle `until` multiple values
         elif until in ("next", "last") and combine in (True, "a"):
@@ -3162,7 +3162,7 @@ class Workplane(object):
             from warnings import warn
 
             warn(
-                "sweepAlongWires keyword argument is is deprecated and will "
+                "sweepAlongWires keyword argument is deprecated and will "
                 "be removed in the next version; use multisection instead",
                 DeprecationWarning,
             )
@@ -3440,6 +3440,7 @@ class Workplane(object):
         self: T,
         until: Union[float, Literal["next", "last"], Face],
         clean: bool = True,
+        both: bool = False,
         taper: Optional[float] = None,
     ) -> T:
         """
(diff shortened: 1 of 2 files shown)


adam-urbanczyk:
@YHRen would you be able to increase the coverage of the diff?

YHRen:
> @YHRen would you be able to increase the coverage of the diff?

@adam-urbanczyk Yes. Let me give it a try later tonight.

YHRen:
@adam-urbanczyk  Please take a look once you have time.

adam-urbanczyk (approved):
Thanks @YHRen . I added a glue step before cutting and now the spurious face seems to be gone.

cadquery/cq.py
```diff
@@ -3451,19 +3452,43 @@ def cutBlind(
         """
         # Handling of `until` passed values
         s: Union[Compound, Solid, Shape]
+        if isinstance(both, float) and taper == None:
```
lorenzncode, line 3455: Should the deprecation warning include a check for `int` type input?

lorenzncode (commented):
I'd suggest to add to docstring after clean (similar to extrude docstring):
```
        :param both: cut in both directions symmetrically
```

cadquery/cq.py
YHRen: @lorenzncode `int` will block boolean.

```
>>> isinstance(True, int)
True
```

YHRen:
> Thanks @YHRen . I added a glue step before cutting and now the spurious face seems to be gone.

This is very cool! I was no aware of `fuse` function. Thank you!

YHRen:
> 



> I'd suggest to add to docstring after clean (similar to extrude docstring):
> 
> ```
>         :param both: cut in both directions symmetrically
> ```

Added. Good point.

cadquery/cq.py
lorenzncode: The pytest also required an explicit conversion to float to trigger the warning which may not be typical of normal usage.

lorenzncode (approved):
I tested the branch and now extrude in subtractive mode (combine="s"), and with both=True works as expected.  cutBlind is backwards compatible when taper is provided as float value.  

Thanks @YHRen!

adam-urbanczyk:
Thanks again @YHRen !

YHRen:
@lorenzncode  @adam-urbanczyk  It's my pleasure. Thank you for maintaining cadquery.

Issue #1221, opened 2023-01-11
Example for revolve provides 3-tuples but links to revolve doc saying 2-tuples for axis of rotation


jabowery:
An example for revolve provides 3-tuples:
![image](https://user-images.githubusercontent.com/57646/211693973-b55a6df5-9de2-4854-a972-d395cb23ef4c.png)
but [the document says to use 2-tuples](https://cadquery.readthedocs.io/en/latest/classreference.html?highlight=revolve#cadquery.Workplane.revolve):

```
axisStart (tuple, a two tuple) – the start point of the axis of rotation

axisEnd (tuple, a two tuple) – the end point of the axis of rotation
```

lorenzncode:
Also see #1125.  In this case I would probably remove the type field in the docstring.

lorenzncode:
The docstrings/docs have been updated.  The type hints are displayed in both the signature and the parameter list.

PR #1251, opened 2023-01-27
Fix rst syntax errors

sethfischer:
Fix rst syntax errors that result in errors in the HTML docs, or generate Sphinx errors/warnings during sphinx build.

Update documented parameters for `Workplane.interpPlate` and `Solid.interpPlate` to match method definitions.

### DIFF
--- a/cadquery/cq.py
+++ b/cadquery/cq.py
@@ -154,7 +154,7 @@ class Workplane(object):
 
     .. note::
         You can also create workplanes on the surface of existing faces using
-        :py:meth:`CQ.workplane`
+        :meth:`workplane`
     """
 
     objects: List[CQObject]
@@ -227,8 +227,7 @@ class Workplane(object):
         Tags the current CQ object for later reference.
 
         :param name: the name to tag this object with
-        :type name: string
-        :returns: self, a cq object with tag applied
+        :returns: self, a CQ object with tag applied
         """
         self._tag = name
         self.ctx.tags[name] = self
@@ -279,14 +278,14 @@ class Workplane(object):
         """
         Splits a solid on the stack into two parts, optionally keeping the separate parts.
 
-        :param boolean keepTop: True to keep the top, False or None to discard it
-        :param boolean keepBottom: True to keep the bottom, False or None to discard it
+        :param bool keepTop: True to keep the top, False or None to discard it
+        :param bool keepBottom: True to keep the bottom, False or None to discard it
         :raises ValueError: if keepTop and keepBottom are both false.
         :raises ValueError: if there is no solid in the current stack or parent chain
         :returns: CQ object with the desired objects on the stack.
 
         The most common operation splits a solid and keeps one half. This sample creates
-        split bushing::
+        a split bushing::
 
             # drill a hole in the side
             c = Workplane().box(1,1,1).faces(">Z").workplane().circle(0.25).cutThruAll()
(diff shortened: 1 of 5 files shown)


sethfischer:
In this PR I've attempted to correct all remaining rst syntax errors. I identified errors by investigating Sphinx build errors and warnings, and also searched the HTML output for raw rst option names such as ":param" , ":type", and ":rtype".

lorenzncode:
I also started working on this.   So far I've added only shapes.py.  See #1126.  

Compare Face.makeNSidedSurface.

sethfischer:
@lorenzncode maybe your #1126 should be reviewed and merged first? The I'll rebase this of master and sort any merge conflicts.

lorenzncode:
I'll add the docstring changes I had planned for #1126 with this PR.  That way we will have a single PR with all the docstring fixes for review.

lorenzncode:
Fixes #922, fixes #1125, fixes #1221.

cadquery/cq.py
```diff
@@ -423,7 +422,7 @@ def vals(self) -> List[CQObject]:
         :rtype: list of occ_impl objects
         :returns: the values of the objects on the stack.
 
-        Contrast with :py:meth:`all`, which returns CQ objects for all of the items on the stack
+        Contrast with :meth:`all`, which returns CQ objects for all of the items on the stack
```
jmwright, line 425: These types of entries seem to work with `:py:`, so I'm curious what the motivation is to remove it. Was it causing some sort of error?

jmwright (approved):
Thanks @sethfischer That was a lot of work but it will definitely improve the docs.

adam-urbanczyk (approved):
Thanks!

adam-urbanczyk:
@jmwright I missed that you have some open comments - should I revert?

jmwright:
@adam-urbanczyk No, they're relatively minor and can be done in another PR, if needed.

cadquery/cq.py
lorenzncode: @jmwright You are correct, it's not an error to specify the domain here.  Since the default domain is already Python (py), I thought it would be simpler and less verbose to omit the domain.

cadquery/cq.py
```diff
@@ -2798,14 +2777,12 @@ def cboreHole(
...
-        :type diameter: float > 0
-        :param cboreDiameter: the diameter of the cbore
-        :type cboreDiameter: float > 0 and > diameter
+        :param cboreDiameter: the diameter of the cbore, must be greater than hole diameter
         :param cboreDepth: depth of the counterbore
         :type cboreDepth: float > 0
         :param depth: the depth of the hole
-        :type depth: float > 0 or None to drill thru the entire part.
-        :param boolean clean: call :py:meth:`clean` afterwards to have a clean shape
+        :type depth: float > 0 or None to drill thru the entire part
```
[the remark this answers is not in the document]
lorenzncode: I went back and forth on this one.  Initially I removed them, then decided to leave the depth related field descriptions as is.  Stating that a diameter must be positive I think is unnecessary.  Perhaps for the depth fields the additional description there will be useful.  It is a bit subjective.  The information could also be folded into the param part of the field.

I'm happy with the doc updates now.  Are the docstrings perfect?  No.  There are still inconsistencies in many areas such as punctuation, case, line breaks, etc.  This is understandable with the history of the project and different authors.   IMO, remaining issues are minor and the docs are in much better shape.

jmwright:
Based on the feedback from @lorenzncode I don't have any objections to leaving this merged PR as-is. Nothing more needs to be done.

Issue #1252, opened 2023-01-28
Version reported as 0.0.0 for dev versions when using conda or pip


lorenzncode:
```
>>> import cadquery
>>> cadquery.__version__
'0.0.0'
```

lorenzncode:
This is true at least for master.

Perhaps #1245 will correct the version for the release but then does it go back to 0.0.0 for master?

jmwright:
@lorenzncode Maybe it's related to the environment variable mentioned in #1085 ? Start [here](https://github.com/CadQuery/cadquery/pull/1085#issuecomment-1138674053) and ready the previous few comments to get the context.

lorenzncode:
@jmwright Thanks, I see it's set here with build variable `Build.SourceBranchName`.

https://github.com/CadQuery/conda-packages/blob/40f77ee43585a70e5958f6d7890e793337630ce1/conda-build.yml#L71

adam-urbanczyk:
This is not a "when installing with conda" issue, but rather "when installing a dev version" issue.

Release version for at least 2.2 reports `__version__` correctly.

PR #1255, opened 2023-01-31
Switching version info back to the in-between releases scheme

jmwright:
@adam-urbanczyk @lorenzncode When we're ready, this will switch the version numbers back to dev versions. We should also figure out how to address #1252 for these periods between releases, although I'm not sure I fully understand why conda is reporting `0.0.0`.


jmwright:
@lorenzncode @adam-urbanczyk I've tested the 2.2.0 conda package and things look fine. I'm going to go ahead and merge this so that we can start reviewing new PRs that will go into 2.3.0.

PR #1256, opened 2023-01-31
Fixed simple mistake in changelog that was brought up on Discord

jmwright:


jmwright:
Simple changelog fix that should not effect anything else. Merging.

PR #1260, opened 2023-02-03
Fix _entities

adam-urbanczyk:
Fix iteration over entities and avoid hash collisions. This should solve intermittent test failures.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -28,7 +28,7 @@ init:
 install:
     - mamba env create -f environment.yml
     - conda activate cadquery
-    - pip install git+https://github.com/CadQuery/OCP-stubs.git@7.6.3
+    - pip install git+https://github.com/CadQuery/OCP-stubs.git@7.7.0
 
 build: false
 
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -28,7 +28,6 @@ init:
 install:
     - mamba env create -f environment.yml
     - conda activate cadquery
-    - pip install git+https://github.com/CadQuery/OCP-stubs.git@7.7.0
 
 build: false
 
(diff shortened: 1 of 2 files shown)


cadquery/occ_impl/shapes.py
```diff
@@ -747,18 +751,21 @@ def ShapeType(self) -> Shapes:
...
 
         explorer = TopExp_Explorer(self.wrapped, inverse_shape_LUT[topo_type])
 
         while explorer.More():
             item = explorer.Current()
-            out[
-                item.HashCode(HASH_CODE_MAX)
-            ] = item  # needed to avoid pseudo-duplicate entities
+
+            # needed to avoid pseudo-duplicate entities
```
adam-urbanczyk, line 762: OCP does not wrap inner iterators (including `MapOfShape`), so this kind of construction is needed here.

appveyor.yml
```diff
@@ -28,7 +28,6 @@ init:
 install:
     - mamba env create -f environment.yml
     - conda activate cadquery
-    - pip install git+https://github.com/CadQuery/OCP-stubs.git@7.6.3
```
jmwright, line 31: The stubs really aren't needed anymore?

jmwright (approved):

appveyor.yml
adam-urbanczyk: They already specified in the environment file

jmwright: Ok, thanks.

lorenzncode (approved):
Given the test failures are intermittent I was unable to reproduce the issue with the old method to compare with the fix.  I tested the PR branch and did not find any problem with the new implementation.  Good find if this solves the intermittent test failures.

Issue #1249, opened 2023-01-25
Flatten STEP File Structure


jmwright:
The KiCAD project has existing STEP files that have a very flat structure. The following screenshot shows the structure when the file is imported.

![Screenshot from 2023-01-25 14-00-18](https://user-images.githubusercontent.com/1015439/214659539-92878142-ae97-4308-b7dc-d83fb093e442.png)

The reason they want this is because they can have full board models with thousands of components and they don't want any extra structure to deal with. Their STEPS do include the color, which is a requirement.

I've spent some time doing experiments with exporting to STEP, with some similarities to how we do it currently ([gist here](https://gist.github.com/jmwright/f541f2bf78697a1f157349ccc4a63351)). I have tried different ways of structuring the exported shapes including making a compound, and adding a shape and adding all the faces to that. No matter what I do the structure of the STEP looks like this when imported into FreeCAD.

![Screenshot from 2023-01-25 14-01-07](https://user-images.githubusercontent.com/1015439/214664323-46adb966-12e2-46ad-90c5-822356280967.png)

Am I missing an option somewhere to flatten the structure? Is there a way to accomplish this? Any ideas @adam-urbanczyk ?

jmwright:
The first screenshot only seems to be possible if the bounds of the assembly components are intersecting with one another. The original KiCAD 3D files were exported on FreeCAD, and (using FreeCAD) if I have the components intersect I can get the structure from the first screenshot. If they don't intersect, I get something similar to the structure in the second screenshot. I'm going to close this issue as it seems that the second screenshot will be as good as it's going to get with properly constrained assemblies.

adam-urbanczyk:
I'm not sure that I understand the request here. You want to get rid of the extra `boxes` and `origin` entities?

jmwright:
> Did you maybe mean single solid versus multiple solids?

Yes

>  If what FreeCAD does is the golden standard for this application, then simply take a look at the sources.

I have been. They appear to fuse the solids and then map the original faces to add the colors back in.

adam-urbanczyk:
> 
> I have been. They appear to fuse the solids and then map the original faces to add the colors back in.

Sounds reasonable, do you want to contribute this to CQ? It might be useful to other users.

adam-urbanczyk:
[...]

PR #1261, opened 2023-02-03
Added code for simplified and fused assembly export to STEP

jmwright:
This PR adds two new modes for STEP export.

~~1. Simplified STEP - Keeps the hierarchy to one level but will have multiple shapes, uses the names set for the assembly components as the shape names, and preserves colors.~~
2. Fused STEP - Will only have one shape when the solids in the assembly can be fused, otherwise will have a one-level hierarchy. Preserves colors by adding the faces of the assembly's solids as subshapes to the STEP document.


adam-urbanczyk:
TBH I think that the fused method should not use diff.

kliment:
Is there a way to implement color-preserving fuse without the cut?

adam-urbanczyk:
@kliment the first question is: why do you think you need a fuse to add colors to subshapes?

kliment:
The requirement that Jeremy's work is based on was for turning an assembly of different-colored parts into a single compound which preserves face colors. I believe a fuse is required for this. Obviously it's much easier to add colors without fusing, but this doesn't meet the single compound requirement.

kliment:
For reference, this requirement is coming from the KiCad library team, as we're looking for a replacement for the export method we use in FreeCad, which uses FreeCad's color-preserving fuse operation.

adam-urbanczyk:
As I wrote above, you can add colors to a compound and fusing is not needed to have a single compound... E.g. I can get this without fusing:
afbeelding

jmwright:
If someone can present a fuse-with-colors method that does not need the cut and also works reliably, I'm interested in it. I tried several things and this was the only method that would do what was needed. I may be missing something though as I have trouble following everything that FreeCAD is doing with their export.

jmwright:
@adam-urbanczyk Can you post the code you used to get the result in the screenshot above?

adam-urbanczyk:
There you go:
```python

from OCP.Quantity import Quantity_Color, Quantity_TOC_RGB
from OCP.TopoDS import TopoDS_Compound
from OCP.STEPControl import STEPControl_Writer, STEPControl_AsIs
from OCP.BRep import BRep_Tool, BRep_Builder
from OCP.TCollection import TCollection_ExtendedString
from OCP.TDocStd import TDocStd_Application, TDocStd_Document
from OCP.XCAFDoc import XCAFDoc_DocumentTool, XCAFDoc_ColorType, XCAFDoc_ColorGen
from OCP.STEPCAFControl import STEPCAFControl_Writer
from OCP.XSControl import XSControl_WorkSession
from OCP.TDataStd import TDataStd_Name
import cadquery as cq

# Build a simple assembly
assy = cq.Assembly()
box1 = cq.Workplane().box(10, 10, 10)
assy.add(box1, color=cq.Color(1, 0, 0))
box2 = cq.Workplane().center(0, 10).box(10, 10, 10)
assy.add(box2, color=cq.Color(0, 1, 0))

##############################
## EXPORT ####################
##############################
# The document
app = TDocStd_Application()
doc = TDocStd_Document(TCollection_ExtendedString("BinXCAF"))
app.InitDocument(doc)

# The shape and color tool
shape_tool = XCAFDoc_DocumentTool.ShapeTool_s(doc.Main())
color_tool = XCAFDoc_DocumentTool.ColorTool_s(doc.Main())


# Set up the compound
comp = cq.Compound.makeCompound(
    assy.toCompound().Solids()
)

# Compound assembly type
top_lbl = shape_tool.AddShape(comp.wrapped, False, True)

# NB I just assign random colors here, this should be based on a mapping to the original assy
for f in comp.Solids()[0].Faces():
    cur_lbl = shape_tool.AddSubShape(top_lbl, f.wrapped)
    color_tool.SetColor(cur_lbl, assy.children[1].color.wrapped, XCAFDoc_ColorGen)

# Export to STEP
session = XSControl_WorkSession()
writer = STEPCAFControl_Writer(session, False)
writer.Transfer(doc, STEPControl_AsIs)
status = writer.Write("xde_step_export_test.step")
```

jmwright:
@adam-urbanczyk I can only get your result when I check `Enable STEP Compound merge` within FreeCAD. Edit->Preferences->Import/Export->STEP (Tab). Otherwise I get this:

Screenshot from 2023-02-04 07-57-27

I don't think this is what we want because FreeCAD is artificially merging the compound and not showing the hierarchy that is actually in the file.

I think the only way to do this without a cut is to walk through all the faces of the compound and check if they are within the bounds of a face from the original assembly. I'm guessing OCCT has a way to check that, but I'm not sure what it is yet.

For reference, [here is an explanation](https://gitlab.com/kicad/libraries/kicad-packages3D-generator/-/issues/5#note_1226358351) of why this fused method is important to KiCAD users. And here is a [follow-up comment](https://gitlab.com/kicad/libraries/kicad-packages3D-generator/-/issues/5#note_1226473401) to that one.

adam-urbanczyk:
Fair enough, not using fuse is not an option, but it has nothing to do with colors.

On avoiding cut: If you want to map faces of the arguments of an OCCT operation to faces of the result, you can use the `Modified` method (see [here](https://dev.opencascade.org/doc/refman/html/class_b_rep_algo_a_p_i___builder_algo.html#a523cad14bd5e5104e2b683972a4e6bd8)).

jmwright:
@adam-urbanczyk @kliment There is still some clean-up to do, but the `toFused()` method now uses the history rather than the `cut` operation to handle the modified faces.

kliment:
Looks like the appveyor builds are failing because of a code style issue

cadquery/occ_impl/assembly.py
```diff
@@ -290,3 +294,145 @@ def toJSON(
...
+                # Check for modified faces
+                modded_list = fuse_op.Modified(face.wrapped)
+
+                # If there are no modified faces associated with this face, treat it as-is
+                if modded_list.Size() == 0:
+                    # Add the face as a subshape
+                    cur_lbl = shape_tool.AddSubShape(top_level_lbl, face.wrapped)
+
+                    # If we got a null label, it should be because the face is covered by IsDeleted or Generated
+                    if cur_lbl.IsNull() == True:
```
adam-urbanczyk, line 414: `== True` is not needed

cadquery/occ_impl/assembly.py
```diff
@@ -290,3 +294,145 @@ def toJSON(
...
+                        continue
+
+                    # Set the subshape's color to match the parent assembly component
+                    color_tool.SetColor(cur_lbl, part.color.wrapped, XCAFDoc_ColorGen)
+                # If there are modified faces based on this face, step through all of them and add
+                # them separately as subshapes
+                else:
+                    for mod in modded_list.__iter__():
```
adam-urbanczyk, line 422: `__iter__()` is not needed

cadquery/occ_impl/assembly.py
```diff
@@ -290,3 +294,145 @@ def toJSON(
...
+                        # Add the face as a subshape and set its color to match the parent assembly component
+                        cur_lbl = shape_tool.AddSubShape(top_level_lbl, mod)
+                        color_tool.SetColor(
+                            cur_lbl, part.color.wrapped, XCAFDoc_ColorGen
+                        )
+
+                # Handle generated faces
+                gen_list = fuse_op.Generated(face.wrapped)
+                if gen_list.Size() > 0:
+                    for gen in gen_list.__iter__():
```
adam-urbanczyk, line 432: `__iter__()` is not needed

cadquery/occ_impl/exporters/__init__.py
```diff
@@ -37,6 +37,19 @@ class ExportTypes:
 ]
 
 
+class STEPExportMode:
```
adam-urbanczyk, line 40: I'd propose to use literals and move this to the exporters/assy

cadquery/occ_impl/assembly.py
```diff
@@ -290,3 +294,145 @@ def toJSON(
         rv.extend(toJSON(child, loc, color, tolerance))
 
     return rv
+
+
+def toSimplified(assy: AssemblyProtocol, assy_name: str) -> TDocStd_Document:
```
adam-urbanczyk, line 299: `toSimplifiedCAF`?

cadquery/occ_impl/assembly.py
```diff
@@ -290,3 +294,145 @@ def toJSON(
...
+    for part in assy.children:
+        for solid in part.obj.solids().all():
+            cur_lbl = shape_tool.FindShape(solid.val().wrapped)
+            TDataStd_Name.Set_s(cur_lbl, TCollection_ExtendedString(part.name))
+            color_tool.SetColor(cur_lbl, part.color.wrapped, XCAFDoc_ColorGen)
+
+    return doc
+
+
+def toFused(assy: AssemblyProtocol, assy_name: str) -> TDocStd_Document:
```
adam-urbanczyk, line 341: `toFusedCAF`?

cadquery/occ_impl/assembly.py
```diff
@@ -290,3 +294,145 @@ def toJSON(
...
+    app = XCAFApp_Application.GetApplication_s()
+    doc = TDocStd_Document(TCollection_ExtendedString("XmlOcaf"))
+    app.InitDocument(doc)
+
+    # Shape and color tools
+    shape_tool = XCAFDoc_DocumentTool.ShapeTool_s(doc.Main())
+    color_tool = XCAFDoc_DocumentTool.ColorTool_s(doc.Main())
+    shape_tool.SetAutoNaming_s(False)
+
+    # Set up the compound
```
adam-urbanczyk, line 318: Better use the existing CQ API: `cq.Compound.makeCompound`?

jmwright:
@kliment Yeah, I'm still doing full runs of the KiCAD generator and tweaking `toFused()` as things come up. It's close though, and I'll do some clean up when it's ready.

cadquery/occ_impl/assembly.py
```diff
@@ -290,3 +294,145 @@ def toJSON(
...
+    comp = TopoDS_Compound()
+    comp_builder = BRep_Builder()
+    comp_builder.MakeCompound(comp)
+    for part in assy.children:
+        for solid in part.obj.solids().all():
```
adam-urbanczyk, line 323: I think obj is not always of type Workplane. Could you check with mypy by typing solid first? You might want to use `cq.Assemlby.shapes` here

cadquery/occ_impl/assembly.py
```diff
@@ -290,3 +294,145 @@ def toJSON(
...
+            # Add the current solid of the assembly component to the compound
+            comp_builder.Add(comp, solid.val().wrapped)
```
adam-urbanczyk, line 325: You are not applying the location of the parent here, is that correct?

cadquery/occ_impl/assembly.py
```diff
@@ -290,3 +294,145 @@ def toJSON(
         rv.extend(toJSON(child, loc, color, tolerance))
 
     return rv
+
+
+def toSimplified(assy: AssemblyProtocol, assy_name: str) -> TDocStd_Document:
+    """
+    Converts the assembly to a compound and saves that within the document
```
adam-urbanczyk, line 301: Is this function actually working with deeply nested assys?

adam-urbanczyk (commented):
First pass

cadquery/occ_impl/assembly.py
jmwright: Avoided this after refactor using `IsSubShape()`.

cadquery/occ_impl/exporters/__init__.py
jmwright: @adam-urbanczyk I'm not sure if you're thinking mypy literals or string literals. Is the latest commit what you were thinking? I'm not sure how to set the literal type except in the docstring either because kwargs is being used.

adam-urbanczyk: I guess both, e.g. https://github.com/CadQuery/cadquery/blob/6c8030d26a355b42362b64d502ad06b791e6bb30/cadquery/occ_impl/exporters/__init__.py#L43

cadquery/occ_impl/assembly.py
jmwright: No, it's not. I'll fix that before taking this out of draft.

jmwright:
Could somebody help me with the mypy errors? They don't make sense to me.

https://ci.appveyor.com/project/adam-urbanczyk/cadquery-icqj5/builds/46469815/job/ab07qy3oytpct7w4

adam-urbanczyk:
I'll take a look.

jmwright:
@adam-urbanczyk I think I have most of the mypy errors figured out, but it seems like I'm adding boilerplate code where it shouldn't be needed. I'll commit what I have once the errors are all fixed to see if they are the correct fixes.

cadquery/assembly.py
```diff
@@ -531,6 +542,18 @@ def toCompound(self) -> Compound:
 
         return Compound.makeCompound(shapes).locate(self.loc)
 
+    def toShapeList(self) -> List[Shape]:
+        """
+        Returns a list of all the shapes in the Assembly. This is most useful when exporting
+        the assembly to a file format such as STEP.
+        """
+
+        shapes = self.shapes
```
adam-urbanczyk, line 551: You are extending `self.shapes` - that is a bad idea.

cadquery/occ_impl/assembly.py
```diff
@@ -123,6 +128,9 @@ def children(self) -> Iterable["AssemblyProtocol"]:
     def traverse(self) -> Iterable[Tuple[str, "AssemblyProtocol"]]:
         ...
 
+    def toShapeList(self) -> Iterable[Shape]:
```
adam-urbanczyk, line 131: -> List[Shapes]

cadquery/occ_impl/assembly.py
```diff
@@ -289,3 +297,167 @@ def toJSON(
...
+    color_tool = XCAFDoc_DocumentTool.ColorTool_s(doc.Main())
+
+    # To fuse the parts of the assembly together
+    fuse_op = BRepAlgoAPI_Fuse()
+    args = TopTools_ListOfShape()
+    tools = TopTools_ListOfShape()
+
+    # If there is only one solid, there is no reason to fuse, and it will likely cause problems anyway
+    top_level_shape = None
+    children = cast(List, assy.children)
```
adam-urbanczyk, line 380: That cast should not be needed

adam-urbanczyk:
> @adam-urbanczyk I think I have most of the mypy errors figured out, but it seems like I'm adding boilerplate code where it shouldn't be needed. I'll commit what I have once the errors are all fixed to see if they are the correct fixes.

I added few comments.

cadquery/occ_impl/exporters/assembly.py
```diff
@@ -23,34 +25,63 @@
...
+    DEFAULT = "DEFAULT"
+    SIMPLIFIED = "SIMPLIFIED"
+    FUSED = "FUSED"
+
+
+STEPExportModeLiterals = Literal["DEFAULT", "SIMPLIFIED", "FUSED"]
+
+
+def exportAssembly(
+    assy: AssemblyProtocol, path: str, exportMode: str = "DEFAULT", **kwargs
```
adam-urbanczyk, line 42: exportMode should be typed using Literal

cadquery/assembly.py
```diff
@@ -451,6 +453,15 @@ def save(
             See :meth:`~cadquery.occ_impl.exporters.assembly.exportAssembly`.
         """
 
+        # Handle the export mode setting
```
adam-urbanczyk, line 456: This looks like too much boilerplate and the cast is not needed

cadquery/assembly.py
```diff
@@ -436,6 +437,7 @@ def save(
         self,
         path: str,
         exportType: Optional[ExportLiterals] = None,
+        exportMode: Optional[STEPExportModeLiterals] = None,
```
adam-urbanczyk, line 440: No need to make this optional, just set a backward compatible default

adam-urbanczyk:
@jmwright high level question: what is the use case for (1)? Do you see yourself actually using it?

cadquery/occ_impl/assembly.py
jmwright: Without it mypy shows the following errors.
```
cadquery/occ_impl/assembly.py:381: error: Value of type "Iterable[AssemblyProtocol]" is not indexable  [index]
cadquery/occ_impl/assembly.py:382: error: Value of type "Iterable[AssemblyProtocol]" is not indexable  [index]
cadquery/occ_impl/assembly.py:385: error: Value of type "Iterable[AssemblyProtocol]" is not indexable  [index]
```

cadquery/assembly.py
jmwright: It's modeled after the `exportType` code directly beneath it.

cadquery/assembly.py
jmwright: Ok, this one simplifies the fix for the comment above a lot. I'll make that change.

jmwright:
> high level question: what is the use case for (1)? Do you see yourself actually using it?

The simplified export has lost a lot of its novelty now that #1270 has been merged and I started using the loc for the compound, which causes additional nesting. This simplified method can probably be removed.

jmwright:
@adam-urbanczyk @lorenzncode I think this is ready for final review. The only remaining comment from above that is not resolved is the one about the cast. I cannot remove that cast because I get errors, and would need to know how to fix my code to avoid that.

lorenzncode (commented):
@jmwright  Shapes added to the top level do not seem to be supported.  Can you check?  

For example try removing the add call here in this test:
https://github.com/CadQuery/cadquery/blob/4f5c4f21a5e9e11b65899c2af45faad72c40f961/tests/test_exporters.py#L107

Instead add the `body` in the top level.  Then body is missing in the exported step file.

jmwright:
@lorenzncode I added handling for the case you mentioned, and then refactored the code because the logic was pretty messy. I'm not happy with the double casts that were required for mypy, but I'm not sure how to fix it.

https://github.com/CadQuery/cadquery/pull/1261/files#diff-9fa6458810e4d5a0d17f07210d50976638d18a5bfdecc3fc4194be5bef183eb1R342

lorenzncode:
Thanks @jmwright!  I find some of the assembly fixtures that were added with #1157 are not working with this new export mode.   Let me see if I can help fix these cases.

adam-urbanczyk (commented):
Full review still to come

adam-urbanczyk:
> Thanks @jmwright! I find some of the assembly fixtures that were added with #1157 are not working with this new export mode. Let me see if I can help fix these cases.

Maybe parametrize the tests with respect to this option?

cadquery/assembly.py
```diff
@@ -451,6 +454,12 @@ def save(
             See :meth:`~cadquery.occ_impl.exporters.assembly.exportAssembly`.
         """
 
+        # Make sure the export mode setting is correct
+        if exportMode not in ["DEFAULT", "FUSED"]:
```
adam-urbanczyk, line 458:
```suggestion
        if exportMode not in typing.get_args(STEPExportModeLiterals):
```

cadquery/occ_impl/assembly.py
```diff
@@ -289,3 +301,121 @@ def toJSON(
...
+                else:
+                    tools.Append(shape.wrapped)
+        else:
+            for shape in child.toShapeList():
+                tools.Append(shape.wrapped)
+
+    # If the tools are empty, it means we only had a single shape and do not need to fuse
+    if args.IsEmpty():
+        raise Exception(
+            "Error: A top level shape of assembly " + assy_name + " is not present."
```
adam-urbanczyk, line 360: I'd propose to use f strings.

cadquery/occ_impl/exporters/assembly.py
```diff
@@ -23,34 +25,62 @@
...
 from ..geom import Location
 
 
-def exportAssembly(assy: AssemblyProtocol, path: str, **kwargs) -> bool:
+class ExportModes:
+    DEFAULT = "DEFAULT"
+    FUSED = "FUSED"
+
+
+STEPExportModeLiterals = Literal["DEFAULT", "FUSED"]
```
adam-urbanczyk, line 37: To keep things more consistent, I'd propose to use lowercase for options that are not abbreviations.
```suggestion
STEPExportModeLiterals = Literal["default", "fused"]
```

cadquery/occ_impl/assembly.py
```diff
@@ -289,3 +301,121 @@ def toJSON(
...
 
     return rv
+
+
+def toFusedCAF(
+    assy: AssemblyProtocol,
+    assy_name: str,
+    glue: Optional[bool] = False,
+    tol: Optional[float] = None,
+) -> TDocStd_Document:
```
adam-urbanczyk, line 311: Maybe consider updating the signature to match toCAF. The other value is used in CQ-editor

cadquery/occ_impl/assembly.py
adam-urbanczyk: It is also not clear what you mean here. "Assembly contains no shapes" maybe?

cadquery/occ_impl/assembly.py
```diff
@@ -289,3 +301,121 @@ def toJSON(
...
+    doc = TDocStd_Document(TCollection_ExtendedString("XmlOcaf"))
+    app.InitDocument(doc)
+
+    # Shape and color tools
+    shape_tool = XCAFDoc_DocumentTool.ShapeTool_s(doc.Main())
+    color_tool = XCAFDoc_DocumentTool.ColorTool_s(doc.Main())
+
+    # To fuse the parts of the assembly together
+    fuse_op = BRepAlgoAPI_Fuse()
+    args = TopTools_ListOfShape()
```
adam-urbanczyk, line 332: You could start with a single list and then split it just before passing to fuse_op. This way quite some logic below could be removed.

cadquery/occ_impl/assembly.py
```diff
@@ -289,3 +301,121 @@ def toJSON(
...
+
+    # Walk the assembly->part->shape->face hierarchy and add subshapes for all the faces
+    for part in assy.children:
+        for shape in part.toShapeList():
+            for face in shape.Faces():
+                # Check for modified faces
+                modded_list = fuse_op.Modified(face.wrapped)
+
+                # If there are no modified faces associated with this face, treat it as-is
+                if modded_list.Size() == 0:
```
adam-urbanczyk, line 389: With latest OCP you can use
```suggestion
                if not modded_list:
```

cadquery/occ_impl/assembly.py
```diff
@@ -289,3 +301,121 @@ def toJSON(
...
+                        # Add the face as a subshape and set its color to match the parent assembly component
+                        cur_lbl = shape_tool.AddSubShape(top_level_lbl, mod)
+                        if part.color:
+                            color_tool.SetColor(
+                                cur_lbl, part.color.wrapped, XCAFDoc_ColorGen
+                            )
+
+                # Handle generated faces
+                gen_list = fuse_op.Generated(face.wrapped)
+                if gen_list.Size() > 0:
```
adam-urbanczyk, line 413:
```suggestion
                if gen_list:
```

cadquery/assembly.py
```diff
@@ -451,6 +454,12 @@ def save(
...
+            raise ValueError(
+                'Unknown assembly export mode "' + exportMode + '" for STEP'
```
adam-urbanczyk, line 460:
```suggestion
                f'Unknown assembly export mode {exportMode} for STEP'
```

cadquery/occ_impl/assembly.py
```diff
@@ -289,3 +301,121 @@ def toJSON(
...
+        fuse_op.Build()
+
+        top_level_shape = fuse_op.Shape()
+
+    # Add the fused shape as the top level object in the document
+    top_level_lbl = shape_tool.AddShape(top_level_shape, False)
+    TDataStd_Name.Set_s(top_level_lbl, TCollection_ExtendedString(assy_name))
+
+    # Walk the assembly->part->shape->face hierarchy and add subshapes for all the faces
+    for part in assy.children:
```
adam-urbanczyk, line 382: Could `toShapeList` be not needed?
```suggestion
    for _, part in assy.traverse():
       for shape in part.shapes:
           for face in shape.Faces(): 
```

adam-urbanczyk (commented):
If practical, could you also try to increase the diff coverage to 100%

adam-urbanczyk:
@jmwright I also cannot reproduce the failure locally

jmwright:
@adam-urbanczyk

> I also cannot reproduce the failure locally

Yes, it's odd. I checked to make sure that the CI's OCCT package wasn't updated recently, and it seems to be the same one that has been current for over a month now. I also tried running this branch against the KiCAD generator again hoping that one of the models would throw an error and highlight what was wrong, but none of them failed.

It will be a messy debugging session if I have to put print statements in, commit the code, then wait for CI to fail so I can check the logs.

adam-urbanczyk:
There was a way to log onto appveyor. I'll look it up.

wt., 21 mar 2023, 23:53 użytkownik Jeremy Wright ***@***.***>
napisał:

adam-urbanczyk:
@jmwright try adding this:
```
on_finish:
  - ps: $blockRdp = $true; iex ((new-object net.webclient).DownloadString('https://raw.githubusercontent.com/appveyor/ci/master/scripts/enable-rdp.ps1'))
#  - sh: export APPVEYOR_SSH_BLOCK=true
#  - sh: curl -sflL 'https://raw.githubusercontent.com/appveyor/ci/master/scripts/enable-ssh.sh' | bash -e -
```

adam-urbanczyk:
Not much progress, but so far I confirmed that for some reason the `AddSubShape` call fails in CI.

jmwright:
I've found out that the generated faces are not having the color set correctly anymore either. I have models from the 17th that worked fine, so it's something with this latest refactor. That's when the CI failures started too.  I think I'll start with restructuring how the modified and generated faces handling works.

jmwright:
@lorenzncode Can you try the current commit with the assembly fixtures that were failing?

lorenzncode:
@jmwright I still see the issue.  Try with the assembly defined in fixture `boxes0_assy`.  This results in two cubes with the default STEP export.  With the new option I see a single cube.

adam-urbanczyk:
@jmwright the code has one more problem - it does not apply locations at all. I'll try to come up with how to resolve this without too many changes.

adam-urbanczyk:
I tired to fix it. Some tests might have wrong asserts AFAICT.

jmwright:
The was a random, unrelated test failure in Windows on Python 3.9 that was fixed by re-running the pipeline.

adam-urbanczyk:
OK, I think all is fine now. @jmwright maybe take a look if I did not break something unexpected. @lorenzncode could you check if the issue you saw is solved?

lorenzncode:
Yes, the issue I saw with the `boxes0_assy` fixture is solved.  

I tried more cases and find the colors of the `chassis0_assy` fixture are incorrect.  To reproduce, manually test the code of the test_assembly.py fixture.  Export STEP in both the default and fused modes and compare the output.  The colors of the fused mode do not match the default mode.  There are faces that are expected to be colored but are missing color.

adam-urbanczyk:
Adding extra `copy()` seems to solve the issue.

jmwright:
@adam-urbanczyk I did a full run of the latest changes against the KiCAD generator and do not see any problems. I think this can be merged once @lorenzncode is happy with the fixture integration.

jmwright:
Closes #1249 

Related to #1266

jmwright:
The change in assembly naming and adding tests for colors sound good to me.

adam-urbanczyk:
OK, I think I fixed the single shape export. It seems OCCT treats Compounds differently when adding "subshapes" (faces). I don't fully understand why that is happaing, but it seems to be ralated to this line 
https://github.com/Open-Cascade-SAS/OCCT/blob/e01ce0cd52b37150679c833d1db2cc5edd620341/src/XCAFDoc/XCAFDoc_ShapeTool.cxx#L1110

@lorenzncode I think I'm done with the code. You can integrate your changes. Adding tests for subshape colors is a very good idea.

lorenzncode:
The following example is not working with the fused step export mode.  It seems to be an issue when a sphere intersects with another shape.

```py
import cadquery as cq

b0 = cq.Workplane().sphere(1)
b1 = cq.Workplane(origin=(1, 0, 0)).box(1, 1, 1)

assy = cq.Assembly()
assy.add(b0, color=cq.Color(1, 0, 0, 1))
assy.add(b1, color=cq.Color(0, 1, 0, 1))

fname = "out.step"
assy.save(
    path=fname,
    exportMode=cq.exporters.assembly.ExportModes.FUSED,
)

show_object(assy)
```

Here cq-editor is on the left and the STEP viewer on the right.  The STEP is missing the sphere shape.:

image

lorenzncode:
BREP format looks good.
```py
(b1 + b0).val().exportBrep('out1.brep')
```

image

After changing the origin of b1 in the previous example the STEP result is as expected:

```py
import cadquery as cq

b0 = cq.Workplane().sphere(1)
b1 = cq.Workplane(origin=(-1, 0, 0)).box(1, 1, 1)  # <-- origin modified

assy = cq.Assembly()
assy.add(b0, color=cq.Color(1, 0, 0, 1))
assy.add(b1, color=cq.Color(0, 1, 0, 1))

fname = "out2.step"
assy.save(
    path=fname,
    exportMode=cq.exporters.assembly.ExportModes.FUSED,
)
```
image

Perhaps the issue is related to fuse+STEP for some specific shapes rather than assys.

adam-urbanczyk:
I added #1312 to track this issue. If @jmwright agrees, I'd propose to consider it non-blocking for this PR.

jmwright:
> I'd propose to consider it non-blocking for this PR.

I agree.

lorenzncode:
Should the fuse op call `SimplifyResult()`?  It can potentially affect the size of the step file and the visualization.
Here two cylinders with and without SimplifyResult:
image

adam-urbanczyk:
IMO not, the purpose is to be able to find original faces.

adam-urbanczyk:
@lorenzncode what is the status here? Still adding some tests? AFAICT they are quite extensive as-is.

adam-urbanczyk:
> Agreed, the tests are probably sufficient to catch most regressions. I wanted to get a better idea of the performance. Experimenting with fuse op `SetRunParallel(True)` I did not find significant speedup for the case of intersecting shapes. Setting `glue=True` was much faster - consider enabling glue by default?

I'd say no, it should be generic by default.

cadquery/occ_impl/exporters/assembly.py
```diff
@@ -23,34 +25,62 @@
...
-
         If False, writes STEP file without pcurves. This decreases the size of the resulting STEP file.
     :type write_pcurves: bool
     :param precision_mode: Controls the uncertainty value for STEP entities. Specify -1, 0, or 1. Default 0.
         See OCCT documentation.
     :type precision_mode: int
+    :param fuzzy_tol: OCCT fuse operation tolerance setting used only for fused assembly export.
+    :type fuzzy_tol: float
+    :param glue: Glue setting used only for fused assembly export. Can be used to improve performance, but only for
+        non-overlapping shapes.
```
lorenzncode, line 64: I'd remove the statement on performance as I find improvement for both overlap and non-overlapping cases.

jmwright: Done

adam-urbanczyk: I don't think this is correct. AFAICT `glue=True` only works correctly for non-overlapping solids. See https://dev.opencascade.org/content/gluing-options-boolean-component

lorenzncode: In my (limited) testing I found speed up in both overlapping and non-overlapping cases.  Though I now see my experiment is a simple case like figure 4 of the article.  

```py
for i in range(1, N+1, 1):
    for j in range(1, N+1, 1):
        assy.add(
            box,
            name=f"box{i}_{j}",
            color=cq.Color("red"),
            #loc=cq.Location((1.0 * i, 1.0 * j, 0)),
            loc=cq.Location((0.9 * i, 0.9 * j, 0)),
        )
```

Without glue (N=50), fuse op time was ~300s total assy save time ~320s.  With glue=True, fuse op ~72s, total ~94s.  So based on this testing I made the comment about the docstring as it did not appear to be correct.

Now testing a case more like figure 5 I see the fuse op does not work with `glue=True`.


To provide guidance in the docstring (or elsewhere in docs) on when to specify glue, perhaps something like this is more accurate: 

```
    :param glue: Enable gluing mode for improved performance during fused assembly export.
        This option should only be used for non-intersecting shapes or those that are only touching or partially overlapping.
        Note that when glue is enabled, the resulting fused shape may be invalid if shapes are intersecting in an incompatible way.
        Defaults to False.
```

adam-urbanczyk: LGTM

adam-urbanczyk: I applied your suggestion @lorenzncode

cadquery/occ_impl/exporters/assembly.py
```diff
@@ -23,34 +25,62 @@
...
+
+
+STEPExportModeLiterals = Literal["default", "fused"]
+
+
+def exportAssembly(
+    assy: AssemblyProtocol,
+    path: str,
+    exportMode: STEPExportModeLiterals = "default",
+    **kwargs
```
adam-urbanczyk, line 44: Hm, actually why not list the prams explicitly in the signature. That is certainly more friendly for static analysis.

cadquery/occ_impl/exporters/__init__.py
jmwright: > Is this resolved? Maybe I misunderstood. I think the request is to remove the class `ExportModes'.

I'm not sure. I tried to match how it was done elsewhere in the code, but may have missed the mark.

doc/importexport.rst
```diff
@@ -76,8 +76,114 @@ There are no parameters for this method other than the file path to import.
...
+    box = cq.Workplane().box(10, 10, 10)
+
+    # Export the box
+    cq.exporters.export(box, "/path/to/step/box.stp", cq.exporters.ExportTypes.STEP)
+
+Setting Extra Options
+----------------------
+
+There are multiple options that can be set when exporting an object to a STEP file, but those can only be accessed by calling
+the :py:meth:`Shape.exportSTEP`` method on the lower level CadQuery object. For an explanation of the options available, see
```
adam-urbanczyk, line 114:
```suggestion
the :py:meth:`Shape.exportSTEP`` method on the lower level CadQuery object or the :py:meth:`Assembly.exportAssembly`` method. For an explanation of the options available, see
```

IS this what you mant @lorenzncode ?

cadquery/occ_impl/exporters/__init__.py
adam-urbanczyk: Yes, this is resolved

cadquery/occ_impl/exporters/assembly.py
adam-urbanczyk: Nevermind, too much scope creep

doc/importexport.rst
lorenzncode: @adam-urbanczyk Yes, partly.  I would suggest also to remove this statement: 
    "but those can only be accessed by calling the :py:meth:`Shape.exportSTEP`` method on the lower level CadQuery object."

It is incorrect, because you can do e.g.:

`assy.save('out.step', write_pcurves=False)`

doc/importexport.rst
```diff
@@ -76,8 +76,114 @@ There are no parameters for this method other than the file path to import.
...
+
+This section covers exporting CadQuery Workplane objects to STEP. For exporting assemblies to STEP, see the next section.
+
+Default
+--------
+
+The exporters module handles exporting Workplane objects to STEP. It is not necessary to set the export type explicitly
+since it will be determined from the file extension. Below is an example.
+
+.. code-block:: python
```
lorenzncode, line 90: An empty line must follow the start of each code-block to avoid errors like the following when building the docs: `ERROR: Error in "code-block" directive:
maximum 1 argument(s) allowed, 11 supplied.`

doc/importexport.rst
```diff
@@ -76,8 +76,114 @@ There are no parameters for this method other than the file path to import.
...
+
+    # Create a sample assembly
+    assy = cq.Assembly()
+    body = cq.Workplane().box(10, 10, 10)
+    assy.add(body, color=cq.Color(1, 0, 0), name="body")
+    pin = cq.Workplane().center(2, 2).cylinder(radius=2, height=20)
+    assy.add(pin, color=cq.Color(0, 1, 0), name="pin")
+
+    # Save the assembly to STEP
+    assy.save('/path/to/step/output/directory', exporters.ExportTypes.STEP)
```
lorenzncode, line 148: I believe a filename must be specifed.  This example would fail if only a dir was give.  

Suggest something like:
`assy.save('/path/to/step/output/directory/filename.step')`

doc/importexport.rst
```diff
@@ -76,8 +76,114 @@ There are no parameters for this method other than the file path to import.
...
+
+    # Create a sample assembly
+    assy = cq.Assembly()
+    body = cq.Workplane().box(10, 10, 10)
+    assy.add(body, color=cq.Color(1, 0, 0), name="body")
+    pin = cq.Workplane().center(2, 2).cylinder(radius=2, height=20)
+    assy.add(pin, color=cq.Color(0, 1, 0), name="pin")
+
+    # Save the assembly to STEP
+    assy.save('/path/to/step/output/directory', exporters.ExportTypes.STEP, export_mode=exporters.STEPExportMode.FUSED)
```
lorenzncode, line 171: Cannot save to a directory.

doc/importexport.rst
```diff
@@ -76,8 +76,114 @@ There are no parameters for this method other than the file path to import.
...
+
+Naming
+-------
+
+It is also possible to set the name of the top level assembly object in the STEP file with either the DEFAULT or FUSED methods.
+This is done by setting the name property of the assembly before calling :py:meth:`Assembly.save`.
+
+.. code-block:: python
+    assy = Assembly(name="my_assembly")
+    assy.save('/path/to/step/output/directory', exporters.ExportTypes.STEP, export_mode=exporters.STEPExportMode.FUSED)
```
lorenzncode, line 181: directory issue again.  

Also I notice none of the examples use literals.  Perhaps convert an examlpe to show use of literals.

assy.save("out.step", exportMode="fused")
and
assy.save("out.stp", "STEP")

doc/importexport.rst
adam-urbanczyk: OK, could you update yourself then?

lorenzncode: I'll add a few changes based on your latest commit.

lorenzncode:
What is the preferred argument name for the "export mode" of method `exportAssembly`?  Currently the name is "exportMode", where the doc examples use the name "export_mode".

lorenzncode:
For now, I've updated the doc examples to correct the export mode argument name (with regards to the current code).

jmwright:
I think `exportMode` matches the other parameters. I have been trying to only use the underscore naming convention within kwargs.

jmwright:
I made that change, and even though the tests still pass, it broke the fuse mode so that the setting seems to be ignored now. I may just revert the change and let you guys set it to what you think is best.

adam-urbanczyk:
Hm, AFAIR the tests check the number of resulting solids. How did you conclude that your change is not working?

jmwright:
The problem ended up being that I didn't get the parameter name changed everywhere in the KiCAD generator.

adam-urbanczyk:
OK, so are we good to merge this?

jmwright:
+1 to merge

lorenzncode:
I like the `mode` rename.  I did not find any other issues.

adam-urbanczyk:
Merged, thanks for all the hard work @jmwright  and @lorenzncode !

jmwright:
Thanks for all the feedback @adam-urbanczyk and @lorenzncode

PR #1270, opened 2023-02-11
Remove one level of hierarchy in STEP export

### LINKED ISSUE #1266: Eliminate level of hierarchy Assembly export
Assembly STEP export currently creates a root node with hardcoded name "CQ assembly". 

image

(from #1239 code with unsupported constraint removed)

This root node could be considered extraneous and be removed.
image

Regarding the root name,  I think the existing naming can be used here - the user specifies a name else it is auto generated uuid.

lorenzncode:
Do not hard code root name if unspecified

Resolves #1266

### DIFF
--- a/cadquery/occ_impl/assembly.py
+++ b/cadquery/occ_impl/assembly.py
@@ -10,6 +10,11 @@ from OCP.TDataStd import TDataStd_Name
 from OCP.TDF import TDF_Label
 from OCP.TopLoc import TopLoc_Location
 from OCP.Quantity import Quantity_ColorRGBA
+from OCP.XCAFPrs import (
+    XCAFPrs_DocumentExplorer,
+    XCAFPrs_DocumentExplorerFlags_None,
+    XCAFPrs_Style,
+)
 
 from vtkmodules.vtkRenderingCore import (
     vtkActor,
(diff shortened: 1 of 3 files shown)


cadquery/occ_impl/assembly.py
```diff
@@ -140,7 +140,7 @@ def toCAF(
     mesh: bool = False,
     tolerance: float = 1e-3,
     angularTolerance: float = 0.1,
-) -> Tuple[TDF_Label, TDocStd_Document]:
+) -> TDocStd_Document:
```
adam-urbanczyk, line 143: Why change the signature, you can still return the top level label

lorenzncode: I see.  I may not have  considered potential external usage of the function.   Internal usage was ignoring the label.  Yes, this change could be reverted to restore the original signature.

lorenzncode: oops - not to mention breaking cq-editor.

lorenzncode: I'm getting the label from XCAFPrs_DocumentExplorer, however, it could instead by set inside _toCAF if preferred.

adam-urbanczyk: I'd propose to return it from `_toCAF` and do not use any additional methods.

jmwright (approved):
Thanks @lorenzncode !

PR #1275, opened 2023-02-17
add extrustion distance to get sample code working

felix-otto:
Without the parameter I got the following error message: TypeError: extrude() missing 1 required positional argument: 'distance'


adam-urbanczyk:
Thanks @felix-otto , docs update - merging.

Issue #487, opened 2020-10-19
SVG output is forced into pixel dimensions


ArmoredBlood:
I'm working on revamping the SVG exporter so the SVG output is actually useful for CNC operations such as laser cutters, potters, and routers/mills.

Currently, the svg exporter outputs a fixed orientation with a static, hardcoded triad (origin marker). It also scales the output to fit inside hardcoded pixel dimensions, with the option for custom pixel dimensions.

My current plan (i.e. what I already have implemented) is to expose the basic "views" that are seen in cq-editor and freecad: front, back, left, right, top, bottom, isometric, and a "legacy" view which is the current hardcoded perspective.

The bigger question (the reason I'm opening this issue for discussion) is that I'd like to remove (or somehow optionalize) the pixel dimensions from the svg output and let it output actual dimensions to the file. That way, the user can immediately import the svg into cam software without any additional work.

I'm currently working out how to get around the scaling and to let the size of the model dictate the size of the resulting svg file.

I don't see the value in a non-dimensionally accurate, scaled down SVG of my model with a static triad that doesn't rotate with the object.

What are your thoughts on that?

dcowden:
Your approach sounds good to me. I am likely the author of the original
legacy view.

The objective of the SVG output at the time was in the very early days of
CQ, when the primary objective was to put a picture into the sphinx docs.
I used the docs as a way to visually check that the models were correct (
which, BTW, is still very valid-- a human an check pictures for accuracy
amazingly quickly )

Anyway, there's still a need to produce an SVG that is suitable for
pictorial display-- you're keeping the legacy view, so that'd handle that
requirement.

It might be cool if you can output svg with the right units, but then have
a parameter you can pass in that gives the size in px you want-- then it's
scaled. that, combined with selection of views as you have discussed, would
produce the legacy view without having another set of code still hanging
around.

thanks for contributing the work to make CNC-capable stuff, that's awesome!

ArmoredBlood:
That's a very valid use case that I hadn't thought of. In that case, I'll try to figure out a clean way to accomodate both processes.

dcowden:
@ArmoredBlood great!
 
FWIW, the [example docs](https://cadquery.readthedocs.io/en/latest/examples.html) for which i originally wrote the SVG view now use a much fancier interactive display, so i'm not sure anyone actually uses the legacy view any more

ArmoredBlood:
[...]

PR #1277, opened 2023-02-26
SVG export enhancements

### LINKED ISSUE #729: Make default projection dir consistent between svg export and cq-editor
The svg export and CQ-editor preview doesn't have the same direction direction, which is confusing.

Code example:

```py
import cadquery as cq
from cadquery.cq import Workplane as WP

fig = WP('YX').box(5, 5, 5).union(WP('YX').circle(1).extrude(15))

cq.exporters.export(fig, 'fig.svg')
show_object(fig)
```

Svg export and CQ-editor preview:

imageimage

### LINKED ISSUE #730: svg export: ability to define the camera position and zoom
Let's start with this code:

```py
import cadquery as cq
from cadquery.cq import Workplane as WP

def fig(wp):
    cube1 = WP(wp).box(5, 5, 5)
    rod1 = WP(wp).circle(1).extrude(15)
    return cube1.union(rod1)

svg_opts = {'height': 100, 'showHidden': False, 'showAxes': False}
cq.exporters.export(fig('XY'), 'fig1.svg', opt=svg_opts)
cq.exporters.export(fig('XZ'), 'fig2.svg', opt=svg_opts)
```

imageimage

These images share the same height. But what I need is to have the same box size and position. So instead of defining svg `width` and `height` in svg export options, I would like to set the camera position and zoom.

My current hack is to add a "background plate", like this:

```py
bg = WP('ZX').transformed(offset=V(-3, 0, -5)).box(1, 17, 25)`)
```

imageimage

Then remove this plate in Inkscape, but it's a little bit tedious.

lorenzncode:
Add option to set scaling based on model dimension instead of autoscale. 
Change default projection to match cq-editor.
Add option and set default "up" direction so it does not flip with changes to projectionDir. 
Remove margin from autoscale calculation when autoscale enabled.  This allows margins to be edited without affecting scale.

Resolves #729, resolves #730.

### DIFF
--- a/cadquery/occ_impl/exporters/svg.py
+++ b/cadquery/occ_impl/exporters/svg.py
@@ -1,7 +1,7 @@
 import io as StringIO
 
 from ..shapes import Shape, Compound, TOLERANCE
-from ..geom import BoundBox
+from ..geom import BoundBox, Vector
 
 
 from OCP.gp import gp_Ax2, gp_Pnt, gp_Dir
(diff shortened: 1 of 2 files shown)


jmwright:
It seems like this PR is also related to #1282 and #487

Jopie01:
Currently it is not supported to let the exporter decide the size of the image. you have set a height or width.

It would be nice to let the exporter decide the height and width based on the boundingbox and the margin. All based on a unitscale of 1 instead of the 0.75.

PR #1085, opened 2022-05-19
Updated setup.py with OCP being available on PyPi now

jmwright:
When I test locally this takes care of all dependencies, and gives users a couple options (development and IPython installs).

@dcowden Does this look good to you from your experience with setup.py for CQ 1.x?

@adam-urbanczyk Do you think not installing IPython by default will cause any problems? I've made IPython an optional install that the user has to specifically opt into. I did this because I believe that IPython is only used for jupyter_tools.py, which many users don't need. IPython brings its own set of dependencies which most users won't use.

PR #1286, opened 2023-03-14
Set version for unreleased packages

adam-urbanczyk:
Report version also in unreleased packages

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -36,11 +36,9 @@ if not is_rtd and not is_appveyor and not is_azure and not is_conda:
         "path",
     ]
 
-    setup_reqs = ["setuptools_scm"]
 
 setup(
     name="cadquery",
-    use_scm_version=True,
     version="2.3.0dev",  # Update this for the next release
     url="https://github.com/CadQuery/cadquery",
     license="Apache Public License 2.0",
--- a/setup.py
+++ b/setup.py
@@ -41,7 +41,7 @@ if not is_rtd and not is_appveyor and not is_azure and not is_conda:
 setup(
     name="cadquery",
     use_scm_version=True,
-    # version="2.3.0",  # Uncomment this for the next release
+    version="2.3.0dev",  # Update this for the next release
     url="https://github.com/CadQuery/cadquery",
     license="Apache Public License 2.0",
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -41,7 +41,7 @@
 setup(
     name="cadquery",
     use_scm_version=True,
-    # version="2.3.0",  # Uncomment this for the next release
+    version="2.3.0dev",  # Update this for the next release
```
jmwright, line 44: Does this override the automated SCM version?

jmwright: There was a suggestion in [this comment] (https://github.com/CadQuery/cadquery/pull/1085#issuecomment-1137708449) on how it might be possible to get the automatic SCM versioning working with Conda too.

adam-urbanczyk: For the release conda package I get the correct version. When installing from a local repo, I get 0.0.0. #1085 already mentions issues, so it is maybe better to just remove it.

lorenzncode: Does moving this line out of the if statement fix the SCM version issue?
https://github.com/CadQuery/cadquery/blob/2d355177871c0b1a275640659ba203cf3cdf267c/setup.py#L39

adam-urbanczyk: It does work with this addition and BTW `use_scm_version` does override `version`. I'd like to still get rid of all this magic and just report the version as 2.3dev. Do you see any practical added value in not removing it?

lorenzncode: > Do you see any practical added value in not removing it?

No, not really.  The simpler `2.3dev` should be good enough.

adam-urbanczyk: TBH I'd really like to get rid of this. If it is needed `git reflog` can be used. So far, I have never asked for exact commit. When reporting issues, you should reproduce first on the current master.

fpq473: Sounds reasonable, thanks for taking the time to explain!

lorenzncode (approved):

jmwright:
Is this ready to merge?

adam-urbanczyk:
> Is this ready to merge?

Yes, please.

jmwright (approved):

PR #1309, opened 2023-04-09
Update Shape.geomType __doc__

voneiden:
The description was based on  [CQ1 implementation](https://github.com/CadQuery/cadquery/blob/ceecbf690502625f91b9f1c0f96e5d2b8cd22835/cadquery/occ_impl/shapes.py#L422). The return type annotation seems to be correct.


adam-urbanczyk:
Thanks @voneiden !

PR #1317, opened 2023-04-20
exportSVG produces svgs with incorrect size.

giannissc:
The exportSVG function produced incorectly scaled svgs. Also the UnitOfMeasure is not something that should be guessed based on a heuristic but something that should be passed along by the user with the default being mm. The UnitScale has been set to 3.7795 as per the svg spec and creates svgs with the correct size.

### DIFF
--- a/cadquery/occ_impl/exporters/svg.py
+++ b/cadquery/occ_impl/exporters/svg.py
@@ -12,26 +12,22 @@ from OCP.GCPnts import GCPnts_QuasiUniformDeflection
 
 DISCRETIZATION_TOLERANCE = 1e-3
 
-SVG_TEMPLATE = """<?xml version="1.0" encoding="UTF-8" standalone="no"?>
-<svg
-   xmlns:svg="http://www.w3.org/2000/svg"
-   xmlns="http://www.w3.org/2000/svg"
-   width="%(width)s"
-   height="%(height)s"
-
->
-    <g transform="scale(%(unitScale)s, -%(unitScale)s)   translate(%(xTranslate)s,%(yTranslate)s)" stroke-width="%(strokeWidth)s"  fill="none">
-       <!-- hidden lines -->
-       <g  stroke="rgb(%(hiddenColor)s)" fill="none" stroke-dasharray="%(strokeWidth)s,%(strokeWidth)s" >
+SVG_TEMPLATE = """<?xml version='1.0' encoding='UTF-8' standalone='no'?>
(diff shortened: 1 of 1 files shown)


cadquery/occ_impl/exporters/svg.py
```diff
@@ -51,36 +46,14 @@
         -->
     </g>"""
 
-PATHTEMPLATE = '\t\t\t<path d="%s" />\n'
+PATHTEMPLATE = '   <g stroke-opacity="1" stroke-linejoin="bevel" font-style="normal" stroke-linecap="square" font-family="MS Shell Dlg 2" stroke="rgb(%(strokeColor)s)" fill="none" font-weight="400" transform="scale(%(unitScale)s, %(unitScale)s) translate(%(xTranslate)s,%(yTranslate)s)" stroke-width="%(strokeWidth)s" font-size="55.5625">\n    <path vector-effect="none" d="%(path)s" fill-rule="evenodd"/>\n   </g>\n'
```
voneiden, line 49: > `transform="scale(%(unitScale)s, %(unitScale)s)`

Y axis should (must?) be flipped here (like in the old implementation). CQ/OpenCascade has Y axis moving upwards whereas SVG has Y axis moving downwards. Without flipping it the resulting SVG will be a horizontal mirror of the actual projection.

cadquery/occ_impl/exporters/svg.py
```diff
@@ -236,12 +210,15 @@ def getSVG(shape, opts=None):
...
+    if uom == UNITS.MM:
+        unitScale = 3.7795
+    else:
+        unitScale = 91
 
     # compute amount to translate-- move the top left into view
     (xTranslate, yTranslate) = (
         (0 - bb.xmin) + marginLeft / unitScale,
-        (0 - bb.ymax) - marginTop / unitScale,
+        (0 + bb.ymax) + marginTop / unitScale,
```
voneiden, line 221: Shouldn't be necessary if the Y axis is flipped.  (ln 49)

voneiden (changes requested):
Took me a while to realize that the original implementation scales the projection to fit inside the dimensions of the SVG. This PR changes that behaviour so that the projection is scaled to 1:1. However, if we're doing 1:1 SVG's then IMO the dimensions of the SVG should be automatically calculated. 

With the current state of the PR, if the part is too big to fit the default dimensions it will just get cropped and the user needs to start guessing the right dimensions (in pixels!) to get it to fit.

However, before pouring more effort on this, you might want to get more comments/thoughts on the topic. Normally it's preferable to open an issue before opening a PR. ( https://github.com/CadQuery/cadquery/blob/master/README.md#contributing-code )

Issue #1323, opened 2023-04-25
[SVG] UnitOfMeasure should be an explicit parameter provided by the user


giannissc:
Also the UnitOfMeasure is not something that should be guessed based on a heuristic but something that should be passed along by the user with the default being mm.

For more information see: #1317

PR #1325, opened 2023-04-25
[SVG] The user should be able to provide the units of measurement for an svg

giannissc:
Closes: #1323

CC: @voneiden

### DIFF
--- a/cadquery/occ_impl/exporters/svg.py
+++ b/cadquery/occ_impl/exporters/svg.py
@@ -59,28 +59,6 @@ class UNITS:
     IN = "in"
 
 
-def guessUnitOfMeasure(shape):
-    """
-    Guess the unit of measure of a shape.
-    """
-    bb = BoundBox._fromTopoDS(shape.wrapped)
-
-    dimList = [bb.xlen, bb.ylen, bb.zlen]
-    # no real part would likely be bigger than 10 inches on any side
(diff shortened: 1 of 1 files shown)


jmwright:
How does this relate to [this existing draft PR](https://github.com/CadQuery/cadquery/pull/1277)?

giannissc:
I wasn't actually aware of that PR @jmwright

giannissc:
Since this is such a small and straightforward change should we get this merged? @voneiden, @jmwright

adam-urbanczyk:
@lorenzncode what do you think? Do you want to finish your PR first?

lorenzncode:
I find this PR breaks SVG export with `exporters.export`.  For example in the following width is not respected.

`cq.exporters.export(result, "out.svg", opt={"width":100})`

This PR adds a units parameter but I don't see where the units are put to use.  In #1317 it looks like the idea is to influence scaling.  

I'd also like to have optional control of the scale and that is one of the features of #1277.

giannissc:
Right now this has nothing to do with scaling. uom is used internally to set the scaling factor. What I changed here is the ability of the user to explicitly set the unit of measure instead of the current heuristic which works like this: <0.1 must be inches and >10 must be millimeter. I am preparing a separate PR for the scaling changing for which I will be taking a look at your PR to combine them @lorenzncode

PR #1332, opened 2023-05-06
Add tol to cut and intersect

adam-urbanczyk:
Adds fuzzy cut and intersect

### DIFF
--- a/cadquery/cq.py
+++ b/cadquery/cq.py
@@ -3331,13 +3331,17 @@ class Workplane(object):
         return self.union(toUnion)
 
     def cut(
-        self: T, toCut: Union["Workplane", Solid, Compound], clean: bool = True
+        self: T,
+        toCut: Union["Workplane", Solid, Compound],
+        clean: bool = True,
+        tol: Optional[float] = None,
     ) -> T:
         """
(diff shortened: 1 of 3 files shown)


jmwright:
One thing that I think is expected behavior but seems like it could confuse users, is that setting the tolerance too high will lead to no resulting geometry, with no error. The following code shows nothing in CQ-editor, but does not throw an error either.
```python
import cadquery as cq

box1 = cq.Workplane().box(10, 10, 10)
box2 = cq.Workplane().center(5, 5).box(10, 10, 10)

res = box2.intersect(box1, tol=3.0)

show_object(res)
```
That is a contrived example, but it is not hard to imagine that it will happen with lower tolerances with complex geometry. There have been a few models in the KiCAD library that had fuse glitches. In those cases the issue was solved by shifting parts and/or breaking sharp edges between geometries, but I think there is potential there for the same kinds of issues with `cut` and `intersect`. I don't have a solution except maybe to note that setting the tolerance too high may result in incorrect or missing geometry.

lorenzncode (approved):
LGTM.  I've done some manual testing and did not find any problem.

PR #1336, opened 2023-05-20
Fix sketch solver line_point method relative/absolute mixup

### LINKED ISSUE #1127: sketch_solver.line_point has incorrect implementation?
### Background
Tested out sketching with constraints and found myself in a situation where it seemed like my Distance constraint between a midpoint of a line and a midpoint of an arc was not respected.

```
rotary_enc_od = 6
result = (
    cq.Sketch()
        .segment((-1, 0), (1, 0), "seg")
        .arc((0, 0), rotary_enc_od/2, 0, 200, "arc")
        .constrain("arc", "FixedPoint", None)
        .constrain("arc", "Radius", rotary_enc_od/2)
        .constrain("arc", "seg", "Coincident", None)
        .constrain("seg", "arc", "Coincident", None)
        .constrain("seg", "Orientation", (1, 0))
        .constrain("arc", "seg", "Distance", (0.5, 0.5, 4.5))
        .solve()
        .assemble()
)
show_object(result)
```

Output 
image

Expected
image

Difference is subtle, but the current output produces a distance that is about 3.88 (measured from stl output) instead of the constrained 4.5. 

### Probable cause

After fiddling a bit with a debugger I noticed that line_point returned a rather odd value for the midpoint (-1.64, -0.85) instead of an expected (0, -0.56). 

https://github.com/CadQuery/cadquery/blob/master/cadquery/sketch.py#L924 sends lines/segments to the solver as a tuple (p1.x, p1.y, p2.x, p2.y) where p1 is start point and p2 is end point.

https://github.com/CadQuery/cadquery/blob/master/cadquery/occ_impl/sketch_solver.py#L85 appears to behave as if the above tuple is actually (p1.x, p1.y, d.x, d.y) where p1 is start point and d is a vector from the start point to the end point.

### Possible fix

I can get expected results if I change the implementation of line_point to 

```
def line_point(x, val):
    return x[:2] + val * (x[2:] - x[:2])
``` 

Thoughts?

voneiden:
Closes #1127

### DIFF
--- a/tests/test_sketch.py
+++ b/tests/test_sketch.py
@@ -640,7 +640,7 @@ def test_constraint_solver():
 
     midpoint = (seg2.startPoint() + seg2.endPoint()) / 2
 
-    (midpoint - seg1.startPoint()).Length == approx(2)
+    assert (midpoint - seg1.startPoint()).Length == approx(2)
 
     s5 = (
         Sketch()
--- a/cadquery/occ_impl/sketch_solver.py
+++ b/cadquery/occ_impl/sketch_solver.py
@@ -84,7 +84,7 @@ def arc_point(x, val):
 
 def line_point(x, val):
 
-    return x[:2] + val * x[2:]
+    return x[:2] + val * (x[2:] - x[:2])
 
 
 def arc_first_tangent(x):
--- a/environment.yml
+++ b/environment.yml
@@ -1,7 +1,7 @@
 name: cadquery
 channels:
-  - cadquery
   - conda-forge
+  - cadquery
 dependencies:
   - python>=3.8
   - ipython


tests/test_sketch.py
```diff
@@ -708,6 +708,39 @@ def test_constraint_solver():
...
+    s8.solve()
+
+    assert s8._solve_status["status"] == 4
+
+    s8.assemble()
+    assert s8._faces.isValid()
+
+    # 2x2=4
+    assert s8._faces.Area() == approx(4)
+
```
lorenzncode, line 743: This test may be redundant - can be removed.

lorenzncode (changes requested):
The fix itself looks good.

There is an assert missing here in the tests: 
https://github.com/CadQuery/cadquery/blob/7143eadb8e69bd9c3cecdf28b3cb77e7eaf974c0/tests/test_sketch.py#L643

I'd suggest to add the assert to line 643.  It should be sufficient for testing the fix.

```py
    assert (midpoint - seg1.startPoint()).Length == approx(2)
```

I've added a comment regarding the new pytest with sketch "s8".  Since the sketch solver is experimental, it might be easier to keep the tests simple (skip adding the new test) since rework may be required later.  @voneiden , @adam-urbanczyk , if you disagree and want to keep the additional test please disregard that item.

voneiden:
There we go, verified that the assert is indeed enough to test this fix.

lorenzncode (approved):
Thanks @voneiden!

voneiden:
Was that an accidental wrong branch for your commit 62d2a4c @adam-urbanczyk ?

adam-urbanczyk:
It was on purpose - trying to fix appveyor

adam-urbanczyk:
Green, thanks @voneiden !

Issue #1224, opened 2023-01-12
Sketch.hull() causes a ZeroDivisionError instead of some guidance around what specifically went wrong


alikureishy:
## To Reproduce

I was going through the Sketch tutorial, and playing around with hull(), and encountered this error. What am I doing wrong in this script below?

```python
cast: cq.Sketch = (
    cq.Sketch()
    .trapezoid(4, 3, 90)
    .vertices()
    .circle(.5, mode='s')
    .vertices()
    .circle(0.4, mode='a')
    .reset()
    .edges()
    .hull()          <----------- ZeroDivisionError
)
```

This is the model right *before* the ```hull()``` invocation above:
<img width="724" alt="Screenshot 2023-01-11 at 7 38 04 PM" src="https://user-images.githubusercontent.com/1193219/211970126-f158476f-86ff-4abc-bc55-88a03afbd2bd.png">

## Backtrace

```
ZeroDivisionError                         Traceback (most recent call last)
/var/folders/fs/jcstsp1j5fn8rz8bhfqxsglh0000gn/T/ipykernel_33196/1777645838.py in <cell line: 1>()
      8     .reset()
      9     .edges()
---> 10     .hull()
     11 )
     12 show(cast)

~/anaconda3/envs/cad/lib/python3.10/site-packages/cadquery/sketch.py in hull(self, mode, tag)
    502 
    503         if self._selection:
--> 504             rv = find_hull(el for el in self._selection if isinstance(el, Edge))
    505         elif self._faces:
    506             rv = find_hull(el for el in self._faces.Edges())

~/anaconda3/envs/cad/lib/python3.10/site-packages/cadquery/hull.py in find_hull(edges)
    393 
    394         for e in entities:
--> 395             angle, segment = get_angle(current_e, e)
    396             angles.append(angle if angle >= current_angle else inf)
    397             segments.append(segment)

~/anaconda3/envs/cad/lib/python3.10/site-packages/cadquery/hull.py in get_angle(current, e)
    310             return arc_pt(current, e)
    311         else:
--> 312             return arc_arc(current, e)
    313 
    314 

~/anaconda3/envs/cad/lib/python3.10/site-packages/cadquery/hull.py in arc_arc(a1, a2)
    245     elif r1 < r2:
    246         arc_tmp = Arc(a2.c, r2 - r1, a2.a1, a2.a2)
[...]

PR #1344, opened 2023-06-10
Fix some issues with Sketch.hull()

TerranceN:
I wanted to use hull for something I'm building, and immediately ran into this issue: https://github.com/CadQuery/cadquery/issues/1224, so I gave it a shot at fixing it.

I wouldn't put money on this being perfect / not having an issue I missed, but this worked for the few examples I tried, and for the project I'm working on.

### DIFF
--- a/cadquery/hull.py
+++ b/cadquery/hull.py
@@ -200,8 +200,7 @@ def _pt_arc(p: Point, a: Arc) -> Tuple[float, float, float, float]:
 
     return x1, y1, x2, y2
 
-
-def pt_arc(p: Point, a: Arc) -> Tuple[float, Segment]:
+def pt_arc(angle: float, p: Point, a: Arc) -> Tuple[float, Segment]:
 
     x, y = p.x, p.y
     x1, y1, x2, y2 = _pt_arc(p, a)
@@ -211,12 +210,12 @@ def pt_arc(p: Point, a: Arc) -> Tuple[float, Segment]:
 
     angles = atan2p(x1 - x, y1 - y), atan2p(x2 - x, y2 - y)
     points = Point(x1, y1), Point(x2, y2)
-    ix = int(argmin(angles))
+    ix = int(argmin([a + 2*pi if a < angle else a for a in angles]))
 
     return angles[ix], Segment(p, points[ix])
 
 
-def arc_pt(a: Arc, p: Point) -> Tuple[float, Segment]:
+def arc_pt(angle: float, a: Arc, p: Point) -> Tuple[float, Segment]:
 
     x, y = p.x, p.y
     x1, y1, x2, y2 = _pt_arc(p, a)
@@ -227,12 +226,12 @@ def arc_pt(a: Arc, p: Point) -> Tuple[float, Segment]:
     angles = atan2p(x - x1, y - y1), atan2p(x - x2, y - y2)
     points = Point(x1, y1), Point(x2, y2)
 
-    ix = int(argmax(angles))
+    ix = int(argmax([a + 2*pi if a < angle else a for a in angles]))
 
     return angles[ix], Segment(points[ix], p)
 
 
-def arc_arc(a1: Arc, a2: Arc) -> Tuple[float, Segment]:
+def arc_arc(angle: float, a1: Arc, a2: Arc) -> Tuple[float, Segment]:
 
     r1 = a1.r
     xc1, yc1 = a1.c.x, a1.c.y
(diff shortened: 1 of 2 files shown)


cadquery/hull.py
```diff
@@ -187,6 +189,9 @@ def _pt_arc(p: Point, a: Arc) -> Tuple[float, float, float, float]:
     xc, yc = a.c.x, a.c.y
     dx, dy = x - xc, y - yc
     l = sqrt(dx ** 2 + dy ** 2)
+    ld = l ** 2 - r ** 2
+    if (ld < 0.001):
+        return x, y, x, y
```
TerranceN, line 194: This is when the point is on the arc's circle, so I return a no-length segment (which is later ignored). The point might not necessarily be on the arc though, so this should return a segment to the start or end of the arc.

This is taken care of elsewhere because I add the start and end points of the arc to the list of points in `convert_and_validate`, so if the next segment is between the point and a start or end point of an arc, it's handled elsewhere as a `pt_pt` comparison.

cadquery/hull.py
```diff
@@ -195,33 +200,38 @@ def _pt_arc(p: Point, a: Arc) -> Tuple[float, float, float, float]:
...
     x1, y1, x2, y2 = _pt_arc(p, a)
 
+    if ((x1, y1) == (x2, y2)):
+        return inf, Segment(Point(inf, inf), Point(inf, inf))
+
     angles = atan2p(x - x1, y - y1), atan2p(x - x2, y - y2)
     points = Point(x1, y1), Point(x2, y2)
 
-    ix = int(argmax(angles))
+    ix = int(argmax([a + 2*pi if a < angle else a for a in angles]))
```
TerranceN, line 229: When the angle loops around back past 0 radians, this caused the argmin/argmax of `pt_arc` and `arc_pt` to choose the wrong side of the arc to continue from.

cadquery/hull.py
```diff
@@ -292,24 +302,36 @@ def arc_arc(a1: Arc, a2: Arc) -> Tuple[float, Segment]:
         Segment(Point(x12, y12), Point(x22, y22)),
     )
 
-    return angles[ix], segments[ix]
+    segment_angle = atan2p(segments[ix].b.x - xc2, segments[ix].b.y - yc2)
+    if (segment_angle < a2.a1 or segment_angle > a2.a2):
+        return inf, Segment(Point(inf, inf), Point(inf, inf))
```
TerranceN, line 307: I don't understand this function that well, so this might have been from something else wrong and is now redundant, but somehow it was selecting a point not actually on the arc, so this if statement throws that out if that happens.

I'd bet this isn't the best way to handle it and there's more edge cases that would fail here.

cadquery/hull.py
```diff
@@ -292,24 +302,36 @@ def arc_arc(a1: Arc, a2: Arc) -> Tuple[float, Segment]:
...
+    def center(e: Entity):
+        if isinstance(e, Point):
+            return e
+        return e.c
+    if center(current) == center(e):
+        if (isinstance(current, Arc) and isinstance(e, Arc)):
+            if (current.s == e.e and not e.s == current.e):
+                return pt_pt(current.e, e.s)
+            if (e.s == current.e and not current.s == e.e):
+                return (current.a2 + pi) % (2 * pi), Segment(Point(inf, inf), Point(inf, inf))
```
TerranceN, line 322: These changes allow a circle that's broken into multiple arcs continue around the circle

cadquery/hull.py
```diff
@@ -385,14 +409,17 @@ def find_hull(edges: Iterable[Edge]) -> Wire:
     current_angle = 0.0
     finished = False
 
+    iterations = 0
+
     # march around
-    while not finished:
+    while not finished and iterations < len(entities):
+        iterations += 1
```
TerranceN, line 416: While working on this it was very easy for this to loop infinitely. Setting a sensible maximum number of iterations (if we've iterated more than the number of entities, _something_ has gone wrong) seems better than having the system hang in an infinite loop.

tests/test_hull.py
```diff
@@ -18,7 +18,7 @@ def test_hull():
 
     h = hull.find_hull(edges)
 
-    assert len(h.Vertices()) == 11
+    assert len(h.Vertices()) == 10
```
TerranceN, line 21: Somehow my changes made this hull have one fewer vertices. Based on how the points and arcs are laid out, I'm guessing there were multiple possible next entities with the same angle and my hull changes made it skip one.

There should probably be a 'tie-breaker' for choosing the smallest angle so that the system is more deterministic. (either always choosing the closer to ensure the loop always ends up closed, or choosing the further, but we'd have to be more careful about choosing the starting entity, lest we start on an entity we later skip and fail to close the loop)

PR #1350, opened 2023-06-16
Svg update

jmwright:
PRs #1277 and #1325 are both open with SVG export changes, but neither one fixed this issue for me. There was an extra bottom-right margin that I could not remove in the resulting SVG, and I wanted my SVGs to fill the image edge to edge for my use case while still respecting the top/left margins. These changes accomplish that.

Below are before and after examples.

Before:
Screenshot from 2023-06-14 21-38-18

After:
Screenshot from 2023-06-14 21-36-17

### DIFF
--- a/cadquery/occ_impl/exporters/svg.py
+++ b/cadquery/occ_impl/exporters/svg.py
@@ -247,6 +247,12 @@ def getSVG(shape, opts=None):
         else:
             width = height * (bb.xlen / bb.ylen)
 
+        image_width = width + (marginLeft * 2.0)
+        image_height = height + (marginTop * 2.0)
+    else:
+        image_width = width
+        image_height = height
+
     # width pixels for x, height pixels for y
     unitScale = min(width / bb.xlen * bb_scale, height / bb.ylen * bb_scale)
 
@@ -290,8 +296,8 @@ def getSVG(shape, opts=None):
             "visibleContent": visibleContent,
             "xTranslate": str(xTranslate),
             "yTranslate": str(yTranslate),
-            "width": str(width),
-            "height": str(height),
+            "width": str(image_width),
+            "height": str(image_height),
             "textboxY": str(height - 30),
             "uom": str(uom),
             "axesIndicator": axesIndicator,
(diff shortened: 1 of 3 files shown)


adam-urbanczyk (approved):
Thanks, LGTM. I assume that the black border in the "after" picture is some artifact not coming from CQ.

jmwright:
> I assume that the black border in the "after" picture is some artifact not coming from CQ.

Correct. I must have gotten part of the frame of the app when I did the screenshot.

lorenzncode (commented):
I'd like to propose the following changes:

1. Explicitly choose to recompute width *or* height. 

   - With the current PR the user does not know ahead of time which dimension will be recomputed.

     I get surprised to find either one of width/height changed. 
     
   - It is more straightforward to specify a fixed width or height.  The user will likely have one in mind and does not need to specify both since the other is recomputed. 

     I can then say I want a width of 200 pixels and the height fits automatically.


2.  The current PR handling of margin is inconsistent with fitView=False.  I suggest to respect the width or height value given - do not change it according to margin (margin padded to inside of width and height).


Please see these changes here: 
https://github.com/lorenzncode/cadquery/commit/6aab3be7f46bcd5e2529724c2fb00c1e67c7d180


The fitView option was removed.  Instead, you can fit by specifying -1 for one of the dimensions.

This results in a width of 200, height is computed:
```py
width = 200
height = -1 
```

This results in a height of 100, width is computed:
```py
width = -1 
height = 100 
```

jmwright:
@lorenzncode I've made those changes (except for the documentation changes), but I'm not convinced that using `-1` for autofit is the best way. Maybe we should allow the user to leave out the other dimension from `opts` or set it to `None`.

adam-urbanczyk:
None sounds more pythonic

lorenzncode (approved):
Looks good, thanks @jmwright!

jmwright:
@adam-urbanczyk It looks like AppVeyor is grabbing the OCP 7.7.1 pre-release and it might be breaking the builds.

```
TypeError: GetColor(): incompatible function arguments. The following argument types are supported:
```

adam-urbanczyk:
Should be solved, I moved 7.7.1 to dev

adam-urbanczyk:
Hm, not quite. Seems that 7.7.1 is available from the conda-forge channel even though the PR is not merged.

PR #1353, opened 2023-06-20
Imprinting and matching

adam-urbanczyk:
- [x] Extract nested iterator
- [x] Add tests
- [x] Refactor toVTK to use the new iterator
- [x] toVTK rendering improvements -> split into triangles with normals and edges/points without 
- [x] Refactor other methods to use the new iterator

### DIFF
--- a/cadquery/assembly.py
+++ b/cadquery/assembly.py
@@ -543,6 +543,28 @@ class Assembly(object):
 
         return rv
 
+    def __iter__(
+        self,
+        loc: Optional[Location] = None,
+        name: Optional[str] = None,
+        color: Optional[Color] = None,
+    ) -> Iterator[Tuple[Shape, str, Location, Optional[Color]]]:
+        """
+        Assembly iterator yielding shapes, locatons, names and colors.
+        """
+
+        name = f"{name}/{self.name}" if name else self.name
+        loc = loc * self.loc if loc else self.loc
+        color = self.color if self.color else color
+
+        if self.obj:
+            yield self.obj if isinstance(self.obj, Shape) else Compound.makeCompound(
+                s for s in self.obj.vals() if isinstance(s, Shape)
+            ), name, loc, color
+
+        for ch in self.children:
+            yield from ch.__iter__(loc, name, color)
+
     def toCompound(self) -> Compound:
         """
         Returns a Compound made from this Assembly (including all children) with the
--- a/cadquery/assembly.py
+++ b/cadquery/assembly.py
@@ -550,7 +550,7 @@ class Assembly(object):
         color: Optional[Color] = None,
     ) -> Iterator[Tuple[Shape, str, Location, Optional[Color]]]:
         """
-        Assembly iterator yielding shapes, locatons, names and colors.
+        Assembly iterator yielding shapes, names, locations and colors.
         """
 
         name = f"{name}/{self.name}" if name else self.name
--- a/cadquery/occ_impl/assembly.py
+++ b/cadquery/occ_impl/assembly.py
@@ -246,23 +246,19 @@ def toCAF(
 
 def toVTK(
     assy: AssemblyProtocol,
-    renderer: vtkRenderer = vtkRenderer(),
-    loc: Location = Location(),
     color: Tuple[float, float, float, float] = (1.0, 1.0, 1.0, 1.0),
     tolerance: float = 1e-3,
     angularTolerance: float = 0.1,
 ) -> vtkRenderer:
 
-    loc = loc * assy.loc
-    trans, rot = loc.toTuple()
+    renderer = vtkRenderer()
 
-    if assy.color:
-        color = assy.color.toTuple()
+    for shape, _, loc, col_ in assy:
 
-    if assy.shapes:
-        data = Compound.makeCompound(assy.shapes).toVtkPolyData(
-            tolerance, angularTolerance
-        )
+        col = col_.toTuple() if col_ else color
+        trans, rot = loc.toTuple()
+
+        data = shape.toVtkPolyData(tolerance, angularTolerance)
 
         mapper = vtkMapper()
         mapper.SetInputData(data)
@@ -271,13 +267,10 @@ def toVTK(
         actor.SetMapper(mapper)
         actor.SetPosition(*trans)
         actor.SetOrientation(*map(degrees, rot))
-        actor.GetProperty().SetColor(*color[:3])
-        actor.GetProperty().SetOpacity(color[3])
+        actor.GetProperty().SetColor(*col[:3])
+        actor.GetProperty().SetOpacity(col[3])
 
         renderer.AddActor(actor)
-
-    for child in assy.children:
-        renderer = toVTK(child, renderer, loc, color, tolerance, angularTolerance)
 
     return renderer
 
--- a/cadquery/occ_impl/assembly.py
+++ b/cadquery/occ_impl/assembly.py
@@ -350,19 +350,9 @@ def toFusedCAF(
     shapes: List[Shape] = []
     colors = []
 
-    def extract_shapes(assy, parent_loc=None, parent_color=None):
-
-        loc = parent_loc * assy.loc if parent_loc else assy.loc
-        color = assy.color if assy.color else parent_color
-
-        for shape in assy.shapes:
-            shapes.append(shape.moved(loc).copy())
-            colors.append(color)
-
-        for ch in assy.children:
-            extract_shapes(ch, loc, color)
-
-    extract_shapes(assy)
+    for shape, _, loc, color in assy:
+        shapes.append(shape.moved(loc).copy())
+        colors.append(color)
 
     # Initialize with a dummy value for mypy
     top_level_shape = cast(TopoDS_Shape, None)
--- a/cadquery/occ_impl/assembly.py
+++ b/cadquery/occ_impl/assembly.py
@@ -32,6 +32,9 @@ from vtkmodules.vtkRenderingCore import (
     vtkRenderer,
 )
 
+from vtkmodules.vtkFiltersExtraction import vtkExtractCellsByType
+from vtkmodules.vtkCommonDataModel import VTK_TRIANGLE, VTK_LINE, VTK_VERTEX
+
 from .geom import Location
 from .shapes import Shape, Solid, Compound
 from .exporters.vtk import toString
@@ -260,8 +263,29 @@ def toVTK(
 
         data = shape.toVtkPolyData(tolerance, angularTolerance)
 
+        # extract faces
+        extr = vtkExtractCellsByType()
+        extr.SetInputDataObject(data)
+
+        extr.AddCellType(VTK_LINE)
+        extr.AddCellType(VTK_VERTEX)
+        extr.Update()
+        data_edges = extr.GetOutput()
+
+        # extract edges
+        extr = vtkExtractCellsByType()
+        extr.SetInputDataObject(data)
+
+        extr.AddCellType(VTK_TRIANGLE)
+        extr.Update()
+        data_faces = extr.GetOutput()
+
+        # remove normals from edges
+        data_edges.GetPointData().RemoveArray("Normals")
+
+        # add both to the renderer
         mapper = vtkMapper()
-        mapper.SetInputData(data)
+        mapper.AddInputDataObject(data_faces)
 
         actor = vtkActor()
         actor.SetMapper(mapper)
@@ -272,12 +296,23 @@ def toVTK(
 
         renderer.AddActor(actor)
 
+        mapper = vtkMapper()
+        mapper.AddInputDataObject(data_edges)
+
+        actor = vtkActor()
+        actor.SetMapper(mapper)
+        actor.SetPosition(*trans)
+        actor.SetOrientation(*map(degrees, rot))
+        actor.GetProperty().SetColor(0, 0, 0)
+        actor.GetProperty().SetLineWidth(2)
+
+        renderer.AddActor(actor)
+
     return renderer
 
 
 def toJSON(
     assy: AssemblyProtocol,
-    loc: Location = Location(),
     color: Tuple[float, float, float, float] = (1.0, 1.0, 1.0, 1.0),
     tolerance: float = 1e-3,
 ) -> List[Dict[str, Any]]:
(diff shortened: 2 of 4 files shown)


adam-urbanczyk:
@shimwell this PR still needs some polishing, but I think that the functionality is there. Could you verify that this is what you want:
```python
b1 = cq.Workplane().box(1, 1, 1)
b2 = cq.Workplane(origin=(1,0,0)).box(1,1,1)

assy = cq.Assembly().add(b1, color=cq.Color("red")).add(b2)

r, o = cq.occ_impl.assembly.imprint(assy)
```
where `r` is the imprinted shape, and `o` is a dict mapping solids from r to original ids:
```python
s1,s2 = r.Solids()
assert(s1 in o)
assert(s2 in o)
```

shimwell:
That does look ideal from your description above and the code above.

I shall try to install this branch and run some models to see how it behaves with curves, faces that are partially shared, faces share between more that two volumes and anything else I can think of.

Many thanks for all of this targeted development

shimwell:
Ok I think I understand how to correlate the original assembly objects to the imprinted solids.

I adapted the code to print out:
- imprinted solids IDs
- correspondence dictionary between original objects IDs and imprinted solids IDs
- original assembly objects IDs

I notice the assembly has 3 objects so guess the first one is the assembly structure itself. I wonder if that is always present. But i can see matching numbers in the original, imprinted and the correspondence dict which is just what I needed.

Note to self the ID appears to be made from the assembly ID and the solid ID. Perphaps use ```stri.split('./')``` when compairing

```python

import cadquery as cq

b1 = cq.Workplane().box(1, 1, 1)
b2 = cq.Workplane(origin=(1,0,0)).box(1,1,1)

assy = cq.Assembly().add(b1, color=cq.Color("red")).add(b2)

r, o = cq.occ_impl.assembly.imprint(assy)

print('imprinted solids', r.Solids, '\n')

for k, v in o.items():
    print('imprinted solid', k, 'corresponding original solid id', v)

print('\noriginal solds ID', assy.objects.keys())
```

pint out
```python
imprinted solids <bound method Shape.Solids of <cadquery.occ_impl.shapes.Shape object at 0x7f53abcc17f0>> 

imprinted solid <cadquery.occ_impl.shapes.Solid object at 0x7f53abcc15b0> corresponding original solid id ('9910a432-0fb2-11ee-a766-a434d95c2a9c/9910a5e0-0fb2-11ee-a766-a434d95c2a9c',)
imprinted solid <cadquery.occ_impl.shapes.Solid object at 0x7f53abcc1640> corresponding original solid id ('9910a432-0fb2-11ee-a766-a434d95c2a9c/9910a6a8-0fb2-11ee-a766-a434d95c2a9c',)

original solds ID dict_keys(['9910a432-0fb2-11ee-a766-a434d95c2a9c', '9910a5e0-0fb2-11ee-a766-a434d95c2a9c', '9910a6a8-0fb2-11ee-a766-a434d95c2a9c'])
```

So in this case the first original solid is the first imprinted solid
The second original solid is the second imprinted solid.

adam-urbanczyk:
> I notice the assembly has 3 objects so guess the first one is the assembly structure itself. I wonder if that is always present. But i can see matching numbers in the original, imprinted and the correspondence dict which is just what I needed.

Yes the top level assy is empty - it has no Shape object. In general it does not have to be like that.

shimwell:
I've ran a few tests on odd shapes and checked the number of faces is correct and this works in the cases I tried.

The unique ID dictionaries also work nicely.

I've been trying to pass the resulting imprinted assembly into gmsh by trying the ```_address``` attribute but am not having much luck, perhaps I am doing something wrong but it looks like the ```_address``` is not available for TopoDS_Compound.

```python
import cadquery as cq
small = cq.Workplane().box(1, 1, 1)
big = cq.Workplane(origin=(1,0,0)).box(1,2,2)
assy = cq.Assembly()
assy.add(small)
assy.add(big)
r, o = cq.occ_impl.assembly.imprint(assy)
print(len(r.Faces()))  # prints 12 which is correct
topods = r.wrapped  # perhaps this is not the best way to get a topoDS
object_to_pass_to_gmsh=topods._address()
```
error message
```
OCP.TopoDS.TopoDS_Compound' object has no attribute '_address'
```

adam-urbanczyk:
I think you need to install `ocp=7.7.1`. It is currently available via `cadquery/dev`.

adam-urbanczyk:
@shimwell did you manage? FYI: I think we'll switch to 7.7.1 after releasing CQ 2.3, but that should not stop you form using it in a local dev env.

shimwell:
I have been trying but struggling to get the environment to include everything (ocp7.7.1, gmsh, cadquery imprint branch). I've been trying to conda install from the ocp dev tar.gz, pip install from the cadquery branch and conda install gmsh from conda forge but keep getting package conflicts.

I was able to test a bunch of shapes with share surfaces end up with the right number of surfaces so I think this PR does what it aims to do.

As for the gmsh integration I think that get easier to test with releases

adam-urbanczyk:
Hm, env.yaml does not seem to support labels.

environment.yml
```diff
@@ -1,11 +1,13 @@
...
 channels:
   - conda-forge
+  - cadquery/dev
   - cadquery
 dependencies:
   - python>=3.8
   - ipython
-  - ocp=7.7.*
+  - ocp=7.7.1.*
+  - python-gmsh
```
shimwell:
```suggestion
  - python-gmsh
  - gmsh
```

I'm not sure about this but I think python -gmsh includes in binding while gmsh is the executable

shimwell:
Trying that environment.yml you with a ```pip install .``` and a conda install from the url get me the environment.

```
name: cadquery
channels:
  - conda-forge
  - cadquery
dependencies:
  - python=3.9
  - ipython
  - ocp=7.7.*
  - pyparsing>=2.1.9
  - sphinx=5.0.1
  - sphinx_rtd_theme
  - black=19.10b0
  - click=8.0.4
  - mypy
  - codecov
  - pytest
  - pytest-cov
  - ezdxf
  - ipython
  - typing_extensions
  - nptyping=2.0.1
  - nlopt
  - path
  - casadi
  - multimethod >=1.7,<2.0
  - python-gmsh
  - gmsh
  # - https://anaconda.org/CadQuery/ocp/7.7.1.0/download/linux-64/ocp-7.7.1.0-py39_0.tar.bz2
  - pip
  - pip:
    - --editable=.
    - git+https://github.com/CadQuery/OCP-stubs.git
```

```
cd cadquery
pip install -e .
conda install https://anaconda.org/CadQuery/ocp/7.7.1.0/download/linux-64/ocp-7.7.1.0-py39_0.tar.bz2
```
Then running this script
```python
import cadquery as cq
import gmsh
small = cq.Workplane().box(1, 1, 1)
big = cq.Workplane(origin=(1,0,0)).box(1,2,2)
assy = cq.Assembly()
assy.add(small)
assy.add(big)
r, o = cq.occ_impl.assembly.imprint(assy)
print(len(r.Faces()))  # prints 12 which is correct
topods = r.wrapped
object_to_pass_to_gmsh=topods._address()
gmsh.initialize()
gmsh.option.setNumber("General.Terminal", 1)
volumes = gmsh.model.occ.importShapesNativePointer(object_to_pass_to_gmsh)
gmsh.model.occ.synchronize()
gmsh.option.setNumber("Mesh.Algorithm", 1)
gmsh.option.setNumber("Mesh.MeshSizeMin", 20)
gmsh.option.setNumber("Mesh.MeshSizeMax", 30)
gmsh.model.mesh.generate(2)
gmsh.write("gmsh_of_cadquery_in_memory.msh")
```
Get me to this new (for me) error.
```
  File "/home/j/cad_imprint_test/cq_to_gmsh_in_memory.py", line 15, in <module>
    import gmsh
  File "/home/j/miniforge/envs/cadquery/lib/python3.9/site-packages/gmsh.py", line 86, in <module>
    lib = CDLL(libpath)
  File "/home/j/miniforge/envs/cadquery/lib/python3.9/ctypes/__init__.py", line 374, in __init__
    self._handle = _dlopen(self._name, mode)
OSError: /home/j/miniforge/envs/cadquery/lib/libgmsh.so.4.11: undefined symbol: _ZNK20XCAFDoc_MaterialTool11GetMaterialERK9TDF_LabelRN11opencascade6handleI24TCollection_HAsciiStringEES7_RdS7_S7_
```

environment.yml
adam-urbanczyk: Not needed, gmsh is already a dep of python-gmsh

adam-urbanczyk:
Looks like you are using gmsh that was not built against 7.7.1 but does not pin the occt version fully. You could try installing the latest version/build of gmsh, but I'd propose to keep this discussion outside of this PR. You can still test with saving to disk, the order of solids should be stable after export/import. Keys of the output dict are added in that order too.

adam-urbanczyk:
Let me know what you think @lorenzncode @jmwright . `Assembly.__iter__` can be probably used in more places.

lorenzncode:
> from the example at the top I am just wondering why the values of the `o` dictionary are tuples of len 1 instead of strings

My understanding is because there can be more than one origin per solid in the result.
Say there is an intersection such as: 

```py
b1 = cq.Workplane().box(1, 1, 1)
b2 = cq.Workplane(origin=(-0.2,0,0)).box(1,1,1)
```

jmwright:
@adam-urbanczyk I tested the refactor against the KiCAD generator and the files seem unchanged, so that's good. The rest of the code looks good to me, but I see that changes are still being pushed to this PR.

cadquery/assembly.py
```diff
@@ -543,6 +543,28 @@ def _flatten(self, parents=[]):
...
         return rv
 
+    def __iter__(
+        self,
+        loc: Optional[Location] = None,
+        name: Optional[str] = None,
+        color: Optional[Color] = None,
+    ) -> Iterator[Tuple[Shape, str, Location, Optional[Color]]]:
+        """
+        Assembly iterator yielding shapes, locatons, names and colors.
```
lorenzncode, line 553: typo: locations

cadquery/assembly.py
```diff
@@ -543,6 +543,28 @@ def _flatten(self, parents=[]):
...
+        """
+
+        name = f"{name}/{self.name}" if name else self.name
+        loc = loc * self.loc if loc else self.loc
+        color = self.color if self.color else color
+
+        if self.obj:
+            yield self.obj if isinstance(self.obj, Shape) else Compound.makeCompound(
```
lorenzncode, line 561: The docstring order above is inconsistent with yield order here.

Change the order to (shape, location, name, color)  or change the docstring above to be consistent with the yield order?

cadquery/occ_impl/assembly.py
```diff
@@ -411,3 +413,37 @@ def extract_shapes(assy, parent_loc=None, parent_color=None):
...
+    bldr.SetRunParallel(True)
+    bldr.SetUseOBB(True)
+
+    for obj in id_map:
+        bldr.AddArgument(obj.wrapped)
+
+    bldr.Perform()
+    res = Shape(bldr.Shape())
+
+    # make the conneted solid -> id map
```
lorenzncode, line 441: typo: connected

cadquery/occ_impl/assembly.py
```diff
@@ -411,3 +413,37 @@ def extract_shapes(assy, parent_loc=None, parent_color=None):
...
+    origins: Dict[Shape, Tuple[str, ...]] = {}
+
+    for s in res.Solids():
+        ids = tuple(id_map[Solid(el)] for el in bldr.GetOrigins(s.wrapped))
+        # if GetOrigins yiels nothing, solid was not modified
```
lorenzncode, line 446: typo: yields

shimwell:
There is a line in the environment.yml file that says ```ocp=7.7.*```.

Do we need to change this to ```ocp>7.7.1``` as I think this ```_address``` needs OCP version 7.7.1 or above?

lorenzncode (changes requested):

cadquery/occ_impl/assembly.py
adam-urbanczyk: done

cadquery/occ_impl/assembly.py
adam-urbanczyk: fixed

cadquery/assembly.py
adam-urbanczyk: fixed

cadquery/assembly.py
adam-urbanczyk: fixed

adam-urbanczyk:
@jmwright yes, sorry. Some scope creep. I'm fully done with the refactoring. Last change toVTK is to get better rendering. In another PR I'll add a CQ-native `show_object` function for debugging and testing.

afbeelding

@shimwell not in this PR.

lorenzncode (approved):

jmwright (approved):
@adam-urbanczyk This still seems to work fine with the KiCAD generator.

adam-urbanczyk:
Alright, I'll merge tomorrow. @shimwell any last thoughts?

shimwell:
No further thoughts from me, I've tested this with the workflow I'm hoping to use it in and am getting good results. All good from my point of view

shimwell:
Just to mention I've been using  ```_address()``` feature introduced by this PR and I think it occasionally throws up errors. I shall investigate more and write up a minimal example for an issue but at the moment I'm seeing this error over [here](https://github.com/fusion-energy/model_benchmark_zoo/actions/workflows/two_touching_cuboids_cad_to_dagmc.yml)

```python
object_to_pass_to_gmsh=topods._address()
gmsh.initialize()
gmsh.option.setNumber("General.Terminal", 1)
volumes = gmsh.model.occ.importShapesNativePointer(object_to_pass_to_gmsh)
```
```
     @staticmethod
    def importShapesNativePointer(shape, highestDimOnly=True):
        """
        gmsh.model.occ.importShapesNativePointer(shape, highestDimOnly=True)
    
        Imports an OpenCASCADE `shape' by providing a pointer to a native
        OpenCASCADE `TopoDS_Shape' object (passed as a pointer to void). The
        imported entities are returned in `outDimTags' as a vector of (dim, tag)
        pairs. If the optional argument `highestDimOnly' is set, only import the
        highest dimensional entities in `shape'. In Python, this function can be
        used for integration with PythonOCC, in which the SwigPyObject pointer of
        `TopoDS_Shape' must be passed as an int to `shape', i.e., `shape =
        int(pythonocc_shape.this)'. Warning: this function is unsafe, as providing
        an invalid pointer will lead to undefined behavior.
    
        Return `outDimTags'.
    
        Types:
        - `shape': pointer
        - `outDimTags': vector of pairs of integers
        - `highestDimOnly': boolean
        """
        api_outDimTags_, api_outDimTags_n_ = POINTER(c_int)(), c_size_t()
        ierr = c_int()
        lib.gmshModelOccImportShapesNativePointer(
            c_void_p(shape),
            byref(api_outDimTags_), byref(api_outDimTags_n_),
            c_int(bool(highestDimOnly)),
            byref(ierr))
        if ierr.value != 0:
>           raise Exception(logger.getLastError())
E           Exception: OpenCASCADE exception GeomAdaptor_Surface::UContinuity
```

PR #1356, opened 2023-06-27
Add enlarge method to BoundBox

AshtonOhms:
Surfaces Bnd_Box's Enlarge in the BoundBox class.

### DIFF
--- a/cadquery/occ_impl/geom.py
+++ b/cadquery/occ_impl/geom.py
@@ -846,6 +846,22 @@ class BoundBox(object):
 
         return BoundBox(tmp)
 
+    def enlarge(self, tol: float) -> "BoundBox":
+        """Returns a modified (expanded) bounding box, expanded in all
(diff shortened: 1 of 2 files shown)


adam-urbanczyk:
Thanks, could you first explain why is this needed?

AshtonOhms:
This is useful for detecting when geometry is on the edge of (or just outside of) the original, un-enlarged bounding box.

My particular use case is a custom Selector for geometry within a BoundBox using isInside(). I need to enlarge the box by a small margin, otherwise the Selector misses geometry on the edges. Apart from this niche case, I think this is generally helpful for those edge cases with isInside().

adam-urbanczyk (approved):

jmwright (approved):
Looks good to me. I've experimented with automatically finding interreference and excessive gaps between components in an assembly before, and this could be useful for that. Thanks @AshtonOhms

PR #1362, opened 2023-07-01
docs - Update installation instructions and advise to use mambaforge

### LINKED ISSUE #691: Explicitly advise to use miniconda
And to install it in a portable way

```cmd
bitsadmin /transfer myDownloadJob /download /priority normal https://repo.anaconda.com/miniconda/Miniconda3-latest-Windows-x86_64.exe .
Miniconda3-latest-Windows-x86_64.exe /S /InstallationType=JustMe /D=%MINICONDA_DIRNAME%
set "PATH=%MINICONDA_DIRNAME%;%MINICONDA_DIRNAME%\\Scripts;%PATH%"
activate
```

lorenzncode:
Closes #691

I think many of us have been using mamba install for some time with good results.

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,10 +73,10 @@ conda create -n cadquery
 conda activate cadquery
 
 # Install the latest released version
-conda install -c conda-forge cadquery occt=7.7.0
+mamba install -c conda-forge cadquery
 
 # Or install the dev version to get the latest changes
-conda install -c conda-forge -c cadquery cadquery=master
+mamba install -c conda-forge -c cadquery cadquery=master
 ```
 
 ### CadQuery Installation Via Pip
@@ -192,7 +192,7 @@ If you are going to contribute code, make sure to follow this steps:
 - Activate the new conda environment:
   - `conda activate cq-dev`
 - If desired, install the master branch of cq-editor (Note; a release version may not be compatible with the master branch of cadquery):
-  - `conda install -c cadquery -c conda-forge cq-editor=master`
+  - `mamba install -c cadquery -c conda-forge cq-editor=master`
     Installing cq-editor adds another instance of cadquery which overrides the clone just added. Fix this by reinstalling cadquery using pip:
   - `pip install -e .`
 - Before making any changes verify that the current tests pass. Run `pytest` from the root of your cadquery clone, there should be no failures and the output will look similar to this:
--- a/doc/installation.rst
+++ b/doc/installation.rst
@@ -18,54 +18,57 @@ Begin by installing the conda package manager.  If conda is already installed sk
 Install the Conda Package Manager
 ``````````````````````````````````
 
-In principle, any Conda distribution will work, but it is probably best to install `Miniforge <https://github.com/conda-forge/miniforge>`_ to a local directory and to avoid running `conda init`. After performing a local directory installation, Miniforge can be activated via the [scripts,bin]/activate scripts. This will help avoid polluting and breaking the local Python installation.
+In principle, any Conda distribution will work, but it is probably best to install `Mambaforge <https://github.com/conda-forge/miniforge#mambaforge>`_ to a local directory and to avoid running `conda init`. After performing a local directory installation, Mambaforge can be activated via the [scripts,bin]/activate scripts. This will help avoid polluting and breaking the local Python installation.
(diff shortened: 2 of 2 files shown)


lorenzncode:
Closes #1369

README.md
```diff
@@ -73,10 +73,10 @@ conda create -n cadquery
 conda activate cadquery
 
 # Install the latest released version
-conda install -c conda-forge cadquery occt=7.7.0
+mamba install -c conda-forge cadquery
```
whophil, line 76: `mamba` is not available in the base environment of all conda distributions, so this may be confusing for those who do not use conda regularly.

lorenzncode: For those new to conda,  I think it's likely that they are starting without any conda distribution at all.  In that case, the CadQuery installation documentation covers the steps to install Mambaforge which includes mamba.

This PR changes the recommended install method to use mamba.  This change is meant to prevent common cadquery installation issues:

1.  Install failures due to out of memory issues on low mem machines
2. Slow installation (in some cases extremely slow >10x elapsed time than mamba)

Would it help to mention `conda install` as an alternative or include another link to the mamba related docs?

doc/installation.rst
```diff
@@ -18,54 +18,57 @@ Begin by installing the conda package manager.  If conda is already installed sk
...
 
 conda
 `````
 
-Install the latest released version of cadquery [#f1]_::
+Install the latest released version of cadquery::
 
-    conda create -n cqrel
-    conda activate cqrel
-    conda install -c conda-forge cadquery occt=7.7.0
```
shimwell, line 56: I like this PR and the improvements it makes to thin install docs. Just one question do you think removing the ```-c conda-forge``` part will trip up some conda users who are not using mambaforge or miniforge. I'm assuming some people are going to copy paste these instructions. Would it be worth keeping this in which would still work for all?

adam-urbanczyk: AFAICT  channel spec is still there

README.md
adam-urbanczyk: I think it is clear, maybe simply mention mambaforge in the Readme too?

README.md
```diff
@@ -73,10 +73,10 @@ conda create -n cadquery
 conda activate cadquery
```
adam-urbanczyk, line 73: Still few conda instances missed. Also l64 does mention conda

adam-urbanczyk (approved):
Minor issues in the README

README.md
lorenzncode: Do you mean the instances such as `conda activate`?  I replaced only `conda install` with `mamba install` because when I began using mamba, the mamba docs recommended to continue to use `conda activate` and `conda deactivate`.  Not all conda commands had been implemented in mamba. 

The mamba docs have been updated to remove the warning (it had said you should still use conda for activation/deactivation): 
https://github.com/mamba-org/mamba/commit/63e75d8bc27b2ca35c4c602055ee4dfa57ef2059#diff-863a289903793c21e000164624119574147af60f75ecbc04ca3d511779c10798L54

Though the warning is removed my understanding is that `mamba activate` is not fully implemented (does not work in PowerShell). 
https://github.com/mamba-org/mamba/issues/1994#issuecomment-1271462295

mamba-org/mamba#1717

I'm still using a mix of conda and mamba commands.  Perhaps I could move to use only mamba (unless when using PowerShell).

doc/installation.rst
lorenzncode: The installation guide discusses the default channel where the README does not so I thought I'd leave the explicit channel '-c conda-forge' in the README command.

README.md
lorenzncode: We can consider replacing all conda instances with mamba.

At least this one (line 191) should be replaced with mamba when creating the env from yaml:
 `conda env create -n cq-dev -f environment.yml`

adam-urbanczyk: OK, I did not know that PS is not supported (not using it myself). I agree with the second comment.

README.md
lorenzncode: I've added a link to mambaforge in the README and the note:

``mamba install`` is recommended over ``conda install`` for faster and less memory intensive cadquery installation.

PR #1365, opened 2023-07-05
Pin ocp in environment.yml

adam-urbanczyk:


jmwright (approved):
Once all the checks go green on this I'd suggest we merge it so we can fix AppVeyor ASAP.

PR #1377, opened 2023-07-19
Making all download formats of the RTD documentation available

### LINKED ISSUE #1352: Offline version of documentation (zipped HTML)
Without or with a poor internet connection it would be helpful to have an offline version of the documentation.
Please consider generating a downloadable zipped HTML.

https://docs.readthedocs.io/en/stable/downloadable-documentation.html
https://docs.readthedocs.io/en/stable/guides/enable-offline-formats.html

jmwright:
Closes #1352 

It should make zipped HTML, PDF and epub formats available for download.

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -7,3 +7,6 @@ build:
 
 conda:
   environment: environment.yml
+
+formats:
+  - all
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -8,5 +8,4 @@ build:
 conda:
   environment: environment.yml
 
-formats:
-  - all
+formats: all


adam-urbanczyk (approved):
LGTM, how will a pdf +vtkjs look like though?

jmwright:
> how will a pdf +vtkjs look like though?

I honestly don't know. I thought that RTD might build the artifacts for a PR so that we could check them, but it didn't.

sethfischer:
From https://docs.readthedocs.io/en/stable/guides/pull-requests.html#limitations

> Additional formats like PDF and EPUB aren’t built, to reduce build time.

sethfischer:
I believe the sphinx directives in `cadquery/cq_directive.py` will require modification to produce an output suitable for building a PDF. One approach might be for the directive to detect that a PDF is being built, and rather than generate VTK JS + HTML, generate a raster image such as PNG.

lorenzncode (approved):
Perhaps go ahead with this PR and try 'all' to see what it looks like now.  Then limit to htmlzip until when/if PDF is better supported.

jmwright:
Ok, I'll try it. I don't know that any of us have our GitHub accounts linked to RTD (a requirement for enabling PR builds). At least the accounts for Dave and I predate that login option.

Issue #1378, opened 2023-07-21
Sketch and segment names in test_constraint_solver() are confusing


sanderobdeijn:
In these tests the variable name s1 is used for a sketch and a segment. This makes it harder to read and understand the tests. 
For example 
```python
s1 = (
        Sketch()
        .segment((0.0, 0), (0.0, 2.0), "s1")
        .segment((0.5, 2.5), (1.0, 1), "s2")
        .close("s3")
    )
    s1.constrain("s1", "Fixed", None)
    s1.constrain("s1", "s2", "Coincident", None)
    s1.constrain("s2", "s3", "Coincident", None)
    s1.constrain("s3", "s1", "Coincident", None)
    s1.constrain("s3", "s1", "Angle", 90)
    s1.constrain("s2", "s3", "Angle", 180 - 45)
```

Here s1 is used to store the sketch, but also as the name of the first segment.
The first time I was looking at this it was confusing to me.

sanderobdeijn:
The pr is merged so I will close the issue.

PR #1379, opened 2023-07-21
Make segments names more clear in test_constraint_solver()

sanderobdeijn:
Fix for #1378


sanderobdeijn:
I didn't change the sketch variable names because the Sx seems to be the standard in all the tests. But I could also change those to sketch1, 2, etc

jmwright (approved):
Thanks @sanderobdeijn !

PR #1384, opened 2023-08-05
DXF import/export: handle ellipse arc with reversed normal

### LINKED ISSUE #1381: DXF export issue with ellipses
Attached is a brep of a wire that looks in isometric view like this:

image

Attempting to export this with either the default None approx or spline approx results in broken dxf:

image

MRE:

```python
import cadquery as cq

buggy = cq.Shape.importBrep("buggy.brep")
wp = cq.Workplane()
wp.objects = [buggy]
cq.exporters.export(wp, "buggy_none.dxf", opt={})
cq.exporters.export(wp, "buggy_arcs.dxf", opt={"approx": "arc"})
cq.exporters.export(wp, "buggy_spline.dxf", opt={"approx": "spline"})
```

Using "arc" approx the result is as expected.

dxf_circle has some logic to checking the orientation of the circle/arc  and adjusting the params based on that which is missing from dxf_ellipse. Possible cause? I tried poking around a bit, but I'm not quite sure about how to fix this without accidentally breaking something else.

https://github.com/CadQuery/cadquery/blob/5bce8d033104b89ae78095cb385c015e43bae29a/cadquery/occ_impl/exporters/dxf.py#L209-L214

[buggy.tar.gz](https://github.com/CadQuery/cadquery/files/12140206/buggy.tar.gz)

lorenzncode:
Resolves  #1381

### DIFF
--- a/cadquery/occ_impl/exporters/dxf.py
+++ b/cadquery/occ_impl/exporters/dxf.py
@@ -253,13 +253,18 @@ class DxfDocument:
 
         zdir = ellipse.Axis().Direction()
 
-        if zdir.Z() < 0:
-            geom = GC_MakeArcOfEllipse(
+        if zdir.Z() > 0:
+            start_param = geom.FirstParameter()
+            end_param = geom.LastParameter()
+        else:
+            gc = GC_MakeArcOfEllipse(
                 ellipse,
                 geom.FirstParameter(),
                 geom.LastParameter(),
                 False,  # reverse Sense
             ).Value()
+            start_param = gc.FirstParameter()
+            end_param = gc.LastParameter()
 
         return (
             "ELLIPSE",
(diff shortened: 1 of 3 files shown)


lorenzncode:
The DXF output might still be problematic for some other software.  Another solution could be to convert the edge to the plane normal.

adam-urbanczyk:
Thanks @lorenzncode , do you have some example of this not working?

lorenzncode:
@adam-urbanczyk  Sure, here are a few examples importing ellipse_arc.dxf created in the pytest of the PR. 

The CAD Assistant result is working as expected:
image

Inkscape version 1.2.2 is not working:
image

CamBam 1.0.7233.21743 is not working:
image

adam-urbanczyk:
Thanks!  It'd be good to be able to import correctly in Inkscape. Any clue how to attack this?

lorenzncode:
Yes, I found another way is to reverse the Sense of the ellipse with reversed normal.  I checked that it now imports correctly in Inkscape.

adam-urbanczyk (approved):

jmwright:
Should I review this? Everything else is green, it just needs one more review.

lorenzncode:
@jmwright Yes, please review.

jmwright (approved):
Looks good, thanks @lorenzncode

PR #1388, opened 2023-08-15
Fix for incorrect tapered extrude height found in #1383

### LINKED ISSUE #1383: Extrude with taper on any Workplane return a wrong extrude length
Generic use of the extrude(until = x, taper = y) command will result in extrude length less than x.

```
a = list(range(0,60))
z = []
for angle in a:
    part = cq.Solid(cq.Workplane('XY').rect(1,1).extrude(1, taper = angle).toOCC())
    bb = part.BoundingBox()
    z.append(bb.zlen)
```

Running the code above create a bunch of extrude with height 1, with taper ranging from 0 to 60 degree.
Plotting the taper angle (0 to 60 degree) with resulted extrude height (suppose to be 1):
image

Only taper = 0 return the correct height = 1, the rest is all less than 1.
Any thought/alternative solution to get the correct height?
Thanks!

jmwright:
Closes #1383 

It seems that [LocOpe_DPrism](https://dev.opencascade.org/doc/refman/html/class_loc_ope___d_prism.html) is introducing the issue fround in #1383 where the extrude height requested by the user is not respected. I compensated for this in the code and then added a test so we can detect if the behavior ever changes in OCCT.

### DIFF
--- a/cadquery/occ_impl/shapes.py
+++ b/cadquery/occ_impl/shapes.py
@@ -3249,7 +3249,9 @@ class Solid(Shape, Mixin3D):
 
             # Divided by cos of taper angle to ensure the height chosen by the user is respected
             prism_builder = LocOpe_DPrism(
-                face.wrapped, (d * vecNormal.Length) / cos(radians(taper)), d * radians(taper)
+                face.wrapped,
+                (d * vecNormal.Length) / cos(radians(taper)),
+                d * radians(taper),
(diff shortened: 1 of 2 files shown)


adam-urbanczyk (approved):
LGTM, thanks!

lorenzncode (approved):
Thanks @jmwright!

PR #1418, opened 2023-10-13
Added argument handling for ASCII export of assemblies to GLTF and STL

jmwright:
The `binary` option for `occ_impl.exporters.assembly.exportGLTF()` was not getting passed through from the `assembly.save()` call. This PR fixes that and also adds the `ascii` option for STL export as well.

### DIFF
--- a/cadquery/assembly.py
+++ b/cadquery/assembly.py
@@ -468,8 +468,6 @@ class Assembly(object):
         :param angularTolerance: the angular tolerance, in radians. Only used for GLTF, VRML. Default 0.1.
         :param \**kwargs: Additional keyword arguments.  Only used for STEP, GLTF and STL.
             See :meth:`~cadquery.occ_impl.exporters.assembly.exportAssembly`.
-        :param binary: GLTF only - Sets whether or not GTLF export should be text (.gltf) or binary (.glb)
-        :type binary: bool
         :param ascii: STL only - Sets whether or not STL export should be text or binary
         :type ascii: bool
         """
@@ -494,8 +492,9 @@ class Assembly(object):
         elif exportType == "GLTF":
             # Handle the binary option for GLTF export
             binary = True
-            if "binary" in kwargs:
-                binary = bool(kwargs.get("binary"))
+            path_parts = path.split(".")
+            if len(path_parts) > 0 and path_parts[-1] == "gltf":
+                binary = False
 
             exportGLTF(self, path, binary, tolerance, angularTolerance)
         elif exportType == "VTKJS":
--- a/cadquery/assembly.py
+++ b/cadquery/assembly.py
@@ -464,9 +464,9 @@ class Assembly(object):
 
         :param path: Path and filename for writing.
         :param exportType: export format (default: None, results in format being inferred form the path)
-        :param tolerance: the deflection tolerance, in model units. Only used for GLTF, VRML. Default 0.1.
-        :param angularTolerance: the angular tolerance, in radians. Only used for GLTF, VRML. Default 0.1.
-        :param \**kwargs: Additional keyword arguments.  Only used for STEP, GLTF and STL.
+        :param tolerance: the deflection tolerance, in model units. Only used for glTF, VRML. Default 0.1.
+        :param angularTolerance: the angular tolerance, in radians. Only used for glTF, VRML. Default 0.1.
+        :param \**kwargs: Additional keyword arguments.  Only used for STEP, glTF and STL.
             See :meth:`~cadquery.occ_impl.exporters.assembly.exportAssembly`.
         :param ascii: STL only - Sets whether or not STL export should be text or binary
         :type ascii: bool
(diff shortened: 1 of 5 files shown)


jmwright:
@lorenzncode @adam-urbanczyk Please take a look when you have time.

lorenzncode (changes requested):
Thanks @jmwright.   Yes, IIRC there was a timing issue where the STL assembly export PR was created before but merged after STL binary support was added.

I would suggest not to add both `binary` and `ascii` keywords.
```py
# ?
assy.save("out.stl", binary=True, ascii=True)
```

Can glTF export instead be enhanced to handle the format automatically based on extension?  

I expect this to export to JSON.:
```
assy.save("out.gltf")
```

Expect binary glTF instead of error: 
```
assy.save("out.glb")
```

jmwright:
@lorenzncode I made that change, but do you think that level of implicit behavior will be confusing for users? I added a note to the import/export documentation, which apparently had not been updated yet to show that we support GLTF export.

lorenzncode:
> I made that change, but do you think that level of implicit behavior will be confusing for users?

@jmwright No, not at all!  I've opened issue #1419.  Other tools handle the glTF format automatically by file extension as well.  The format is encoded in the file extension explicitly already by convention.

cadquery/assembly.py
```diff
@@ -488,11 +490,22 @@ def save(
         elif exportType == "VRML":
             exportVRML(self, path, tolerance, angularTolerance)
         elif exportType == "GLTF":
-            exportGLTF(self, path, True, tolerance, angularTolerance)
+            # Handle the binary option for GLTF export
```
lorenzncode, line 493: I've opened #1420.  IMO, `save` must also recognize the .glb extension automatically.

cadquery/assembly.py
```diff
@@ -488,11 +490,22 @@ def save(
...
+            binary = True
+            path_parts = path.split(".")
+            if len(path_parts) > 0 and path_parts[-1] == "gltf":
+                binary = False
```
lorenzncode, line 497: Pending discussion/conclusion of #1419, the binary handling could be moved to `exportGLTF`.

lorenzncode (changes requested):

cadquery/assembly.py
jmwright: @lorenzncode I've restructured this PR to take care of #1419 and #1420

cadquery/occ_impl/exporters/assembly.py
```diff
@@ -185,14 +186,24 @@ def exportVRML(
 def exportGLTF(
     assy: AssemblyProtocol,
     path: str,
-    binary: bool = True,
+    binary: Optional[bool] = None,
```
adam-urbanczyk, line 189: Why not `binary: bool = False`?

adam-urbanczyk: Ah, never mind. You want to infer from the extension

adam-urbanczyk (commented):
LGTM, you might want to fix coverage though.

cadquery/assembly.py
lorenzncode: With the PR branch this is working:
`cq.exporters.assembly.exportGLTF(assy, "assy.glb")`

This fails:
`assy.save("assy.glb")`

doc/importexport.rst
```diff
@@ -32,6 +32,7 @@ Export Formats
 * VRML
 * VTP
 * 3MF
+* GLTF
```
lorenzncode, line 35: Where the documentation refers to the file format I'd suggest to change GLTF -> glTF.

doc/importexport.rst
```diff
@@ -42,6 +43,7 @@ Notes on the Formats
 * TJS is short for ThreeJS, and is a JSON mesh format that is useful for displaying 3D models in web browsers. The TJS format is used to display embedded 3D examples within the CadQuery documentation.
 * VRML is a mesh-based format for representing interactive 3D objects in a web browser.
 * VTP is a mesh-based format used by the VTK library.
+* GLTF is a mesh-based format useful for viewing models on the web. Whether the resulting GLTF file is binary (.glb) or text (.gltf) is set by the file extension.
```
lorenzncode, line 46: Change GLTF to glTF.

cadquery/assembly.py
```diff
@@ -466,8 +466,10 @@ def save(
         :param exportType: export format (default: None, results in format being inferred form the path)
         :param tolerance: the deflection tolerance, in model units. Only used for GLTF, VRML. Default 0.1.
         :param angularTolerance: the angular tolerance, in radians. Only used for GLTF, VRML. Default 0.1.
-        :param \**kwargs: Additional keyword arguments.  Only used for STEP.
+        :param \**kwargs: Additional keyword arguments.  Only used for STEP, GLTF and STL.
```
lorenzncode, line 469: I think it should be changed here in the docstring too for param kwargs: GLTF -> glTF
Also for tolerance, angularTolerance.

jmwright:
codecov/patch keeps failing, but it does not seem like CodeCov is showing me the correct information when I click through to its report. What am I missing?

jmwright:
I finally was able to fix CodeCov.

@lorenzncode I think I have all of your suggestions implemented.

lorenzncode (approved):
Thanks @jmwright !  I tested the branch and did not find any other issues.

lorenzncode:
@adam-urbanczyk, @jmwright Is there a policy against git sqaush?  If not, perhaps squash might be applied more on merge.
 
For example, I like black, but maybe squash commits such as black fixes, codecov tests, etc.:
```
$ git log | grep -i black | wc
     89     460    3151
```

jmwright:
@lorenzncode squash is good, I just always forget to do it.

adam-urbanczyk:
I always squash when merging.

PR #1423, opened 2023-10-21
Fix example documentation typos

sermelo:
Two typos are fixes in this pull request:

- In the example "Using Point Lists" the `pushPoints` method is used. That method should be referenced in the "API reference" section instead of the `points`.
- In the "Copying Workplanes" example there is a missing ":" in the "API reference" hlist that make the section look empty


adam-urbanczyk:
Thanks @sermelo !

PR #1432, opened 2023-11-01
Modify STL export to enable non-relative tolerancing and speed up export.

brhubbar:
First off, thank you for building this tool - it's been awesome to get to know! Greatly appreciate the API and the comments contained throughout, and it has saved me a lot of headache!

In the event that a model has very large and very small geometries, relative sizing for the mesh results in one of those scales being severely over- or under-resolved. For example, a dodecagon might be sufficient for a small hole, but a large hole requires many more facets to be properly represented. With relative sizing, getting the large hole up to a sufficient resolution results in the small hole meshing down to sizes that approach floating point errors and result in geometry collapses. 

Also, the docs explicitly state that mesh.Perform() is called by the constructor, so calling it again doubles runtime with no change in the result (that I've been able to observe)

None of the changes here modify the existing default behavior.

This PR is ready for review ✔️

### DIFF
--- a/cadquery/occ_impl/shapes.py
+++ b/cadquery/occ_impl/shapes.py
@@ -434,7 +434,9 @@ class Shape(object):
             Setting this value to True may be problematic for parts with a wide range of feature sizes, as large features will be very faceted, or small features will be very dense.
         """
         # The constructor used here automatically calls mesh.Perform(). https://dev.opencascade.org/doc/refman/html/class_b_rep_mesh___incremental_mesh.html#a3a383b3afe164161a3aa59a492180ac6
-        BRepMesh_IncrementalMesh(self.wrapped, tolerance, isToleranceRelative, angularTolerance, isParallel)
+        BRepMesh_IncrementalMesh(
+            self.wrapped, tolerance, isToleranceRelative, angularTolerance, isParallel
+        )
 
         writer = StlAPI_Writer()
         writer.ASCIIMode = bool(ascii)
--- a/cadquery/occ_impl/shapes.py
+++ b/cadquery/occ_impl/shapes.py
@@ -418,7 +418,7 @@ class Shape(object):
         angularTolerance: float = 0.1,
         ascii: bool = False,
         isToleranceRelative: bool = True,
-        isParallel: bool = False,
+        parallel: bool = True,
     ) -> bool:
         """
         Exports a shape to a specified STL file.
@@ -430,12 +430,13 @@ class Shape(object):
             Default is 1e-3, which is a good starting point for a range of cases.
         :param angularTolerance: Angular deflection setting which limits the angle between subsequent segments in a polyline. Default is 0.1.
         :param ascii: Export the file as ASCII (True) or binary (False) STL format.  Default is binary.
-        :param isToleranceRelative: If True, deflection used for discretization of each edge will be <tolerance> * <size of="" edge>="">. Deflection used for the faces will be the maximum deflection of their edges.
-            Setting this value to True may be problematic for parts with a wide range of feature sizes, as large features will be very faceted, or small features will be very dense.
+        :param isToleranceRelative: If True, tolerance will be scaled by the size of the edge being meshed. Default is True.
+            Setting this value to True may cause large features to become faceted, or small features dense.
+        :param parallel: If True, OCCT will use parallel processing to mesh the shape. Default is True.
         """
         # The constructor used here automatically calls mesh.Perform(). https://dev.opencascade.org/doc/refman/html/class_b_rep_mesh___incremental_mesh.html#a3a383b3afe164161a3aa59a492180ac6
         BRepMesh_IncrementalMesh(
-            self.wrapped, tolerance, isToleranceRelative, angularTolerance, isParallel
+            self.wrapped, tolerance, isToleranceRelative, angularTolerance, parallel
         )
 
         writer = StlAPI_Writer()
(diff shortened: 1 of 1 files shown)


cadquery/occ_impl/shapes.py
```diff
@@ -428,17 +430,16 @@ def exportStl(
...
+            self.wrapped, tolerance, isToleranceRelative, angularTolerance, isParallel
+        )
 
         writer = StlAPI_Writer()
-
-        if ascii:
-            writer.ASCIIMode = True
-        else:
-            writer.ASCIIMode = False
+        writer.ASCIIMode = bool(ascii)
```
adam-urbanczyk, line 442:
```suggestion
        writer.ASCIIMode = ascii
```

adam-urbanczyk (changes requested):
LGTM, one minor tweak

cadquery/occ_impl/shapes.py
brhubbar: Huh - didn't realize I could just commit it from the browser here. Change made, thanks for the feedback. Just wanted to be certain I didn't change the behavior at all.

adam-urbanczyk (approved):

cadquery/occ_impl/shapes.py
```diff
@@ -417,6 +417,8 @@ def exportStl(
         tolerance: float = 1e-3,
         angularTolerance: float = 0.1,
         ascii: bool = False,
+        isToleranceRelative: bool = True,
+        isParallel: bool = False,
```
lorenzncode, line 421: For consistency change parameter name to `parallel`?  See _bool_op.

cadquery/occ_impl/shapes.py
```diff
@@ -428,17 +430,16 @@ def exportStl(
             Default is 1e-3, which is a good starting point for a range of cases.
         :param angularTolerance: Angular deflection setting which limits the angle between subsequent segments in a polyline. Default is 0.1.
         :param ascii: Export the file as ASCII (True) or binary (False) STL format.  Default is binary.
+        :param isToleranceRelative: If True, deflection used for discretization of each edge will be <tolerance> * <size of="" edge>="">. Deflection used for the faces will be the maximum deflection of their edges.
```
lorenzncode, line 433: Should this part of the docstring be `<tolerance> * <size of edge>` ? 

While there is no hard guideline on comment or docstring line length,  I would also suggest to add a few line breaks.  This is currently the longest docstring line in the file.

lorenzncode (commented):
> mesh.Perform() is called by the constructor, so calling it again doubles runtime 

Unfortunately I cannot reproduce the doubled runtime or performance benefit of removing the call.  I agree the the Perform call can be removed. 
I believe OCCT skips building the mesh again if it already exists.


I did find significant performance improvement with the parallel option enabled.

In one example, time to generate a large mesh with 4.5e6 facets, ~220MB STL file, was 18 seconds vs 122 seconds.

I'd suggest to enable the parallel flag by default.  It is enabled for boolean operations:
https://github.com/CadQuery/cadquery/blob/7bc8c00dc1a81633daea4810e04762d61a2abd5a/cadquery/occ_impl/shapes.py#L1053

cadquery/occ_impl/shapes.py
brhubbar: Yeah I can shorten that down. I lazily copied it from the OCCT docs

cadquery/occ_impl/shapes.py
brhubbar: Will do (tomorrow)

brhubbar:
> Unfortunately I cannot reproduce the doubled runtime or performance benefit of removing the call. I agree the the Perform call can be removed.
I believe OCCT skips building the mesh again if it already exists.
>
> I did find significant performance improvement with the parallel option enabled.
> 
> In one example, time to generate a large mesh with 4.5e6 facets, ~220MB STL file, was 18 seconds vs 122 seconds.

Interesting the difference here - I’m running Windows + Pip package, I wonder if that’s a source of the difference

> I'd suggest to enable the parallel flag by default. It is enabled for boolean operations:

Can do - I only did not in order to not change any current behavior.

lorenzncode (approved):

cadquery/occ_impl/shapes.py
```diff
@@ -417,6 +417,8 @@ def exportStl(
         tolerance: float = 1e-3,
         angularTolerance: float = 0.1,
         ascii: bool = False,
+        isToleranceRelative: bool = True,
```
adam-urbanczyk, line 420: Could you simply use `relative` here?

adam-urbanczyk (commented):
Just some bike-shedding

cadquery/occ_impl/shapes.py
brhubbar: I made it more specific in an attempt to be clear that it applies to `tolerance` only, not `angularTolerance`. I tend toward including ‘is’ in Boolean names, but I realize that’s not consistent with the rest of cadquery.

I suppose someone could also interpret `relative` to have some effect on the spatial placement of the exported shape relative to the origin.

Both of the above can be resolved by reading the docstring though. I can switch it if you like.

lorenzncode: +1 for `relative`

adam-urbanczyk: > Both of the above can be resolved by reading the docstring though. I can switch it if you like.

yes, please

brhubbar: Done ✅

adam-urbanczyk: Thank you!

adam-urbanczyk:
Thanks @brhubbar !

PR #1451, opened 2023-11-30
Fix parenthesis bug in `_isCoPlanar` function

egrim:
I came across what appears to be a parenthesis misplacement for `abs` in the `_isCoPlanar` function.


jmwright (approved):
Thanks @egrim 

@lorenzncode @adam-urbanczyk Could one of you review this when you get a chance?

adam-urbanczyk (approved):

Issue #1457, opened 2023-12-05
Assembly save mode=default vs mode=fused


ZTThomasOberthanner:
My goal is to export objects into a stl-file. Afterwards these objects are processed using blender split by loose parts.

When reading the docs [cadquery.Assembly.save](https://cadquery.readthedocs.io/en/latest/classreference.html#cadquery.Assembly.save) it is possible to fuse all parts into a single one using <code>mode=fused</code>. But even using <code>mode=default</code> all objects are fused into a single part. Is this a bug or my misunderstanding?

These are 3 different assemblies, all made of two boxes:
<img width="849" alt="image" src="https://github.com/CadQuery/cadquery/assets/92015468/cdb0200d-44dd-4985-b2e8-9ca21ce86a38">
```
import cadquery as cq

box = cq.Workplane().box(1, 1, 1)

assy_default = cq.Assembly(name="toplevel")
assy_default.add(box, name="box1")
assy_default.add(box, name="box2", loc=cq.Location((1, 0, 0)))
assy_default.save("boxes_default.stl", exportType='STL', mode='default')

assy_fused = cq.Assembly(name="toplevel")
assy_fused.add(box, name="box1", loc=cq.Location((0, 5, 0)))
assy_fused.add(box, name="box2", loc=cq.Location((1, 5, 0)))
assy_fused.save("boxes_fused.stl", exportType='STL', mode='fused')

assy_split = cq.Assembly(name="toplevel")
assy_split.add(box, name="box1", loc=cq.Location((0, 10, 0)))
assy_split.add(box, name="box2", loc=cq.Location((2, 10, 0)))
assy_split.save("boxes_split.stl", exportType='STL')
```

PS: I wanted to paste the output using CQ-editor, but got a ValueError: Unknown Format: STL

jmwright:
For STL export the assembly is made into a single compound by calling `toCompound()` on it.

https://github.com/CadQuery/cadquery/blob/6869dc625bf29e098204fcd71bbe1025fd553644/changes.md?plain=1#L74

The notes at the top of the document [here](https://help.prusa3d.com/article/split-to-objects-parts_1751) explain why we do this with STL files.

The fused mode was added mainly for STEP, which supports multiple parts/colors in the file.

ZTThomasOberthanner:
thank you for the quick reply - I will create a pull request to add this information to the comments/documentation

ZTThomasOberthanner:
https://github.com/CadQuery/cadquery/pull/1458

PR #1458, opened 2023-12-05
added documentation for Assembly.save() param mode

ZTThomasOberthanner:
mode is only used for STEP files as discussed here 
[https://github.com/CadQuery/cadquery/issues/1457](https://github.com/CadQuery/cadquery/issues/1457)
so I suggest to a line to the comments describing that behaviour


adam-urbanczyk:
Thanks @ZTThomasOberthanner !

PR #1467, opened 2023-12-18
Add Brep to supported importShape functions

brhubbar:
- Makes Brep support available to the importer 'factory' implemented separate to the `Shape` object
- Accounts for the case where a Brep returns as a non-iterable, such as a Solid
- ~~**This is not unit tested** but has worked for my purposes~~ This is unit tested, thanks for the assist from Adam
- Ready for review ✔️

### DIFF
--- a/cadquery/occ_impl/importers/__init__.py
+++ b/cadquery/occ_impl/importers/__init__.py
@@ -14,6 +14,7 @@ RAD2DEG = 360.0 / (2 * pi)
 class ImportTypes:
     STEP = "STEP"
     DXF = "DXF"
+    BREP = "BREP"
 
 
 class UNITS:
@@ -22,7 +23,7 @@ class UNITS:
 
 
 def importShape(
-    importType: Literal["STEP", "DXF"], fileName: str, *args, **kwargs
+    importType: Literal["STEP", "DXF", "BREP"], fileName: str, *args, **kwargs
 ) -> "cq.Workplane":
     """
     Imports a file based on the type (STEP, STL, etc)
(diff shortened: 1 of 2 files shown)


adam-urbanczyk:
Could you add a test?

brhubbar:
Hmm. I seem to be a bit stuck. I baked the test into the existing importStep, since the results should be the same for both cases. However, `importBrep` seems to be returning a Shell on the azure pipeline:

```
AssertionError: 'Shell' != 'Solid'
```

I cannot replicate this locally (I get a shape type of `Solid`). @adam-urbanczyk  - can you 

- see if you can reproduce the failure coming from Azure and maybe help me troubleshoot
- or, grant permission to not assert that the import be a Solid? This option seems wrong to me - I don't see why a box *wouldn't* be a solid on import, since it *is* a closed surface by definition. However, I don't want to overconstrain myself if I'm misunderstanding

adam-urbanczyk:
@brhubbar I think I fixed the issue. Could you make the diff coverage 100%?

brhubbar:
Okay - sorry for all of the lazy test writing; I've been trying to send up fixes in between other things and it has not been working. I fixed my environment, am getting results consistent with the CI pipelines, and have some time to think about this. 

My latest test revealed an inconsistency between the STEP and BREP importers. This is because the STEP importer might import multiple "Shapes" and *always* tosses them into a list for `Workplane.newObject` - even if the STEP import returns a single Compound. The Brep importer was splitting compounds out into multiple Solids. While this made sense to me - it isn't consistent with how the brep is exported, so I changed it to just toss the whole compound as a single object onto the workplane. A comment in the importer explains how to get the previous behavior.

brhubbar:
Hey @adam-urbanczyk  - just checking if there's anything more you would like to see from me on this - as far as I can tell, tests are passing now (assuming they test the actual behavior you'd like to see)

adam-urbanczyk (approved):

jmwright (approved):
Thanks @brhubbar 

@adam-urbanczyk Do you know what is causing Python 3.12 on Windows to hang on Azure? It hangs on the first Bash step.

adam-urbanczyk:
I think it works now.

jmwright:
Looks good, merging.

PR #1440, opened 2023-11-11
OCP 7.7.2

adam-urbanczyk:
Move to OCP 7.7.2

PR #1468, opened 2023-12-20
Require at least 3.8

adam-urbanczyk:
Related to #1440


jmwright:
@adam-urbanczyk Should the line be something like `python>=3.8,<=3.12` to prevent breakage when Python 3.13/3.14 is released?

adam-urbanczyk:
Trivial change, going ahead with the merge.

PR #1492, opened 2024-01-13
Fix documentation hole, to many parameter

jlombacher:
Just minor, but was confusing to me


jlombacher:
> Thanks @jlombacher
> 
> Why do you set `clean` to `True` since that is the default in the method signature?

Good point just removed it.

jmwright (approved):
Thanks @jlombacher 

Simple change and I have confirmed that the example was broken before. Merging.

PR #1501, opened 2024-01-17
Update primer.rst - missing show_object(assy)

bmsleight:
Example does not image without show_object(assy)

### DIFF
--- a/doc/primer.rst
+++ b/doc/primer.rst
@@ -863,6 +863,8 @@ constraints to obtain a fully parametric assembly. This can be achieved in the f
         .solve()
     )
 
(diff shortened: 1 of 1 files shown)


jmwright:
@bmsleight None of the other examples in this section include `show_object` either. I think the code is just meant to be representative, and could be run in any execution environment (CQ-editor, jupyter-cadquery, etc), requiring different show calls.

adam-urbanczyk:
Agree with @jmwright - opened https://github.com/CadQuery/CQ-editor/pull/437 to address this for CQ-editor.

PR #1514, opened 2024-01-30
filter/map/apply/sort/[]/invoke

voneiden:
Some time back I made [cq-filter](https://github.com/voneiden/cq-filter) to allow easier manipulation of workplane objects without breaking out of the fluent API. @adam-urbanczyk  requested in https://github.com/voneiden/cq-filter/issues/1 to merge some features from cq-filter into cadquery, namely:

* __getitem__
* filter
* sort
 
and optionally

* group

----

First off, I've decided to rename `filter`  to `filter_objects` and `sort` to `sort_objects`. I think the verbosity is warranted as `Workplane` is a fairly complex object. 

The cq-filter implementation of  `group` created an intermediary workplane with extra fields to support slicing groups. I never particularly liked doing it this way, so I spent some time thinking how this could be done without the intermediary workplane. One option, used in this PR, is to create a separate Group object that supports slicing. This does require however the ability for the `Group` to iterate through candidate objects independently, so it can not be used with `filter_objects` as it allows the supplied filter function to operate only one one `CQObject` at a time. 

Therefore I needed to introduce another method, currently named as  `map_objects` which hands over the workplane objects to the map function and expects a new list of objects to be returned.

Group supports fixed window groups (x<sub>n</sub> - x<sub>0</sub> <= tol) and moving window groups (x<sub>n</sub> - x<sub>n-1</sub> <= tol)

------

Currently the PR is missing documentation and tests, but I'm opening this to allow discussion of implementation details. I'm flexible regarding the scope of this PR, if it feels like `Group` doesn't belong here then we'll drop it. 

I'm a bit low on free dev time for the time being so this might progress at a snails pace but we'll get there.

### DIFF
--- a/cadquery/cq.py
+++ b/cadquery/cq.py
@@ -4373,8 +4373,10 @@ class Workplane(object):
 
         if isinstance(item, Iterable):
             rv = self.newObject(self.objects[i] for i in item)
-        else:
+        elif isinstance(item, slice):
             rv = self.newObject(self.objects[item])
+        else:
+            rv = self.newObject([self.objects[item]])
 
         return rv
 
--- a/cadquery/cq.py
+++ b/cadquery/cq.py
@@ -4382,36 +4382,36 @@ class Workplane(object):
 
     def filter(self: T, f: Callable[[CQObject], bool]) -> T:
         """
-        TODO
-        :param f:
-        :return:
+        Filter items using a boolean predicate.
+        :param f: Callable to be used for filtering.
+        :return: Workplane object with filtered items.
         """
 
         return self.newObject(filter(f, self.objects))
 
     def map(self: T, f: Callable[[CQObject], CQObject]):
         """
-        TODO
-        :param f: 
-        :return: 
+        Apply a callable to every item separately.
+        :param f: Callable to be applied to every item separately.
+        :return: Workplane object with f applied to all items.
         """
 
         return self.newObject(map(f, self.objects))
 
     def apply(self: T, f: Callable[[Iterable[CQObject]], Iterable[CQObject]]):
         """
-        TODO
-        :param f: 
-        :return: 
+        Apply a callable to all items at once.
+        :param f: Callable to be applied.
+        :return: Workplane object with f applied to all items.
         """
 
         return self.newObject(f(self.objects))
 
     def sort(self: T, key: Callable[[CQObject], Any]) -> T:
         """
-        TODO
-        :param key:
-        :return:
+        Sort items using a callable.
+        :param key: Callable to be used for sorting.
+        :return: Workplane object with items sorted.
         """
 
         return self.newObject(sorted(self.objects, key=key))
--- a/cadquery/cq.py
+++ b/cadquery/cq.py
@@ -4481,10 +4481,11 @@ class Workplane(object):
         self: T, f: Union[Callable[[T], T], Callable[[T], None], Callable[[], None]]
     ):
         """
-        Invoke a callable mapping Workplane to Workplane, Workplane to None or
-        without any parameters. In the two latter cases self is returned.
+        Invoke a callable mapping Workplane to Workplane or None. Supports also
+        callables that take no arguments such as breakpoint. Returns self if callable
+        returns None.
         :param f: Callable to be invoked.
-        :return: Workplane object
+        :return: Workplane object.
         """
 
         if isbuiltin(f):
@@ -4501,7 +4502,7 @@ class Workplane(object):
             if res is not None:
                 rv = res
         else:
-            raise ValueError("Provided function {f} accepts too many arguemnts")
+            raise ValueError("Provided function {f} accepts too many arguments")
 
         return rv
 
--- a/cadquery/cq.py
+++ b/cadquery/cq.py
@@ -35,7 +35,7 @@ from typing import (
     Dict,
 )
 from typing_extensions import Literal
-from inspect import Parameter, Signature
+from inspect import Parameter, Signature, isbuiltin
 
 
 from .occ_impl.geom import Vector, Plane, Location
@@ -4487,7 +4487,11 @@ class Workplane(object):
         :return: Workplane object
         """
 
-        arity = f.__code__.co_argcount  # NB: this is not understood by mypy
+        if isbuiltin(f):
+            arity = 0  # assume 0 arity for builtins; they cannot be introspected
+        else:
+            arity = f.__code__.co_argcount  # NB: this is not understood by mypy
+
         rv = self
 
         if arity == 0:
--- a/cadquery/cq.py
+++ b/cadquery/cq.py
@@ -4369,6 +4369,77 @@ class Workplane(object):
                 _selectShapes(self.objects)
             )._repr_javascript_()
 
+    def __getitem__(self, item):
+        return self.newObject(self.objects[item])
+
+    def filter_objects(self: T, f: Callable[[CQObject], bool]) -> T:
+        """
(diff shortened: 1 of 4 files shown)


adam-urbanczyk:
Thanks, it too me a while to get to this. I'll do some rework. I'd like to descope group for now, feels like too big of a change.

michaelgale:
I like the intent behind `cq_filter` and the proposed changes.  I wonder if these features are in fact  already available with the use of the `Selector` classes.  A big clue is the fact that any derived sub-class of `Selector` must implement their own `filter` method.  In my [`cq-kit`](https://github.com/michaelgale/cq-kit/) package I extend `Selector` with a multitude of utility selectors similar to the proposed features such as:
`EdgeLengthSelector`, `WireLengthSelector`, `RadiusSelector`, `EdgeCountSelector`, `SameLengthAsObjectSelector`, `SharedVerticesWithObjectSelector`, and much more

With `Selector` classes you can effectively *filter* any CQ object(s)  with the added benefit that `Selector` classes can be usefully combined logically and arithmetically since they implement operator overloading of `+`, `-`, `and`, `not`, etc.

adam-urbanczyk:
`filter` is indeed a quick and dirty way to implement selectors. You'd use it for one-offs and implement a proper `cq.Selector` for something more reusable. `map` and `apply` are for something else.

adam-urbanczyk:
@jmwright @lorenzncode @voneiden what do you think about it? I'd say this would be it (minus docs and groupby).

I'm still wondering if we should allow to map via `Callable[[Iterable[CQObject]], Sequence[CQObject]]` too. Regarding `groupby` it could be implemented using `cq.Compound` so that no new classes need to be created.

voneiden:
Looks pretty neat and clean to me, I like it.

Trivial: do you mind if I tidy up the commits a bit? I'd like to amend my original commit message at least since it's very ambiguous.

adam-urbanczyk:
I'm going to squash anyhows, so why not.

Additional idea: allow callbacks with 0 arity or no rv for debugging (if multimethod allows to dispatch on that). Use case would be quick debugging.
```python
w.apply(breakpoint).apply(show)
```

voneiden:
Alright, if gets squashed then I suppose it's fine as is with the surrounding context. 

The idea you presented seems like it could be useful, but IMO the current interface is very clean and easy to understand. I'd slightly lean towards having a separate method for this use case but no strong feelings. Multimethod could fine too, if the docstring explains the secondary behaviour in layman terms.

lorenzncode:
Looks good with minimal testing so far! I'll play with it more.  

> if we should allow to map via Callable[[Iterable[CQObject]], Sequence[CQObject]] too.

Does this mean something like the following works without an explicit call to `compound`:

```py
import cadquery as cq
from cadquery.occ_impl.shapes import *


def mycallback(obj):
    if obj.Center().x > 0:
        # return compound([obj.moved((0, -2, 0)), obj.moved((0, 2, 0))])
        return [obj.moved((0, -2, 0)), obj.moved((0, 2, 0))]
    else:
        return obj


r = cq.Workplane().rarray(20, 20, 2, 2).box(1, 1, 1, combine=False).map(mycallback)
```

adam-urbanczyk:
> > if we should allow to map via Callable[[Iterable[CQObject]], Sequence[CQObject]] too.
> 
> Does this mean something like the following works without an explicit call to `compound`:
> 

This would actually imply `Callable[[Iterable[CQObject]], Union[CQObject, Sequence[CQObject]]` which is slightly more generic. Do you find it useful?

lorenzncode:
> This would actually imply `Callable[[Iterable[CQObject]], Union[CQObject, Sequence[CQObject]]` which is slightly more generic. Do you find it useful?

If the more generic form can work it might be useful.   It looks like there is a workaround for this case if not.

adam-urbanczyk:
In the meantime I added `invoke` for calling breakpoint or show inline. It can be also used to implement plugins without monkey patching.

adam-urbanczyk (approved):

adam-urbanczyk:
I added some docs and I think this is ready to be merged.  Any thoughts @lorenzncode @voneiden @jmwright ?

adam-urbanczyk (approved):

lorenzncode (approved):
`invoke` is another very useful addition!  I like the new option for extending CQ.  The docs are clear.

Regarding `invoke` with builtin and arity 0 - my understanding is `breakpoint` is the primary intended use case (perhaps other special cases).

cadquery/cq.py
```diff
@@ -4430,6 +4430,81 @@ def _repr_javascript_(self) -> Any:
...
+        if arity == 0:
+            f()  # type: ignore
+        elif arity == 1:
+            res = f(self)  # type: ignore
+            if res is not None:
+                rv = res
+        else:
+            raise ValueError("Provided function {f} accepts too many arguemnts")
+
+        return rv
```
voneiden, line 4506: Not requesting this to be changed, but I think something like this would be a bit more mundane...? 

```suggestion
        rv = self
        
        try:
            res = f(self)  # type: ignore
        except TypeError:
            res = f()  # type: ignore
            
        if res is not None:
            rv = res
            
        return rv 
```

cadquery/cq.py
```diff
@@ -4430,6 +4430,81 @@ def _repr_javascript_(self) -> Any:
...
+        rv = self
+
+        if arity == 0:
+            f()  # type: ignore
+        elif arity == 1:
+            res = f(self)  # type: ignore
+            if res is not None:
+                rv = res
+        else:
+            raise ValueError("Provided function {f} accepts too many arguemnts")
```
voneiden, line 4504: typo 

```suggestion
            raise ValueError("Provided function {f} accepts too many arguments")
```

cadquery/cq.py
```diff
@@ -4430,6 +4430,81 @@ def _repr_javascript_(self) -> Any:
...
+        return self.newObject(sorted(self.objects, key=key))
+
+    def invoke(
+        self: T, f: Union[Callable[[T], T], Callable[[T], None], Callable[[], None]]
+    ):
+        """
+        Invoke a callable mapping Workplane to Workplane, Workplane to None or
+        without any parameters. In the two latter cases self is returned.
+        :param f: Callable to be invoked.
+        :return: Workplane object
```
voneiden, line 4487: Consistency 

```suggestion
        :return: Workplane object.
```

doc/extending.rst
```diff
@@ -193,3 +195,42 @@ This ultra simple plugin makes cubes of the specified size for each stack point.
...
+            cq.Workplane("XY")
+            .box(6.0, 8.0, 0.5)
+            .faces(">Z")
+            .rect(4.0, 4.0, forConstruction=True)
+            .vertices()
+            .invoke(makeCubes(1.0))
+            .combineSolids()
+        )
+
+Such an approach if more friendly for auto-completion and static analysis tools.
```
voneiden, line 236: typo? 

```suggestion
Such an approach is more friendly for auto-completion and static analysis tools.
```

voneiden (commented):
Sorry for the slow review, I'm in :palm_tree:  mode. :smiley:

jmwright (approved):
Looks good to me. Thanks @voneiden !

cadquery/cq.py
adam-urbanczyk: Thanks, but I find mine more clear. Also it has slightly different semantics, f could throw for other reason than wrong number of parameters.

adam-urbanczyk:
Merging, thanks @voneiden !

PR #1523, opened 2024-02-16
Fixed gltf export instructions per #1508

### LINKED ISSUE #1508: "Importing and Exporting Files" Docs wrongly state that one can export glTF files with exporters.export()
On the docs page [Importing and Exporting Files](https://cadquery.readthedocs.io/en/latest/importexport.html) exporting to glTF is said to be done like so:

```python
import cadquery as cq

result = cq.Workplane().box(10, 10, 10)

exporters.export(result, "/path/to/file/object.[file_extension]")
```

That doesn't work and you instead need to use the save function on an assembly like this:
```python
import cadquery as cq

result = cq.Workplane().box(10, 10, 10)
assembly = cq.Assembly(result)
cq.assembly.Assembly.save(assembly, "/path/to/file/object.[file_extension]" )

```

Docs should probably be updated with a section for this.

jmwright:
Closes #1508 

This is a straight-forward documentation addition, so I think we can merge with one review.


adam-urbanczyk (approved):

adam-urbanczyk:
Thanks again @jmwright !

PR #1525, opened 2024-02-18
Isolate STEPControl_Reader ReadFile

lorenzncode:
Interface to OCCT Message

### DIFF
--- a/cadquery/__init__.py
+++ b/cadquery/__init__.py
@@ -21,6 +21,7 @@ from .occ_impl.shapes import (
 )
 from .occ_impl import exporters
 from .occ_impl import importers
+from .occ_impl import message
 
 # these items are the common implementation
 
@@ -76,4 +77,5 @@ __all__ = [
     "Selector",
     "plugins",
     "Sketch",
+    "message",
 ]
--- a/cadquery/occ_impl/importers/__init__.py
+++ b/cadquery/occ_impl/importers/__init__.py
@@ -7,6 +7,8 @@ from OCP.STEPControl import STEPControl_Reader
 from ... import cq
 from ..shapes import Shape
 from .dxf import _importDXF
+from .. import message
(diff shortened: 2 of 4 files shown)


lorenzncode:
This would allow to process the `STEPControl_Reader` object without continuing to import the STEP file for example to extract header info.  Another use case might be to abort STEP import if certain alerts or messages are issued on read.  There may be other more general purpose OCCT message reporting uses.

```py
import cadquery as cq

report = cq.message.Message.add_report()
reader = cq.importers.readStep("BROKEN_FILE.STEP")
alerts = report.GetAlerts(cq.message.Level.info.value)

if alerts:
    raise ValueError("STEP file read resulted in alerts, skip importStep")
else:
    res = cq.importers.importStep(reader)
```

```py
import cadquery as cq

def import_step_abort_on_alert(filename, report=None):
    """
    Import STEP
    Abort in case of OCCT alert info messages

    This is NOT guaranteed to eliminate all crashes.
    It is an example based a known corrupted STEP file.  See issue #1521.

    It *might* help filter some problematic STEP files.
    In other cases perhaps trace level messages must be inspected or there 
    may not be any way to know up front whether a STEP file is problematic.
    """

    if report is None:
        report = cq.message.Message.add_report()

    report.Clear(cq.message.Level.info.value)
    reader = cq.importers.readStep(filename)
    alerts = report.GetAlerts(cq.message.Level.info.value)

    if alerts:
        raise ValueError("abort STEP import")
    else:
        res = cq.importers.importStep(reader)

    return res


step_files = ("BROKEN_FILE.STEP", "good.step", "BROKEN_FILE.STEP", "good.step")

report = cq.message.Message.add_report()
for f in step_files:
    try:
        res = import_step_abort_on_alert(f, report)
    except ValueError:
        res = None
```

PR #1526, opened 2024-02-26
Added an import for show_object.

pjuangph:
This now works with VS Code


adam-urbanczyk:
The idea is that show_object is provided by the execution environment. It was not ommited by mistake.

pjuangph:
It would be helpful to include an example of importing show_object. I wasn't able to find it in the documentation. I did find the statement about execution environment but more help would be great.

Issue #1153, opened 2022-09-14
Workplane.save()


roipoussiere:
I like the simplicity of [`Assembly.save()`](https://cadquery.readthedocs.io/en/latest/classreference.html#cadquery.Assembly.save):

```py
asm = ( cq.Assembly()
    .add(box)
    .save('box.step')
)
```

To export a model, in the other hand, the syntax is not as nice:

```py
box = cq.WorkPlane().box(1, 2, 3)
cq.exporters.export(box, 'box.step')
```

I wonder if the assembly export syntax could be applied on a model as well, for consistency and simplicity:

```py
box = ( cq.WorkPlane()
    .box(1, 2, 3)
    .save('box.step')
)
```

adam-urbanczyk:
It would be nice indeed. @jmwright @lorenzncode would you agree?

lorenzncode:
Yes, that might be convenient and I think it could support saving to BREP file too!

jmwright:
+1

adam-urbanczyk:
Workplane, Sketch, Shape and Assembly support `export`.

PR #1531, opened 2024-03-06
define a Workplane.export() (#1153)

aavogt:
I tried to put it in cq.py under class Workplane, but there were problems with circular imports.

The docstring is not quite right since `:w:` should probably be `:self:`.

```cq.Workplane.export.__doc__
'\n    Export Workplane or Shape to file. Multiple entities are converted to compound.\n\n    :param w:  Shape or Workplane to be exported.\n    :param fname: output filename.\n    :param exportType: the exportFormat to use. If None will be inferred from the extension. Default: None.\n    :param tolerance: the deflection tolerance, in model units. Default 0.1.\n    :param angularTolerance: the angular tolerance, in radians. Default 0.1.\n    :param opt: additional options passed to the specific exporter. Default None.\n ```


adam-urbanczyk:
Thanks, `Workplane.export` was implemented in the meantime.

PR #1549, opened 2024-03-24
Add support for 2D and 3D Wire fillet

dov:
Add support for 2D and 3D Wire fillet through OCCT ChFi2d_FilletAPI

### DIFF
--- a/cadquery/occ_impl/shapes.py
+++ b/cadquery/occ_impl/shapes.py
@@ -2395,6 +2395,7 @@ class Wire(Shape, Mixin1D):
         :return: A wire with filleted corners
         """
         edges = list(self)
+        all_vertices = self.Vertices()
         newEdges = []
         currentEdge = edges[0]
 
@@ -2413,8 +2414,7 @@ class Wire(Shape, Mixin1D):
             # Check conditions for skipping fillet:
             #  1. The edges are are parallell
             #  2. The vertex is not in the vertices white list
-            vertex = currentEdge.startPoint()
-            if normalDir.Length == 0 or (vertex not in verticesSet):
+            if normalDir.Length == 0 or (all_vertices[i + 1] not in verticesSet):
                 newEdges.append(currentEdge)
                 currentEdge = nextEdge
                 continue
@@ -2446,6 +2446,47 @@ class Wire(Shape, Mixin1D):
         newEdges.append(currentEdge)
 
         return Wire.assembleEdges(newEdges)
+
+
+    def firstVertex(self) -> Vertex:
+        """
+        First vertex of the wire.
+        """
+
+        exp = BRepTools_WireExplorer(self.wrapped)
+
+        return Vertex(exp.CurrentVertex())
+
+
+    def lastVertex(self) -> Vertex:
+        """
+        Last vertex of the wire.
+        """
+
+        exp = BRepTools_WireExplorer(self.wrapped)
+        while exp.More():
+            exp.Next()
+
+        return Vertex(exp.CurrentVertex())
+
+
+    def Vertices(self) -> List[Vertex]:
+        """
+        Ordered list of vertices of the wire.
+        """
+
+        rv = []
+
+        exp = BRepTools_WireExplorer(self.wrapped)
+        rv.append(Vertex(exp.CurrentVertex()))
+
+        while exp.More():
+            exp.Next()
+
+        rv.append(Vertex(exp.CurrentVertex()))
+
+        return rv 
+
 
     def __iter__(self) -> Iterator[Edge]:
         """
--- a/cadquery/occ_impl/shapes.py
+++ b/cadquery/occ_impl/shapes.py
@@ -2385,22 +2385,18 @@ class Wire(Shape, Mixin1D):
         self, radius: float, vertices: Optional[Iterable[Vertex]] = None
     ) -> "Wire":
         """
-
         Apply 2D or 3D fillet to a wire
-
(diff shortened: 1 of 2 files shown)


adam-urbanczyk:
Hey, thanks, but could you _not_ force push

dov:
@adam-urbanczyk 
Regarding force push. Sorry, I was (ab-)using it in order to get the CI support. I'll make sure in the future to clean all the CI errors before creating pull requests.

Regarding the PR itself, I added a vertices parameter, but in terms of a list of vertex indices. I was wondering whether it might be better to give the coordinates. What do you think?

adam-urbanczyk:
Every push will trigger CI, am I missing something?

cadquery/occ_impl/shapes.py
```diff
@@ -2379,6 +2381,65 @@ def chamfer2D(self, d: float, vertices: Iterable[Vertex]) -> "Wire":
 
         return f.chamfer2D(d, vertices).outerWire()
 
+    def fillet(self, radius: float, vertices: Optional[Iterable[int]] = None) -> "Wire":
```
adam-urbanczyk, line 2384: I'd propose to use `Optional[Iterable[Vertex]]` here

cadquery/occ_impl/shapes.py
```diff
@@ -2379,6 +2381,75 @@ def chamfer2D(self, d: float, vertices: Iterable[Vertex]) -> "Wire":
...
+
+        :return: A wire with filleted corners
+        """
+        edges = list(self)
+        newEdges = []
+        currentEdge = edges[0]
+
+        # Make vertices into a set so that we can query for membership
+        # in O(log n).
+        if vertices is not None:
```
adam-urbanczyk, line 2403: you should simply use vertices as-is an convert them to set

cadquery/occ_impl/shapes.py
```diff
@@ -2379,6 +2381,75 @@ def chamfer2D(self, d: float, vertices: Iterable[Vertex]) -> "Wire":
...
+
+            # Create a plane that is spanned by currentEdge and nextEdge
+            currentDir = currentEdge.tangentAt(1)
+            nextDir = nextEdge.tangentAt(0)
+            normalDir = currentDir.cross(nextDir)
+
+            # Check conditions for skipping fillet:
+            #  1. The edges are are parallell
+            #  2. The vertex is not in the vertices white list
+            vertex = currentEdge.startPoint().toTuple()
```
adam-urbanczyk, line 2417: remove `toTuple()`

cadquery/occ_impl/shapes.py
dov: Sorry. I don't understand.  `Vertex` isn't immutable, and you can't put it into python set. Which is why I'm doing the casting to an immutable tuple. Did I miss something?

adam-urbanczyk: I applied my suggestions, let's see if it works

dov: I checked the immutability aspect of Vertex and it seems I was wrong. It's just a wrapper for `TopoDS_Vertex` which seems to be immutable. So using sets on Vertex is ok. 

However using vertices directly without casting to Vertex or tuple will fail if the user passes a list for coordinates to vertices.

I.e. the following code worked before your change, but will fail after your change:

```
        points = [
            [0.000, 0.000, 0.000],
            [-0.287, 1.183, -0.592],
            [-1.404, 4.113, -2.787],
            [-1.332, 1.522, 0.553],
            [7.062, 0.433, -0.097],
            [8.539, -0.000, -0.000],
        ]
        wire = Wire.makePolygon(points, close=False)

        # Fillet a single vertex
        wfillet = wire.fillet(radius=0.560, vertices=[points[1]])
        assert len(wfillet.Edges()) == len(points)

```

dov: I'd rather do a potentially redundant cast in the implementation to catch a casual misuse of the API than having the code raise a cryptic exception. But if it's fine with you, I'll yield.

adam-urbanczyk: The API is specified by the type annotations, IMO nothing is cryptic here. The general vision is that you use selectors to specify fillets and chamfers.

adam-urbanczyk (approved):

dov:
:partying_face:

adam-urbanczyk:
FYI: in the end I refactored `Wire.Vertices()` so that the rv is ordered.

cadquery/occ_impl/shapes.py
```diff
@@ -2379,6 +2381,90 @@ def chamfer2D(self, d: float, vertices: Iterable[Vertex]) -> "Wire":
...
+        for i in range(len(edges) - 1):
+            nextEdge = edges[i + 1]
+
+            # Create a plane that is spanned by currentEdge and nextEdge
+            currentDir = currentEdge.tangentAt(1)
+            nextDir = nextEdge.tangentAt(0)
+            normalDir = currentDir.cross(nextDir)
+
+            # Check conditions for skipping fillet:
+            #  1. The edges are are parallell
```
jmwright, line 2411:
```suggestion
            #  1. The edges are are parallel
```

jmwright (approved):
Thanks @dov there is just the one comment suggestion from me.

dov:
> Thanks @dov there is just the one comment suggestion from me.

Thanks @jmwright . I fixed it and also removed another typo in the comment.

adam-urbanczyk:
Thanks @dov !

PR #1560, opened 2024-04-12
Fix union with None

### LINKED ISSUE #1536: Workplane.union error with None
I use the following instead of `union`:

```python
import cadquery as cq
cq.Workplane.union2 = lambda self, x: x is None and self or self.union(x)
```
This avoids a `ValueError: Cannot union type '<class 'NoneType'>'`. `cq.Workplane().union(None)` makes that exception. The type hint says that union(None) should work.

I use it to share code for two similar parts:
```python
def mkresult(withBase=True):
    return (cq.Workplane("XY").circle(68).extrude(2)
        .union2(withBase and base or None)
        ....
```

adam-urbanczyk:
Closes #1536

### DIFF
--- a/cadquery/cq.py
+++ b/cadquery/cq.py
@@ -3370,6 +3370,8 @@ class Workplane(object):
             self._mergeTags(toUnion)
         elif isinstance(toUnion, (Solid, Compound)):
             newS = [toUnion]
+        elif toUnion is None:
+            newS = []
         else:
             raise ValueError("Cannot union type '{}'".format(type(toUnion)))
 
--- a/tests/test_cadquery.py
+++ b/tests/test_cadquery.py
@@ -2797,6 +2797,15 @@ class TestCadQuery(BaseTest):
         objects2 = objects1.add(objects2).combine(glue=True, tol=None)
         self.assertEqual(11, objects2.faces().size())
 
+    def testUnionNoArgs(self):
+        # combine using union with no arguments
+        s = Workplane(Plane.XY())
+
+        objects1 = s.rect(2.0, 2.0).extrude(0.5)
+        objects2 = s.rect(1.0, 1.0).extrude(0.5).translate((0, 0, 0.5))
+        objects2 = objects1.add(objects2).union(glue=True, tol=None)
+        self.assertEqual(11, objects2.faces().size())
+
     def testCombineSolidsInLoop(self):
         # duplicates a memory problem of some kind reported when combining lots of objects
         s = Workplane("XY").rect(0.5, 0.5).extrude(5.0)


michaelgale:
Wow!  Really grateful that this has finally been changed.  I can't tell you how much boilerplate code I've written to handle the inevitable union to "None" scenario when making composite solids/geometries from iteration.

adam-urbanczyk:
Thanks @michaelgale ! Was there something stopping you from opening an issue or PR? Or did we miss it somehow?

michaelgale:
> Thanks @michaelgale ! Was there something stopping you from opening an issue or PR? Or did we miss it somehow?

I am somewhat hesitant to suggest changes to a mature core API since these types of changes can have unpredictable consequences for other users.  Since this issue didn't represent a bug or unexpected behaviour, I didn't think to ever raise a PR or issue.  Since CQ is so easily extended with plugins, I tend to rely on this mechanism to introduce syntactic shortcuts and convenience functions to streamline my integrations with CQ.  In future, I might lower my threshold for raising issues and/or contributing changes with a PR.  Thanks again to the CQ team for maintaining this project.

lorenzncode (approved):
Hey just one comment regarding boilerplate.  This behavior results in additional parents which may be different than current boilerplate.  As @michaelgale mentions iteration, I suppose it depends on size of the iteration or sparseness whether there is any significant noticeable impact.


```py
import cadquery as cq

r0 = cq.Workplane().cylinder(1, 10)

for i in range(1, 101):
    if i == 50:
        obj = cq.Workplane().box(1, 1, 10)
    else:
        obj = None

    r0 = r0.union(obj)
```

image

adam-urbanczyk:
@lorenzncode I'm not sure that I understand. Currently union with `None` results in an error.

jmwright:
Looks good, thanks @adam-urbanczyk

lorenzncode:
@adam-urbanczyk  Right, I'm only pointing out that skipping `union` to avoid the error is not *exactly* the same as this fix. 

In #1536 , `union2` is defined to avoid the error by skipping the call to `union`.  

In the PR branch, calling `union` with `None` extends the Workplane chain and does not result in error.  

The result at the end of the chain is the same in either case only that the length of the chain is different.

adam-urbanczyk:
> @adam-urbanczyk Right, I'm only pointing out that skipping `union` to avoid the error is not _exactly_ the same as this fix.
> 
> In #1536 , `union2` is defined to avoid the error by skipping the call to `union`.

That is the intention, `.union()` will in general union items on the stack, so creating a new object is desired. Shall I merge?

lorenzncode:
That makes sense to me.  +1 to merge.  Thanks @adam-urbanczyk

PR #1563, opened 2024-04-16
Make offset2D act on objects instead of pending wires

blastrock:
Hi,

This PR is a bugfix, but I'm not sure this is the desired behavior, or even if this is a bug in the first place.

Here's a script to describe my issue:

```py
    cq.Workplane("XY")
    .rect(5, 5)
    # .offset2D(1)
    .workplane(offset=5)
    .rect(3, 5)
    .workplane(offset=5)
    .rect(10, 20)
    # .offset2D(1)
    .consolidateWires()
    # .loft()
```

This object contains 3 rects on top of each other. Now if I uncomment the `offset2D` calls, I expect them to act on the single `rect` just before each call, so that the bottom and top rects get rounded, and the middle one is left untouched.

What actually happens is that `offset2D` calls `_consolidateWires` and acts on *all* pending wires of all the workplanes on this stack, so the second `offset2D` will add an offset to all 3 rects (the first one will be offset twice).

This patch allows `offset2D` to not act on all pending wires but only on current objects. This gives me the desired result, the middle rect is not offset and has sharp edges:

image
image

---

There are a few questions I have about this:

- Is this fix needed, or is there a way to accomplish what I want without it?
- Is this the desired fix for this issue?
- This implementation of `offset2D` acts only on already-consolidated wires. Should it try to consolidate all the edges that are in `self.object` too, to keep a behavior similar to before this patch?

Note that tests are broken, I am waiting for more input about whether this is the right thing to do before trying to fix them.

### DIFF
--- a/cadquery/cq.py
+++ b/cadquery/cq.py
@@ -4295,7 +4295,7 @@ class Workplane(object):
         :return: CQ object with resulting wire(s).
         """
 
-        ws = self._consolidateWires()
+        ws = [el for el in self.objects if isinstance(el, Wire)]
         rv = list(chain.from_iterable(w.offset2D(d, kind) for w in ws))
 
         self.ctx.pendingEdges = []
@@ -4304,7 +4304,9 @@ class Workplane(object):
                 wire.forConstruction = True
             self.ctx.pendingWires = []
         else:
-            self.ctx.pendingWires = rv
+            for w in ws:
+                self.ctx.pendingWires.remove(w)
+            self.ctx.pendingWires.extend(rv)
 
         return self.newObject(rv)
 


dov:
This can actually be solved by my PR #1574 which allows adding wires directly to the WorkPlane. With this accepted, you could then create a closed wire for each of your three rectangles, and  then `offset2D` (or use `Wire.fillet()` with #1573) only the first and last rectangle. 

Actually even without #1574 being accepted you can easily get this functionality by monkey patching:

```python
def addWire(self, wire):
    """Add a wire at the current point"""
    return self.eachpoint(lambda v : wire.moved(v))

cq.Workplane.addWire = addWire

```

adam-urbanczyk:
Please use `cq.Sketch` for this kind of use cases: https://cadquery.readthedocs.io/en/latest/sketch.html#workplane-integration

blastrock:
Thanks for the answers! I struggled, but I managed to work it out with the sketch API!

In case someone wants to do the same: you need to do a `.wires().offset(x).clean()` to get the right sketch, and then you need to place each sketch with `placeSketch` as shown it the last example of [this page](https://cadquery.readthedocs.io/en/latest/sketch.html#workplane-integration)

PR #1570, opened 2024-04-24
Describing the units of the result

bc0n:
Since CQ does not know the material density, the result of cq.Shape.matrixOfInertia() is a specific inertia which needs to be multiplied by density to get the correct result.


adam-urbanczyk:
Thanks @bc0n !

Docs update - merging.

Issue #1395, opened 2023-08-31
Support eachpoint with `cq.Shape` and `cq.Workplane` in addition to callbacks


adam-urbanczyk:
E.g.
```python
b = cq.Workplane().box(2,2,2)
s = cq.Workplane().sphere(0.5)

res = b.faces().eachpoint(s)
```

dov:
@adam-urbanczyk  In continuation of your comment on #1574 , I would add `cq.Wire` in addition to `cq.Shape` and `cq.Workplane` to this issue.

adam-urbanczyk:
`cq.Wire` **_is a_** `cq.Shape`

dov:
@adam-urbanczyk I got it working for a `Shape` by renaming the argument `arg` and dispatching on its type. However I need some guidance about what semantics you expect for a Workspace. Just moving the workspace by the following lambda obviously does not work:

```python
        if isinstance(arg, Workplane):
            wp = arg
            callback = lambda v : wp.move(v)
```

adam-urbanczyk:
Something like this:

```python
v.moved(loc) for v in arg.vals() if isinstance(v, Shape)
```

dov:
> Something like this:
> 
> ```python
> v.moved(loc) for v in arg.vals() if isinstance(v, Shape)
> ```

Thanks @adam-urbanczyk . In my PR I used the following. Let me know this is what you intended.

```python
                res = [
                    v.moved(p).move(loc)
                    for v in arg.vals()
                    for p in pnts
                    if isinstance(v, Shape)
                ]
```

lorenzncode:
I think this can be closed.  One follow up issue though is `cutEach` typing.  

This now works but fails mypy:
```py
b = cq.Workplane().box(2, 2, 2)
s = cq.Workplane().sphere(0.5)

res = b.faces().cutEach(s)
```

```
test2.py:7: error: Argument 1 to "cutEach" of "Workplane" has incompatible type "Workplane"; expected "Callable[[Location], Shape]"  [arg-type]
```

adam-urbanczyk:
We can probably deprecate `cutEach` it is the same as `eachpoint(..., mode='cut')`.

lorenzncode:
Yes, agree deprecating `cutEach` makes sense.

PR #1574, opened 2024-04-28
Add Workplane.addWire() method for simplifying adding wires to Workplanes

dov:
This PR adds th emethod `Workplane.addWire()`, which simplifies adding a wire to each of the points in the Workplane stack.

It simplifies workflows like adding rounded rectangle holes into a baseplate as shown in the example below.

If you think this is better suited as a plugin, please let me know.

```python
def my_closed_wire_generator():
     '''Create a square with rounded corners'''
     s = 1
     points = [(-s,-s,0), (s,-s),(s,s),(-s,s)]
     return cq.Wire.makePolygon(points, close=True).fillet(0.3)
      
wire = my_closed_wire_generator() # Arbitrary routine generating a closed wire

# Create a workplane with four points on the stack
result = cq.Workplane('XY').box(10,10,1).faces('>Z').rect(7, 7, forConstruction=True).vertices()

# Add the wire to each of the points and cut
result =  result.addWire(wire).cutThruAll()

```
Here's the output:

rounded-square-holes

### DIFF
--- a/cadquery/cq.py
+++ b/cadquery/cq.py
@@ -4396,6 +4396,12 @@ class Workplane(object):
 
         return self.newObject(rv)
 
+    def addWire(self: T, wire: Wire) -> T:
+        """
+        Add a wire for each point on the stack.
+        """
+        return self.eachpoint(lambda v: wire.moved(v))
+
     def _repr_javascript_(self) -> Any:
         """
         Special method for rendering current object in a jupyter notebook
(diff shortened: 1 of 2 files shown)


adam-urbanczyk:
No need to extend the API, just use `eachpoint`. If you want to make it more user friendly, here is an open issue: https://github.com/CadQuery/cadquery/issues/1395.

PR #1578, opened 2024-04-29
Add shape and workplane support to the Workplane.eachpoint() function. Issue #1395

dov:
This PR solves issue #1395 and adds the ability to use the `Workplane.eachpoint()` function with a `Shape` and a `Workplane`.

Here are is an example for using a wire to do a `cutThru()`:

```python
wire = Wire.makeCircle(1, (0, 0, 0), (0, 0, 1))
result = (
    Workplane("XY")
    .box(10, 10, 1)
    .faces(">Z")
    .rect(7, 7, forConstruction=True)
    .vertices()
    .eachpoint(wire)
    .cutThruAll()
)
```

And here is an example for using another workplane, adding a sphere to the center of each face:
```python
box = Workplane().box(2, 2, 2)
sph = Workplane().sphere(1.0)
# Place the sphere in the center of each box face
r = box.faces().eachpoint(sph, combine=True)
```

### DIFF
--- a/cadquery/cq.py
+++ b/cadquery/cq.py
@@ -2500,7 +2500,7 @@ class Workplane(object):
         clean: bool = True,
     ) -> T:
         """
-        Same as each(), except arg is translated by the positions on the stack. If arg is the a callback function, then the the function is called for each point on the stack, and the resulting shape is used.
+        Same as each(), except arg is translated by the positions on the stack. If arg is a callback function, then the function is called for each point on the stack, and the resulting shape is used.
         :return: CadQuery object which contains a list of  vectors (points ) on its stack.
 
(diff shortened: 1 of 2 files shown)


DovXjet:
@adam-urbanczyk Ok. I just did. It isn't trivial since the rest of the files are not compliant. 

(However, I just found "darken", which makes it easy).

jmwright:
@dov The following command should show just what is needed to fix your changes. 
```
black --diff --check .
```

adam-urbanczyk (approved):

jmwright (approved):
Thanks @dov I just had a couple of corrections to a comment. Otherwise it looks good.

jmwright:
Thanks @dov

PR #1579, opened 2024-05-01
fix #786 Asymmetric Chamfer Error

### LINKED ISSUE #786: Asymmetric Chamfer Error
Asymmetric chamfer is not applied consistently.  The chamfers on the top of this ring look correct; however, the inner chamfer on the bottom is reversed.
chamferBug

Error found on cadquery master.
```python
import cadquery as cq

chamferBody = (
    cq.Workplane("XY")
    .circle(8)
    .circle(4)
    .extrude(4)
    .faces(">Z").chamfer(0.5,1.5)
    .faces("<Z").chamfer(0.5,1.5)
)
if "show_object" in locals():
    show_object(chamferBody,name="chamferBody")
```

kolod:
cadquery uses random edge face when call into chamfer_builder.Add()

Fix uses face on which edge selected in chamfer_builder.Add() call

Example:
```
Workplane().box(10, 10, 10).faces("<X").chamfer(1, 2)
```
image

If both faces of the edge are selected chamfer now uses length2 for length to prevent unexpected behavior.

Example:
```
Workplane().box(10, 10, 10).faces("<Y or <X").edges("|Z").chamfer(1, 2)
```
image

### DIFF
--- a/cadquery/cq.py
+++ b/cadquery/cq.py
@@ -1320,6 +1320,7 @@ class Workplane(object):
         :param length2: optional parameter for asymmetrical chamfer
         :raises ValueError: if at least one edge is not selected
         :raises ValueError: if the solid containing the edge is not in the chain
+        :raises ValueError: if the asymmetric mode and can't find face in stack
         :returns: CQ object with the resulting solid selected.
 
         This example will create a unit cube, with the top edges chamfered::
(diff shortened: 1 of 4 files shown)


adam-urbanczyk:
Thanks for the PR, but looks like you still fail on mypy (annotations) and call `show` from the test which causes an error on azure. We'll probably need to discuss the proposed semantics too.

adam-urbanczyk:
@kolod do you intend to work on this?

PR #1583, opened 2024-05-06
Reformat all py files with black

dov:
Currently the code isn't black compliant. This makes it unduly difficult to contribute, because you can't blindly run black on the changed files, but must sift through the black diff output and copy only parts that belong to the changes carried out. This pull request solves that by making all py files in the repo compliant.


adam-urbanczyk:
Please ask first and use the correct version for this project.

adam-urbanczyk:
Cf. https://github.com/CadQuery/cadquery?tab=readme-ov-file#contributing-code

dov:
@adam-urbanczyk Ok, thanks. Indeed I missed that you had forked black.

PR #1593, opened 2024-05-25
Fix workplane cylinder center when generated using a custom direction

### LINKED ISSUE #1591: Problem of center of mass when a cylinder is generated with a non-default direction
Hello, I think I am having a bug when I create a cylinder from a workplane.

```python
>>> cq.Shape.centerOfMass(cq.Workplane("XY").cylinder(radius=1, height=10, direct=cq.Vector(1, 0, 0)).val()) 
Vector: (5.0, 6.263935600236486e-18, -5.0)
>>> cq.Shape.centerOfMass(cq.Workplane("XY").cylinder(radius=1, height=10, direct=cq.Vector(0, 1, 0)).val()) 
Vector: (-6.263935600236486e-18, 5.0, -5.0)
>>> cq.Shape.centerOfMass(cq.Workplane("XY").cylinder(radius=1, height=10, direct=cq.Vector(0, 0, 1)).val()) 
Vector: (-1.554937844271053e-16, -2.7947979049946595e-17, 0.0)
```

The center of mass is centered in 0 only when the direction of the cylinder is the default one `(0, 0, 1)`.
In other cases, it is not well located. I guess it comes from the following part in the `Workplane.cylinder` method that does not take care of the direction vector.

```python
offset = Vector()
if not centered[0]:
    offset += Vector(radius, 0, 0)
if not centered[1]:
    offset += Vector(0, radius, 0)
if centered[2]:
    offset += Vector(0, 0, -height / 2)

s = Solid.makeCylinder(radius, height, offset, direct, angle)
```

kmarchais:
Fixes #1591. 

- This PR fixes the following bug when a cylinder generated from a workplane with a direction different than the default one `(0, 0, 1)` was not centered at the right position.

```python
>>> cq.Shape.centerOfMass(cq.Workplane("XY").cylinder(radius=1, height=10, direct=cq.Vector(1, 0, 0)).val()) 
Vector: (5.0, 0.0, -5.0)
>>> cq.Shape.centerOfMass(cq.Workplane("XY").cylinder(radius=1, height=10, direct=cq.Vector(0, 1, 0)).val()) 
Vector: (0.0, 5.0, -5.0)
>>> cq.Shape.centerOfMass(cq.Workplane("XY").cylinder(radius=1, height=10, direct=cq.Vector(0, 0, 1)).val()) 
Vector: (0.0, 0.0, 0.0)
```

- This PR also brings the possibility to center the cylinder when it is generated using any custom direction.

A test has been added to check these two points. It fails before the fix and passes after the fix.

The case where `centered != (True, True, True)` with a direction different than one along the x, y, or z axes is not managed. What would we like to do in this case?


lorenzncode:
>The case where centered != (True, True, True) with a direction different than one along the x, y, or z axes is not managed. What would we like to do in this case?

Something like the following should handle it.

```py
        
        if isinstance(centered, bool):
            centered = (centered, centered, centered)

        offset = Vector()
        plane = Plane((0, 0, 0), normal=direct)
        if not centered[0]:
            offset += plane.toWorldCoords((radius, 0, 0))
        if not centered[1]:
            offset += plane.toWorldCoords((0, radius, 0))
        if centered[2]:
            offset += plane.toWorldCoords((0, 0, -height / 2))

        s = Solid.makeCylinder(radius, height, offset, direct, angle)

        # We want a cylinder for each point on the workplane
        return self.eachpoint(lambda loc: s.moved(loc), True, combine, clean)
```

kmarchais:
Thanks, I prefer this than what I did but then it requires to be careful about the axis to set for centering.

To center a cylinder along its height but directed along the y axis, it requires to put the third component of `centered` to True as the following example show.

```python
c1 = cq.Shape.centerOfMass(
    cq.Workplane("XY").cylinder(
        radius=1,
        height=10,
        direct=cq.Vector(0, 1, 0),
        centered=(False, False, True),
    ).val()
) 
print(c1)
c2 = cq.Shape.centerOfMass(
    cq.Workplane("XY").cylinder(
        radius=1,
        height=10,
        direct=cq.Vector(0, 1, 0),
        centered=(False, True, False),
    ).val()
)
print(c2)
```

Results in

```python
Vector: (1.0, 0.0, 1.0)
Vector: (0.0, 5.0, 1.0)
```

Is it the expected behavior?

adam-urbanczyk:
```python
res = cq.Workplane("XY").cylinder(
        radius=1,
        height=10,
        direct=cq.Vector(0, 1, 0),
        centered=(False, False, True),
).val()
```
results in this:
afbeelding

```python
res2 = cq.Workplane("XY").cylinder(
        radius=1,
        height=10,
        direct=cq.Vector(0, 1, 0),
        centered=(False, True, False),
).val()
```

and 

in this:
afbeelding

@kmarchais how did you actually generate the reference data for the tests? Visually the result is as expected AFAICT.

kmarchais:
The test was pretty straight forward in the previous implementation, testing all the possible combinations of centering and directions.
With this implementation, I will need to take time to find how to manage it correctly.

Also I want to make sure, is the following the expected result?
```python
>>> cq.Shape.centerOfMass(cq.Workplane("XY").cylinder(radius=10, height=40, direct=cq.Vector(1, 0, 0), centered=False).val())
Vector: (20, -10, 10)
```

adam-urbanczyk:
Looks OK, but the local coordinate system of the cylinder is ill-defined with only Z vector. Is that the point you are trying to make?

afbeelding

kmarchais:
I am just not used to this radius shift and wanted to make sure that this -10 in the y direction is fine.
I will try to update the test soon.

adam-urbanczyk:
Hey @kmarchais any updates on this? Do you intend to finalize it?

kmarchais:
@adam-urbanczyk Thanks for the remainder. I'm on it.

kmarchais:
I could not find a simple way to test every combination of centered and direct parameters. I made it so that it checks different configurations:
- Main axis directions (±X, ±Y, ±Z)
- Centering configurations (all centered, none centered, mixed centering)
- Verifies correct cylinder center position in each case

cadquery/cq.py
```diff
@@ -4120,6 +4120,18 @@ def cylinder(
         If combine is false, the result will be a list of the cylinders produced.
         """
 
+        if isinstance(centered, bool):
+            centered = (centered, centered, centered)
+
+        xyz = [
```
lorenzncode, line 4126: Local variable 'xyz' is assigned to but never used.

lorenzncode:
Another idea is update the typing of the `direct` param
``` py       
    direct: Union[Tuple[float, float, float], Vector] = Vector(0, 0, 1),
```
or `= (0, 0, 1)`
See Plane normal: https://github.com/CadQuery/cadquery/blob/0c472ef741ef545174580e86190ad22294c45814/cadquery/occ_impl/geom.py#L553

cadquery/cq.py
kmarchais: Thank you!

adam-urbanczyk (approved):

adam-urbanczyk:
@lorenzncode it seems that your suggestions were applied. Are you OK with merging?

lorenzncode:
> @lorenzncode it seems that your suggestions were applied. Are you OK with merging?

I came across a couple of cases I'm not certain are correct.


1. Compare `box` and `cylinder` for the following case.  It seems the result is inconsistent.  Why is the cylinder in negative Y?  Shouldn't the bounding box of the cylinder be identical to box1?  That is, both the box and cylinder in +Y. 


```py
import cadquery as cq

centered = (False, False, False)

r = 1
h = 10

cyl1 = (
    cq.Workplane()
    .cylinder(h, r, (1, 0, 0), centered=centered)
)

box1 = (
    cq.Workplane()
    .box(h, r * 2, r * 2, centered=centered)
)
```
image



2. Here, I expected the second cylinder to be a YZ mirror of the first (bound box of cylinder to be in +Z).

```py
import cadquery as cq

r = 1
h = 10
centered = (False, False, False)

cyl1 = (
    cq.Workplane()
    .cylinder(h, r, (1, 0, 0), centered=centered)
)

cyl2 = (
    cq.Workplane()
    .cylinder(h, r, (-1, 0, 0), centered=centered)
)
```
image

lorenzncode:
I tested the snippet again [here](https://github.com/CadQuery/cadquery/pull/1593#issuecomment-2143554373) and get the same result.  I suppose this is fine as is.

If I were to use this method I think I would choose only one of these centering options:

```py
centered = (True, True, True)
# or 
centered = (True, True, False)
```

Anyway the default is `(True, True, True)`.

lorenzncode (approved):

adam-urbanczyk:
Alright, merging. Thanks @kmarchais !

PR #1606, opened 2024-06-18
Fix lineTo to do nothing instead of raising error when 0 length

endikallanomatxin:
Before the change it raised an error when startPoint and endPoint were equal

### DIFF
--- a/cadquery/cq.py
+++ b/cadquery/cq.py
@@ -1622,6 +1622,9 @@ class Workplane(object):
 
         endPoint = self.plane.toWorldCoords((x, y))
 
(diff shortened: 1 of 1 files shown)


adam-urbanczyk:
Why is this desired? 0 length line sounds like a user mistake that should result in an error.

endikallanomatxin:
When automating parametric design, it can be more comfortable not having to deal with the edge case of 0 length. At least in my case, it allowed for easier and more flexible parametric modelling.
(A line that in some configurations has a finite length, can easily be omitted in other configurations when it is not needed)

PR #1612, opened 2024-06-23
use pre-2.0 version of numpy until nptyping is updated to support it

justbuchanan:
See https://github.com/CadQuery/CQ-editor/issues/442 for more context.


adam-urbanczyk:
Thanks, fixed via #1617

PR #1617, opened 2024-06-28
Get rid of nptyping

### LINKED ISSUE #1616: had to downgrade numpy from 2.0.0 to 1.26.4 to make cadquery work
new install of cadquery on windows 10.  Cadquery version 2.4.0
using python 3.12.4
test of install  with "import cadquery" produced an error: AttributeError: module 'numpy' has no attribute 'bool8'.  Did you mean: 'bool'
google search returned that latest numpy no longer has bool8.
So I downgraded numpy from 2.0.0 to 1.26.4
It solve the problem.  basic cadquery test passed.

adam-urbanczyk:
Nptyping does not seem to be actively maintained and is not strictly needed.

Closes #1616 and https://github.com/CadQuery/CQ-editor/issues/442

### DIFF
--- a/cadquery/occ_impl/sketch_solver.py
+++ b/cadquery/occ_impl/sketch_solver.py
@@ -1,7 +1,7 @@
 from typing import Tuple, Union, Any, Callable, List, Optional, Iterable, Dict, Sequence
 from typing import Literal
-from nptyping import NDArray as Array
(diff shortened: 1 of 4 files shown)


adam-urbanczyk:
Simple but important fix - merging.

PR #1631, opened 2024-07-16
sketch offset docs update

eddieliberato:
Very simple and short sketch documentation update. Not related to an issue because it is not really an issue. 

3 small changes:
+ Added some (sub)headings to improve navigation.
+ Changed the order that two topics are presented, as they are related and I suppose it makes sense they are in sequence.
+ Added a short offset() example.

Thank you all for the time and effort on this project :thumbsup:


adam-urbanczyk:
LGTM @eddieliberato , thanks!

PR #1647, opened 2024-08-16
Change Python REPL prompt from "$" to ">>>"

kjcole:
In `installation.rst` the **Test Your Installation** section shows the **Bash** prompt followed by **Python** statements. This commit changes those to the correct REPL prompt.


jmwright (approved):
Thanks @kjcole

PR #1656, opened 2024-08-31
firstSolid() function to find the first solid between 2 points

moOsama76:
…pace points, with an optimized divide and conqer approach


moOsama76:
Can someone help? I don't know what is the failed automated test is about

PR #1657, opened 2024-08-31
firstSolid() function to find the first solid between 2 points

moOsama76:
function to find the first solid between start and end space points, with an optimized divide and conquer approach

### DIFF
--- a/cadquery/cq.py
+++ b/cadquery/cq.py
@@ -4583,6 +4583,49 @@ class Workplane(object):
 
         return self
 
+    def firstSolid(self : T, start:Vector, end:Vector, workplanes_to_consider) -> Vector:
+        """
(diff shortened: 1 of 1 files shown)


moOsama76:
Can someone help? I don't know what the AppVeyor test is about

adam-urbanczyk:
Let's start with why the PR? With the latest extensions (i.e. `.filter()`) to the API, you don't need to touch the codebase. If you want to share, you could also open a PR in the contrib repo.

moOsama76:
I checked filter() and I think the difference is that I return the exact point where there is a solid I don't return a workplane

adam-urbanczyk:
Then `invoke` or a free function...

PR #1690, opened 2024-10-25
fix: Deduplicate objects with same color on step export

### LINKED ISSUE #1689: Exporting STEP with colors doesn't de-duplicate as expected with different, but same-valued Colors
When `add`ing to an `Assembly`, it seems that the color/object tuples aren't correctly de-duplicating:

```
        # define the current color
        current_color = el.color if el.color else color

        # add a leaf with the actual part if needed
        if el.obj:
            # get/register unique parts referenced in the assy
            key0 = (current_color, el.obj)  # (color, shape)  # <<------- a different object (not value) here
            key1 = el.obj  # shape

            if key0 in unique_objs:   # <---- fails this check
                lab = unique_objs[key0]
            else:
                lab = tool.NewShape()
                if key1 in compounds:
```

If you add items to the assembly using a single `Color`, it's works like you expect:

```
def single_color_object():
    # Create an assembly
    assembly = cq.Assembly()
    color = cq.Color(0, 0, 1) # <---- same!

    # Add 4 cubes in a 2x2 grid
    for i in range(2):
        for j in range(2):
            assembly.add(
                cube,
                loc=cq.Location(
                    cq.Vector(i * 2, j * 2, 0),
                ),
                # Same object each time
                color=color,
            )

    return assembly
```
This returns 4 cubes all the same object (exports to ~25kB)

If you make a new color each time with the same value, you get duplicated cubes:
```
def multi_same_color_object():
    # Create an assembly
    assembly = cq.Assembly()

    # Add 4 cubes in a 2x2 grid
    for i in range(2):
        for j in range(2):
            assembly.add(
                cube,
                loc=cq.Location(
                    cq.Vector(i * 2, j * 2, 0),
                ),
                # This is always the same value, but it's a different object
                color=cq.Color(0, 0, 1),
            )

    return assembly
```
This returns 4 same coloured cubes, but they're all "new" objects (exports to ~75kB)

In fact, you can even subvert it the _other_ way by changing the value, not the object:
```
def multi_different_color_object():
    # Create an assembly
    assembly = cq.Assembly()
    color = cq.Color(0, 0, 1)

    # Add 4 cubes in a 2x2 grid
    for i in range(2):
        for j in range(2):
            # Change the color "secretly"
            color.r = i / 2
            color.g = j / 2
            color.b = 0
            assembly.add(
                cube,
                loc=cq.Location(
                    cq.Vector(i * 2, j * 2, 0),
                ),
                # Same object each time
                # but the value changed
                color=color,
            )

    return assembly
```
Saving this makes 4 _identically_ coloured cubes (exports to ~25kB)

It seems to me that this should be a by-value comparison rather than checking if it's literally the same object?

Script to replicate: 
[cq_export_color.zip](https://github.com/user-attachments/files/17503082/cq_export_color.zip)

Version:
```
cadquery                      2.4.0
cadquery-ocp                  7.7.2
```

lorenzncode:
Resolves #1689.

### DIFF
--- a/tests/test_assembly.py
+++ b/tests/test_assembly.py
@@ -398,7 +398,6 @@ def get_doc_nodes(doc, leaf=False):
 
         name_att = TDataStd_Name()
         label.FindAttribute(TDataStd_Name.GetID_s(), name_att)
-        # name = TCollection_ExtendedString(name_att.Get()).ToExtString()
 
         color = style.GetColorSurfRGBA()
         shape = expl.FindShapeFromPathId_s(doc, node.Id)
--- a/tests/test_assembly.py
+++ b/tests/test_assembly.py
@@ -399,7 +399,7 @@ def get_doc_nodes(doc, leaf=False):
 
         name_att = TDataStd_Name()
         label.FindAttribute(TDataStd_Name.GetID_s(), name_att)
-        name = TCollection_ExtendedString(name_att.Get()).ToExtString()
+        # name = TCollection_ExtendedString(name_att.Get()).ToExtString()
 
         color = style.GetColorSurfRGBA()
         shape = expl.FindShapeFromPathId_s(doc, node.Id)
@@ -412,7 +412,6 @@ def get_doc_nodes(doc, leaf=False):
         faces = []
         if not node.IsAssembly:
             it = TDF_ChildIterator(label)
-            i = 0
             while it.More():
                 child = it.Value()
                 child_shape = tool.GetShape_s(child)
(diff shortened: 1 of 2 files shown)


adam-urbanczyk:
Thanks for picking this up @lorenzncode , but shouldn't the solution be defining `__eq__` and `__hash__` for `cq.Color`?

lorenzncode:
Yes, that makes sense - thanks for the input @adam-urbanczyk !  I had a few hours on a train to look at this but please feel free to open an alternate PR or edit if you like.  I'll give your suggestion a try probably on the way back (cq actually works quite well on a chromebook by the way even cq-editor; x86_64 8GB RAM).

tests/test_assembly.py
```diff
@@ -399,7 +398,7 @@ def get_doc_nodes(doc, leaf=False):
 
         name_att = TDataStd_Name()
         label.FindAttribute(TDataStd_Name.GetID_s(), name_att)
-        name = TCollection_ExtendedString(name_att.Get()).ToExtString()
```
adam-urbanczyk, line 402: What is happening here?

adam-urbanczyk (approved):

tests/test_assembly.py
lorenzncode: This was an unused variable.  I'll clean it up and remove the line instead of comment out.

PR #1694, opened 2024-10-29
Support for type annotated parameters

openvmp:
Make it possible to annotate parameters with type information.
It is very important for parameters that need to default to None.


adam-urbanczyk:
Will you fix the coverage?

openvmp:
Most likely. Got distracted.

PR #1705, opened 2024-11-17
Add function to sketch arc given end point and radius

spike77453:
This adds a function to sketch an arc given only the end point and the arc's radius.

The start point of the arc is assumed to be `self._endPoint()` of the sketch. A third point on the arc (the arc's midpoint) is constructed by creating a vector rectangular to the arc's cord with magnitude equal to the [sagitta](https://en.wikipedia.org/wiki/Sagitta_(geometry)) of the arc, added to the midpoint of the arc's cord (see https://math.stackexchange.com/a/3482681).

`Edge.makeThreePointArc` is used to create the arc with these three point.

### DIFF
--- a/cadquery/sketch.py
+++ b/cadquery/sketch.py
@@ -15,7 +15,7 @@ from typing import (
     overload,
 )
 
-from math import tan, sin, cos, pi, radians, remainder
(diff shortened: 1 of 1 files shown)


adam-urbanczyk:
@spike77453 are you willing to add some tests?

PR #1718, opened 2024-11-30
Updated version specification for nlopt to try to resolve issues

jmwright:
A new version of nlopt (2.9.0) was released on on PyPI on November 16th. I have updated setup.py for this version and newer to try to resolve the issues that some users are having.

https://pypi.org/project/nlopt/

I did not try to pin the version for conda because it does not seem to be needed. Let me know if someone thinks I need to do that as well.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -29,7 +29,7 @@ if not is_rtd and not is_appveyor and not is_azure and not is_conda:
         "cadquery-ocp>=7.7.0a0,<7.8",
         "ezdxf",
         "multimethod>=1.11,<2.0",
-        "nlopt",
+        "nlopt>=2.9.0,<3.0",
         "typish",
         "casadi",
         "path",


adam-urbanczyk (approved):
Thanks, LGTM!

lorenzncode (approved):
Thanks @jmwright !

kmarchais:
Hello @jmwright, the nlopt 2.9.0 package is not available for macos x86_64.
It switched from x86_64 to arm between 2.7.1 and 2.8.0.

What kind of issues do this PR solve?
Do you think it is possible to loosen this constraint at least for this specific platform and architecture?

jmwright:
@kmarchais 2.7.1 was not working with numpy >= 2. If I relax the version constraint for MacOS x86_64 things will start breaking with CadQuery master, the upcoming 2.5 release, and beyond. I wonder why nlopt dropped Intel Mac support. There are still a lot in circulation, and it's still a standard runner in most CI systems. Is using conda an option for you?

kmarchais:
Thank you for the info, I understand. We will use it with this constraint then. Using conda is an option, yes. Ideally, I would have preferred to keep our library compatible on Intel Mac with pip.

PR #1732, opened 2024-12-22
Header cleanup

lorenzncode:


lorenzncode:
@dcowden confirmed it's ok to change the file headers to something like this:
```
# Copyright (c) CadQuery Development Team.
# Distributed under the terms of the Apache 2 License
```

Given this, I proposed to eliminate the header info outside setup.py, LICENSE files (currently the outdated header does not appear in every file - only a few select files).  There was agreement on this too.  This PR retains a header in the main cq.py.


I see @adam-urbanczyk  has fixed the mypy issue in another open PR.

adam-urbanczyk:
LGTM, thanks @lorenzncode

PR #1348, opened 2023-06-14
Add TypeVar to deprecate_kwarg to preserve type information

voneiden:
Closes #1308

Committed on behalf of the author @thelittlebug

Issue #1308, opened 2023-04-09
deprecate_kwarg_name is not type aware


thelittlebug:
thats the reason why you get no autocompletion for simple things like the following
```
Workplane().text("You rock", 12, 5,). #at this point the editor does not know how to autocomplete
```

but there is a simple solution to this issue: 
```
TCallable = TypeVar("TCallable", bound=Callable)

class deprecate_kwarg_name:
    def __init__(self, name, new_name):
        self.name = name
        self.new_name = new_name

    def __call__(self, f: TCallable) -> TCallable:
        @wraps(f)
        def wrapped(*args, **kwargs):

            if self.name in kwargs:
                warn(
                    f"Kwarg <{self.name}> will be removed. Please use <{self.new_name}>",
                    FutureWarning,
                )

            return f(*args, **kwargs)

        return cast(TCallable, wrapped)
```

adam-urbanczyk:
Could you open a PR?

thelittlebug:
yes, i will definitely do this. sry, i hadn't the possibilities  for that while i was posting this. 1-2 days and i'm in my used to work area ;)

voneiden:
@thelittlebug I've opened PR #1348 on your behalf, you OK with that?

PR #1733, opened 2024-12-24
Add TypeVar to deprecate_kwarg_name to preserve type information

nint8835:
Fixes issue initially reported in #1308, based off changes in #1348.

Initially reported issue reported issues with functions typed with `deprecate_kwarg_name`, but the fix PR fixed `deprecate_kwarg` which appeared to suffer from the same issue. I'm encountering the initially reported issue with calls to `.text` still, so this PR applies the fix from the initial issue description to the `deprecate_kwarg_name` class as well.

I've tested locally and confirmed this resolves typing issues on `.text` for me.

### DIFF
--- a/cadquery/utils.py
+++ b/cadquery/utils.py
@@ -58,7 +58,7 @@ class deprecate_kwarg_name:
         self.name = name
         self.new_name = new_name
 
-    def __call__(self, f):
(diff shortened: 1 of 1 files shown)


adam-urbanczyk:
Simple change, merging. Thanks @nint8835 !

Issue #1742, opened 2025-01-03
Dependencies not installed with 2.5


kmarchais:
Hello, since release 2.5, the dependencies are not installed automatically like it was before.

But if I install myself the dependencies of cadquery, it works fine:
```python
reqs = [
    "cadquery-ocp",
    "ezdxf",
    "multimethod",
    "nlop",
    "typish",
    "casadi",
    "path",
]
```

jmwright:
@kmarchais Are you using an Anaconda environment?

kmarchais:
No, I use `pip install cadquery`.

adam-urbanczyk:
But are you by any chance calling it within a conda env?

kmarchais:
No I am not in a conda env but after checking I had a env var CONDA_PREFIX set.
After removing it, the issue persists though

jmwright:
Actually, I am able to reproduce this. When I do `pip install -e .` locally without conda initted, all the dependencies install. When I follow the same process with `pip install cadquery`, the dependencies are not installed.

```
(cq-dev) jwright@lmde6:~/repos/cadquery$ pip install cadquery
Collecting cadquery
  Downloading cadquery-2.5.1-py3-none-any.whl (165 kB)
     ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 165.6/165.6 kB 2.4 MB/s eta 0:00:00
Installing collected packages: cadquery
Successfully installed cadquery-2.5.1
```

Is all I see when installing from PyPI. When I check the environment, there is no mention of conda.

adam-urbanczyk:
Then I did something wrong when building the package.

jmwright:
The dependenices did not make it into the METADATA file when the package was uploaded to PyPI.

![Screenshot from 2025-01-03 06-49-53](https://github.com/user-attachments/assets/2d478074-73e6-4367-ab0e-aecb97de588c)

I believe I had this happen once on a previous release. I believe the issue happened during the build phase (i.e. the conda environment variable was present in the environment during the build).

adam-urbanczyk:
yes, see my comment above :)

jmwright:
Item 5 under [Publish to PyPI](https://github.com/CadQuery/cadquery/wiki/Developer-Doc:-Creating-a-New-Release#publish-to-pypi-reference) in the wiki I think describes this problem.

jmwright:
[...]

PR #1743, opened 2025-01-03
Move to release 2.5.2

jmwright:
@adam-urbanczyk If you already have this in progress I can close this PR, I just figured I would save you some effort. I've got some time to do the build and upload too if you want. Totally up to you.

### DIFF
--- a/cadquery/__init__.py
+++ b/cadquery/__init__.py
@@ -4,7 +4,7 @@ try:
     __version__ = version("cadquery")
 except PackageNotFoundError:
     # package is not installed
-    __version__ = "2.5.1"
+    __version__ = "2.5.2"
 
 # these items point to the OCC implementation
 from .occ_impl.geom import Plane, BoundBox, Vector, Matrix, Location
--- a/changes.md
+++ b/changes.md
@@ -1,6 +1,11 @@
 Changes
 =======
 
+2.5.2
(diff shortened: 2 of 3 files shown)


jmwright:
Please also note that I removed the `a0` from the `cadquery-ocp` specification in setup.py because we have beta versions of cadquery-ocp available now, and I think it is best if users do not get them by default. Removing the `a0` should disable automatic installs of pre-release wheels.

adam-urbanczyk:
Is 2.5.2 needed? The packaging issue from  #1742 is not related to code in the repo. I'd rather upload correct 2.5.1

jmwright:
I don't think PyPI will let us overwrite the packages of a release.

jmwright:
@adam-urbanczyk We can try it, but you may get an error when we try to push the release using twine.

adam-urbanczyk:
I think you are right, pypi does not have a concept of build number.

adam-urbanczyk:
@jmwright would you be able to release and upload?

jmwright:
@adam-urbanczyk Do you want to me to push this to PyPI?

jmwright:
> would you be able to release and upload?

We posted at the same time again so I deleted my comment. Sure, I'll create the release and push to PyPI.

Issue #1704, opened 2024-11-17
ImportError: cannot import name 'zoom' from 'ezdxf' (/opt/conda/lib/python3.12/site-packages/ezdxf/__init__.py)


pmbaumgartner:
On a fresh install (on macOS, using docker) I get:

```
ImportError: cannot import name 'zoom' from 'ezdxf' (/opt/conda/lib/python3.12/site-packages/ezdxf/__init__.py)
```

full traceback:

```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/opt/conda/lib/python3.12/site-packages/cadquery/__init__.py", line 22, in <module>
    from .occ_impl import exporters
  File "/opt/conda/lib/python3.12/site-packages/cadquery/occ_impl/exporters/__init__.py", line 18, in <module>
    from .dxf import exportDXF, DxfDocument
  File "/opt/conda/lib/python3.12/site-packages/cadquery/occ_impl/exporters/dxf.py", line 6, in <module>
    from ezdxf import units, zoom
ImportError: cannot import name 'zoom' from 'ezdxf' (/opt/conda/lib/python3.12/site-packages/ezdxf/__init__.py)
```

kariem2k:
This is happening to me too on Windows (Clean Conda installation). I tried 2.4 and master both cause the same error.

jmwright:
Can you run `conda list` in your environment and post it here?

kariem2k:
Of course
[condalist.txt](https://github.com/user-attachments/files/18276275/condalist.txt)

Attached the file.

jmwright:
The version of ezdxf that output is showing is not even close to the current release. The oldest version I can find in my Linux environments is 1.1.0. I don't see any reason why conda would be installing an earlier version (a Windows package is available).

https://anaconda.org/conda-forge/ezdxf

Can you try to force the installation of 1.3.0 in your conda environment?

jmwright:
@pmbaumgartner @kariem2k Are either of you able to try the potential solution above?

jmwright:
@adam-urbanczyk Should we pin the version of ezdxf in environment.yml to force conda to use a current version? 

https://github.com/CadQuery/cadquery/blob/master/environment.yml#L16

jmwright:
@shimwell I have not seen evidence that pip is affected, have you? It seems like a conda solver issue. We could go ahead and add the pin to setup.py for completeness anyway.

jmwright:
[...]

PR #1746, opened 2025-01-08
Added ezdxf version pins

jmwright:
At least some of the variants of conda seem to be solving to much older versions of ezdxf (0.3.0 instead of 1.3.0), so this PR pins the version in environment.yml to try to counteract this, and adds the same pin to setup.py for consistency.

This should address issue #1704

### DIFF
--- a/conda/meta.yaml
+++ b/conda/meta.yaml
@@ -19,7 +19,7 @@ requirements:
     - ocp 7.7.2
     - vtk=*=qt*
     - pyparsing >=2.1.9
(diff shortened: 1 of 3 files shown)


jmwright:
Pretty minor, but it should solve an installation problem for Windows conda users.

adam-urbanczyk:
Thanks `meta.yaml` needs to be updated too.

adam-urbanczyk (approved):

PR #1759, opened 2025-02-02
Implement screenshots in show

adam-urbanczyk:
I had to implement something quickly for my slides, I think it'll be quite useful for others.

- [x] png screenshot
- [x] tune camera position, zoom, bg, wxh, ...
- [x] tender off-screen when non-interactive
- [x] test

img

Note that offscreen rendering does not work in CI with the exception of mac - that's why the test is dummy.

### DIFF
--- a/cadquery/vis.py
+++ b/cadquery/vis.py
@@ -162,6 +162,8 @@ def show(
     trihedron: bool = True,
     bgcolor: tuple[float, float, float] = (1, 1, 1),
     gradient: bool = True,
+    xpos: Union[int, float] = 0,
+    ypos: Union[int, float] = 0,
     **kwrags: Any,
 ):
     """
@@ -266,7 +268,12 @@ def show(
         int(w * width) if isinstance(width, float) else width,
         int(h * height) if isinstance(height, float) else height,
     )  # is height, width specified as float assume it is relative
-    win.SetPosition(-10, 0)
+
+    # set position
+    win.SetPosition(
+        int(w * xpos) if isinstance(xpos, float) else xpos,
+        int(h * ypos) if isinstance(ypos, float) else ypos,
+    )
 
     # show and return
     win.Render()
--- a/cadquery/vis.py
+++ b/cadquery/vis.py
@@ -183,6 +183,11 @@ def show(
 
     # VTK window boilerplate
     win = vtkRenderWindow()
+
+    # Render off-screen when not interacting
+    if not interact:
+        win.SetOffScreenRendering(1)
+
     win.SetWindowName(title)
     win.AddRenderer(renderer)
 
--- a/cadquery/vis.py
+++ b/cadquery/vis.py
@@ -1,7 +1,7 @@
 from . import Shape, Workplane, Assembly, Sketch, Compound, Color, Vector, Location
 from .occ_impl.assembly import _loc2vtk, toVTK
 
-from typing import Union, Any, List, Tuple, Iterable, cast
+from typing import Union, Any, List, Tuple, Iterable, cast, Optional
 
 from typish import instance_of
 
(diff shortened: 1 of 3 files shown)


jmwright:
Is it normal to get a different view in the PNG than I see in the window?

### Window

Screenshot from 2025-02-02 13-35-10

### Output PNG
screenshot

adam-urbanczyk:
I don't think so. I don't see such issues.

jmwright:
I don't know if it makes a difference, but I'm on Debian 12 x84_64.

adam-urbanczyk:
Would you be able to provide a MRE? If there is nothing obvious, I'd still propose to merge this and add a note in the docs.

adam-urbanczyk:
On win I use
```python
show(box(1,1,1), screenshot='img.png', interact=False, width=400, height=400, gradient=False, trihedron=False)
```
and that results in:
img

this is what I get in the window
image

jmwright:
I wonder if there is a timing aspect to it. I cannot reproduce the issue that I had yesterday.

adam-urbanczyk:
Interesting, did you use `interact=False` or `True`?

jmwright:
I believe I tried it both ways, but I know I started with `interact=True`. I ran python from the terminal and entered the commands manually that way yesterday. I have tried it both that way and via a script this morning and cannot reproduce.

adam-urbanczyk:
OK, unless you are able to reproduce it, I propose to park it and will add a note to the docs.

jmwright (approved):
Thanks @adam-urbanczyk

cadquery/vis.py
```diff
@@ -150,10 +152,20 @@ def show(
     edges: bool = False,
     specular: bool = True,
     title: str = "CQ viewer",
+    screenshot: Optional[str] = None,
+    interact: bool = True,
+    zoom: float = 1.0,
+    roll: float = -35,
+    elevation: float = -45,
+    width: Union[int, float] = 0.5,
+    height: Union[int, float] = 0.5,
```
lorenzncode, line 161: I would like to control window position size.  Say I'm working with two displays, with code editor on the first left display and then open the `show` window on the second display.  Maybe the feature can be added here or a separate PR?

```py
    x: Union[int, float] = -10,
    y: Union[int, float] = 0,

    #...
    win.SetPosition(
        int(w * x) if isinstance(x, float) else x,
        int(h * y) if isinstance(y, float) else y
    )
```

```py
# show window in upper left of second (right) display
show(b, width=0.25, height=0.5, x=0.5)

# show window in bottom left of second (right) display
show(b, width=0.25, height=0.5, x=0.5, y=0.5)
```

doc/vis.rst
```diff
@@ -77,6 +77,23 @@ Additionally it is possible to integrate with other libraries using VTK and disp
...
+camera position and windows size.
+
+.. code-block:: python
+
+    from cadquery.vis import show
+    from cadquery.func import box
+
+    b = box(1,1,1)
+
+    show(, width=800, height=800, screenshot='img.png', zoom=2, roll=-20, elevation=-30, interact=False)
```
lorenzncode, line 93: typo - missing positional arg

cadquery/vis.py
adam-urbanczyk: Added xpos, ypos. On win it does not seem to work with 2 displays though.

doc/vis.rst
adam-urbanczyk: Thanks, fixed!

cadquery/vis.py
lorenzncode: Seems to be working here!

Screenshot From 2025-02-07 21-07-12 v01

lorenzncode (approved):
Thanks @adam-urbanczyk , screenshots also working for me.

PR #1766, opened 2025-02-13
Deemphasize Discord

jmwright:
We have been pointing new users to the Discord server, but it looks like they are often not getting help once they get there. New users will have much more luck getting help with the other resources that are listed, and I added GitHub Discussions as a primary way to get help (in place of the Discord server). I have removed all of the prominent mentions of the Discord server, and added a line to let users know that they can go there, but it may not be the best place for them to get help.

The Discord server has turned into a broader discussion area for CodeCAD, often for more advanced topics. I will probably rework the server somewhat to reflect that.


adam-urbanczyk (approved):
LGTM

PR #1770, opened 2025-02-17
Fix copyWorkplane

bfleischhacker:
copyWorkplane currently doesnt actually return a full copy of the tagged workplane.


jmwright:
This looks fine to merge. Are we ready?

PR #1775, opened 2025-02-19
Pinning libiconv to fix gdk-pixbuf link issue on Windows

jmwright:


jmwright:
The libiconv teams seems to be actively working on a fix. Their CI was actively running when I just checked the issue.

https://github.com/conda-forge/libiconv-feedstock/pull/49

Until that has been tested and is definitely working again, this should fix the issue of WIndows CI being broken as long as I put the version pin in all the correct files.

adam-urbanczyk:
Thanks for locating the issue. I'd say let's give them 1-2 days.

jmwright:
The issue was fixed by the upstream project marking their package as broken, so I'm closing this.

PR #1776, opened 2025-02-20
Assembly.remove implementation

jmwright:
Allows removal of a part or subassembly from a parent assembly.


jmwright:
@adam-urbanczyk @lorenzncode This should be ready for review now.

lorenzncode:
Thanks @jmwright .  The only interesting finding I have is regarding performance removing large number of parts.  Here is a plot showing cumulative time in calls to add/remove for same number of parts.  Can performance be improved (only an issue for large number of removed parts)?  

add_remove_profile

jmwright:
@lorenzncode Are you able to profile the code using your test to see where the time is spent? I could see `del` being an inefficient operation, but I don't know for sure.

adam-urbanczyk:
It should be O(1) and it is mentioned in the official [docs](https://docs.python.org/3/library/stdtypes.html#mapping-types-dict).

jmwright:
Ok, no promises, but I will take a look at profiling and optimizing tomorrow morning.

lorenzncode:
Here is a profile test to add/remove N parts.
```py
import cadquery as cq
from cadquery.func import *
import numpy as np
import cProfile
import pstats
import io

pr = cProfile.Profile()
N = 1000000
sph = sphere(1)
xs = 1
assy = cq.Assembly(name="assy")
xvals = np.linspace(0, xs * N, N, endpoint=False)

for i, x in enumerate(xvals):
    assy.add(
        sph,
        name=f"p_{i}",
        # disable Location (slow) to focus on remove
        # still even with Location applied, the remove wall clock time
        # is ~2X slower than add for N=1e6 (best case sort order)
        # loc=cq.Location((x, 0, 0)),
    )

# best case
to_remove = [name for name, child in assy.traverse() if child.obj]
# worst
# to_remove = reversed([name for name, child in assy.traverse() if child.obj])

pr.enable()
for name in to_remove:
    assy.remove(name)
pr.disable()

s = io.StringIO()
ps = pstats.Stats(pr, stream=s).sort_stats(pstats.SortKey.CUMULATIVE).print_stats()
print(s.getvalue())
```
Time is spent on the list remove method (line 252).

lorenzncode:
Thanks for checking.  I don't have a need for such a large assembly - only wanted to find the limit.

cadquery/assembly.py
```diff
@@ -232,6 +232,38 @@ def add(self, arg, **kwargs):
 
         return self
 
+    def remove(self, name: str) -> "Assembly":
```
adam-urbanczyk, line 235: You might want to remove affected constraints too or simply add to the docstring that this method has limitations.

Another thing to remember is that deeply nested assys with constraints might break.

jmwright: I'm probably confused, but I am seeing some non-intuitive behavior when trying to remove constraints that are set between a part in the top-level assembly and parts in a sub-assembly. Consider the following contrived example.
```python
# Create another assembly with a nested subassembly that has a constraint to the top-level part
assy = cq.Assembly()
assy.add(box(1, 1, 1), name="part1")
subassy = cq.Assembly()
subassy.add(box(1, 1, 1), name="part2")
subassy.add(box(2, 2, 2), name="part3", loc=cq.Location(2.0, 2.0, 2.0))
assy.add(subassy, name="subassy")

for constraint in assy.constraints:
    print(constraint.objects)

assy.constrain("part1@faces@>Z", "subassy/part3@faces@<Z", "Axis")
assy.constrain("part1@faces@>Z", "subassy/part2@faces@<Z", "Plane")
assy.solve()

for constraint in assy.constraints:
    print(constraint.objects)
```
I see the following constraints:
```
('part1', 'subassy')
('part1', 'subassy')
```
How can we know which part in the sub-assembly the constraint is in relation to?

jmwright: Ok, I took constraint removal back out and added a note to the docstring.

adam-urbanczyk (approved):
Thanks!

lorenzncode (approved):

PR #1778, opened 2025-02-25
Per item styling in show + camera position

adam-urbanczyk:
- [x] Styling
- [x] Camera position
- [x] Tests
- [x] Docs

```python
from cadquery.vis import *
from cadquery.func import *

show(style(torus(10, 2), color='crimson', edges=True, tubes=True, linewidth=5, tolerance=0.1), style(box(3,3,3), color='green', markersize=0.1, alpha=0.5))
```
image


adam-urbanczyk:
@jon-proximafusion maybe tricky, but the basic setup goes as follows:
```
D = 100
h = 2

show(torus(10,2), position=(D, -D, D/h), roll=-45, zoom=10)
```

`D` is the "scale" parameter, play with `h` and `zoom` to get the desired view.

jmwright:
@adam-urbanczyk Was the torus screenshot above captured with those settings, because I get the following instead when I run that code.

Screenshot from 2025-02-28 09-29-46

adam-urbanczyk:
Yes, I used the settings above. Apparently the precise working of this is system dependent. NB you are quite close, and the example was meant as a starting point.

jmwright:
Ok, thanks. I tested the screenshot functionality with your first example and it worked well. I do not see any issues other than the few small suggestions I made in the code.

lorenzncode:
> I get the following instead when I run that code

The default window width and height is currently half the value returned by `win.GetScreenSize()`. You can remove this system dependency by providing a fixed width and height. It looks like the original aspect ratio is ~8:5. 

```py
show(torus(10, 2), position=(D, -D, D / h), roll=-45, zoom=10, width=1280, height=800)
```

cadquery/occ_impl/assembly.py
```diff
@@ -286,6 +287,71 @@ def _loc2vtk(
...
+        rv.AddPart(actor)
+
+        mapper = vtkMapper()
+        mapper.AddInputDataObject(data_edges)
+
+        actor = vtkActor()
+        actor.SetMapper(mapper)
+        actor.SetPosition(*trans)
+        actor.SetOrientation(*rot)
+        actor.GetProperty().SetColor(0, 0, 0)
```
lorenzncode, line 347: Add option to apply style to edges?

I experimented with```        actor.GetProperty().SetOpacity(0)```
to achieve similar behavior of the CAD Assistant "Smoothly Shaded" option.

image

adam-urbanczyk: I added something for this use case.

```python
show(style(box(1,1,1), edges=False))
```

image

adam-urbanczyk: Edges and mesh can be controlled separately now:
image

lorenzncode (approved):
Excellent feature - thanks @adam-urbanczyk !

PR #1779, opened 2025-02-26
Add STEP Import for Assemblies

jmwright:
The goal is to make it possible to round-trip assemblies to and from STEP without loss of data. This data can include:

* [x] Part location
* [x] Part color
* [ ] Sub-assembly location
* [ ] Sub-assembly color
* [ ] Face labels
* [ ] Face colors

### DIFF
--- a/cadquery/assembly.py
+++ b/cadquery/assembly.py
@@ -34,6 +34,7 @@ from .occ_impl.exporters.assembly import (
     exportGLTF,
     STEPExportModeLiterals,
 )
+from .occ_impl.importers.assembly import importStep as importStepTopLevel
 
 from .selectors import _expression_grammar as _selector_grammar
 from .utils import deprecate
@@ -563,6 +564,19 @@ class Assembly(object):
             raise ValueError(f"Unknown format: {exportType}")
 
         return self
+
+    @staticmethod
+    def importStep(path: str) -> "Assembly":
+        """
+        Reads an assembly from a STEP file.
+
+        :param path: Path and filename for writing.
+        :return: An Assembly object.
+        """
+
+        assy = importStepTopLevel(path)
+
+        return assy
 
     @classmethod
     def load(cls, path: str) -> "Assembly":
--- a/cadquery/occ_impl/importers/assembly.py
+++ b/cadquery/occ_impl/importers/assembly.py
@@ -1,0 +1,119 @@
+from OCP.TCollection import TCollection_ExtendedString
+from OCP.Quantity import Quantity_Color, Quantity_ColorRGBA
+from OCP.TDocStd import TDocStd_Document
+from OCP.IFSelect import IFSelect_RetDone
+from OCP.STEPCAFControl import STEPCAFControl_Reader
+from OCP.XCAFDoc import XCAFDoc_DocumentTool, XCAFDoc_ColorGen, XCAFDoc_ColorSurf
+from OCP.TDF import TDF_Label, TDF_LabelSequence
+
+import cadquery as cq
+from ..assembly import AssemblyProtocol
+
+
+def importStep(path: str) -> AssemblyProtocol:
+    """
+    Import a step file into an assembly.
+    """
+
+    # The assembly that is being built from the step file
+    assy = cq.Assembly()
+
+    # Document that the step file will be read into
+    doc = TDocStd_Document(TCollection_ExtendedString("XmlOcaf"))
+
+    # Create and configure a STEP reader
+    step_reader = STEPCAFControl_Reader()
+    step_reader.SetColorMode(True)
+    step_reader.SetNameMode(True)
+    step_reader.SetLayerMode(True)
+    step_reader.SetSHUOMode(True)
+
+    # Read the STEP file
+    status = step_reader.ReadFile(path)
+    if status != IFSelect_RetDone:
+        raise ValueError(f"Error reading STEP file: {path}")
+
+    # Transfer the contents of the STEP file to the document
+    step_reader.Transfer(doc)
+
+    # Shape and color tools for extracting XCAF data
+    shape_tool = XCAFDoc_DocumentTool.ShapeTool_s(doc.Main())
+    color_tool = XCAFDoc_DocumentTool.ColorTool_s(doc.Main())
+
+    def process_label(label, parent_location=None):
+        """
+        Recursive function that allows us to process the hierarchy of the assembly as represented
+        in the step file.
+        """
+
+        # Handle reference labels
+        if shape_tool.IsReference_s(label):
+            ref_label = TDF_Label()
+            shape_tool.GetReferredShape_s(label, ref_label)
+            process_label(ref_label, parent_location)
+            return
+
+        # Process components
+        comp_labels = TDF_LabelSequence()
+        shape_tool.GetComponents_s(label, comp_labels)
+        for i in range(comp_labels.Length()):
+            sub_label = comp_labels.Value(i + 1)
+
+            # The component level holds the location for its shapes
+            loc = shape_tool.GetLocation_s(sub_label)
+            if loc:
+                location = cq.Location(loc)
+
+                # Make sure that the location object is actually doing something interesting
+                # This is done because the location may have to go through multiple levels of
+                # components before the shapes are found. This allows the top-level component
+                # to specify the location/rotation of the shapes.
+                if location.toTuple()[0] == (0, 0, 0) and location.toTuple()[1] == (
+                    0,
+                    0,
+                    0,
+                ):
+                    location = parent_location
+            else:
+                location = parent_location
+
+            process_label(sub_label, location)
+
+        # Check to see if we have an endpoint shape
+        if shape_tool.IsSimpleShape_s(label):
+            shape = shape_tool.GetShape_s(label)
+
+            # Process the color for the shape, which could be of different types
+            color = Quantity_Color()
+            if color_tool.GetColor_s(label, XCAFDoc_ColorSurf, color):
+                r = color.Red()
+                g = color.Green()
+                b = color.Blue()
+                cq_color = cq.Color(r, g, b)
+            elif color_tool.GetColor_s(label, XCAFDoc_ColorGen, color):
+                r = color.Red()
+                g = color.Green()
+                b = color.Blue()
+                cq_color = cq.Color(r, g, b)
+            else:
+                cq_color = cq.Color(0.5, 0.5, 0.5)
+
+            # Handle the location if it was passed down form a parent component
+            if parent_location is not None:
+                assy.add(cq.Shape.cast(shape), color=cq_color, loc=parent_location)
+            else:
+                assy.add(cq.Shape.cast(shape), color=cq_color)
+
+    # Grab the labels, which should hold the assembly parent
+    labels = TDF_LabelSequence()
+    shape_tool.GetFreeShapes(labels)
+
+    # Make sure that we are working with an assembly
+    if shape_tool.IsAssembly_s(labels.Value(1)):
+        # Start the recursive processing of the assembly
+        process_label(labels.Value(1))
+
+    else:
+        raise ValueError("Step file does not contain an assembly")
+
+    return assy
--- a/tests/test_assembly.py
+++ b/tests/test_assembly.py
@@ -616,6 +616,43 @@ def test_step_export(nested_assy, tmp_path_factory):
     assert pytest.approx(c2.toTuple()) == (0, 4, 0)
 
 
+def test_step_import(tmp_path_factory):
+    """
+    Exports a simple assembly to step with locations and colors and ensures that information is
+    preserved when the assembly is re-imported from the step file.
+    """
+
+    # Use a temporary directory
+    tmpdir = tmp_path_factory.mktemp("out")
+    metadata_path = os.path.join(tmpdir, "metadata.step")
+
+    # Assembly with all the metadata added that needs to be read by the STEP importer
+    cube_1 = cq.Workplane().box(10, 10, 10)
+    cube_2 = cq.Workplane().box(15, 15, 15)
+    assy = cq.Assembly()
+    assy.add(cube_1, name="cube_1", color=cq.Color(1.0, 0.0, 0.0))
+    assy.add(
+        cube_2,
+        name="cube_2",
+        color=cq.Color(0.0, 1.0, 0.0),
+        loc=cq.Location((15, 15, 15)),
+    )
+    exportAssembly(assy, metadata_path)
+
+    assy = cq.Assembly.importStep(path=metadata_path)
+
+    # Make sure that there are the correct number of parts in the assembly
+    assert len(assy.children) == 2
+
+    # Make sure that the colors are correct
+    assert assy.children[0].color == cq.Color(1.0, 0.0, 0.0)
+    assert assy.children[1].color == cq.Color(0.0, 1.0, 0.0)
+
+    # Make sure that the locations are correct
+    assert assy.children[0].loc.toTuple()[0] == (0, 0, 0)
+    assert assy.children[1].loc.toTuple()[0] == (15, 15, 15)
+
+
 @pytest.mark.parametrize(
     "assy_fixture, expected",
     [


cadquery/assembly.py
```diff
@@ -564,6 +565,19 @@ def export(
...
+    @staticmethod
+    def importStep(path: str) -> "Assembly":
+        """
+        Reads an assembly from a STEP file.
+
+        :param path: Path and filename for writing.
+        :return: An Assembly object.
+        """
+
+        assy = cast(Assembly, importStepTopLevel(path))
```
adam-urbanczyk, line 577: Could you change the interface of  `importStepTopLevel` to accept an AssemblyProtocol, e.g. 
```
assy = Assembly()
importTopLevel(assy, path)

return assy
```

This way:
- it will match the intended design
- you'll get rid of a circular dependency
- cast will not be needed anymore

`cq.Assembly.importStep` should also be a class method to allow sub-classing.

jmwright: Thanks for the feedback, I refactored. Let me know if I refactored incorrectly.

adam-urbanczyk: I changed to a classmethod and started using `Self`

jmwright:
@adam-urbanczyk In 4045c5813708dbbd497f2504c5ece94927b8a689 you set the `cadquery.Assembly.importStep` method up to be a class method, and you construct `assy` before calling the lower-level method. However, since the `Assembly.name` property is private and can only be set during instantiation, this creates an issue for me. I need to be able to set the top-level assembly name based on the name set in the STEP file, which requires me to set the `name` property after instantiation.

We need to decide on the proper way to fix this.

adam-urbanczyk:
Hey, I do not understand the issue. You can always do this:

```
assy = Assembly()
assy.name = '123'
```
Do you mean that you need to modify `AssemblyProtocol` to satisify mypy? That also does not sound like an issue to me.

jmwright:
Correct, mypy complains. I can make that change to make `name` a public property if you don't see an issue with doing so.

jmwright:
@adam-urbanczyk @lorenzncode I think this is ready for a look when you get a chance.

lorenzncode:
I tested this branch with the "door" example from the docs.  I made only a color modification to set vslot color to white.

The step file as displayed in FreeCAD:

image


After `importStep` displayed with `cadquery.vis`:

out

It looks like locations are not preserved in this case?

jmwright:
Thanks @lorenzncode I thought I had that fixed. I'll figure out what's going on.

adam-urbanczyk:
I can confirm the issue. It seems like locations are ignored. Additionally, import/export roundtrip seems to distort colors.

jmwright:
Thanks. The colors are probably being shifted because I did not merge master into this branch yet to catch the recent color changes. The location will take some more digging. I have a check or two in the tests to make sure that the location is preserved during import, and those pass. I'll use the the door as an example to make sure it gets properly fixed.

jmwright:
@adam-urbanczyk Actually, if I just run the door assembly code from the docs and display it with `cadquery.vis.show`, some colors are shifted/wrong. The STEP import/export feature was NOT used for this screenshot.

Screenshot from 2025-06-10 10-26-06

jmwright:
@adam-urbanczyk The gold color of the v-slot parts above is because there is no color defined for those components, and `cadquery.vis.show` is defaulting to a different color than CQ-editor and the 3D viewer in the documentation. So I think it's a non-issue unless we want to standardize on a default color across the visualization platforms.

@lorenzncode @adam-urbanczyk Can you provide the steps to reproduce the location issue with the constrainted door assembly? I cannot reproduce, so we are following different processes somehow. The code below has all the components in the correct location when I show it using `cadquery.vis.show`.

```python
import cadquery as cq
from cadquery.occ_impl.exporters.assembly import exportStepMeta
from cadquery.vis import show

# Parameters
H = 400
W = 200
D = 350

PROFILE = cq.importers.importDXF("vslot-2020_1.dxf").wires()

SLOT_D = 6
PANEL_T = 3

HANDLE_D = 20
HANDLE_L = 50
HANDLE_W = 4


def make_vslot(l):
    return PROFILE.toPending().extrude(l)


def make_connector():
    rv = (
        cq.Workplane()
        .box(20, 20, 20)
        .faces("<X")
        .workplane()
        .cboreHole(6, 15, 18)
        .faces("<Z")
        .workplane(centerOption="CenterOfMass")
        .cboreHole(6, 15, 18)
    )

    # tag mating faces
    rv.faces(">X").tag("X").end()
    rv.faces(">Z").tag("Z").end()

    return rv


def make_panel(w, h, t, cutout):
    rv = (
        cq.Workplane("XZ")
        .rect(w, h)
        .extrude(t)
        .faces(">Y")
        .vertices()
        .rect(2 * cutout, 2 * cutout)
        .cutThruAll()
        .faces("<Y")
        .workplane()
        .pushPoints([(-w / 3, HANDLE_L / 2), (-w / 3, -HANDLE_L / 2)])
        .hole(3)
    )

    # tag mating edges
    rv.faces(">Y").edges("%CIRCLE").edges(">Z").tag("hole1")
    rv.faces(">Y").edges("%CIRCLE").edges("<Z").tag("hole2")

    return rv


def make_handle(w, h, r):
    pts = ((0, 0), (w, 0), (w, h), (0, h))

    path = cq.Workplane().polyline(pts)

    rv = (
        cq.Workplane("YZ")
        .rect(r, r)
        .sweep(path, transition="round")
        .tag("solid")
        .faces("<X")
        .workplane()
        .faces("<X", tag="solid")
        .hole(r / 1.5)
    )

    # tag mating faces
    rv.faces("<X").faces(">Y").tag("mate1")
    rv.faces("<X").faces("<Y").tag("mate2")

    return rv


# define the elements
door = (
    cq.Assembly()
    .add(make_vslot(H), name="left")
    .add(make_vslot(H), name="right")
    .add(make_vslot(W), name="top")
    .add(make_vslot(W), name="bottom")
    .add(make_connector(), name="con_tl", color=cq.Color("black"))
    .add(make_connector(), name="con_tr", color=cq.Color("black"))
    .add(make_connector(), name="con_bl", color=cq.Color("black"))
    .add(make_connector(), name="con_br", color=cq.Color("black"))
    .add(
        make_panel(W + 2 * SLOT_D, H + 2 * SLOT_D, PANEL_T, SLOT_D),
        name="panel",
        color=cq.Color(0, 0, 1, 0.2),
    )
    .add(
        make_handle(HANDLE_D, HANDLE_L, HANDLE_W),
        name="handle",
        color=cq.Color("yellow"),
    )
)

# define the constraints
(
    door
    # left profile
    .constrain("left@faces@<Z", "con_bl?Z", "Plane")
    .constrain("left@faces@<X", "con_bl?X", "Axis")
    .constrain("left@faces@>Z", "con_tl?Z", "Plane")
    .constrain("left@faces@<X", "con_tl?X", "Axis")
    # top
    .constrain("top@faces@<Z", "con_tl?X", "Plane")
    .constrain("top@faces@<Y", "con_tl@faces@>Y", "Axis")
    # bottom
    .constrain("bottom@faces@<Y", "con_bl@faces@>Y", "Axis")
    .constrain("bottom@faces@>Z", "con_bl?X", "Plane")
    # right connectors
    .constrain("top@faces@>Z", "con_tr@faces@>X", "Plane")
    .constrain("bottom@faces@<Z", "con_br@faces@>X", "Plane")
    .constrain("left@faces@>Z", "con_tr?Z", "Axis")
    .constrain("left@faces@<Z", "con_br?Z", "Axis")
    # right profile
    .constrain("right@faces@>Z", "con_tr@faces@>Z", "Plane")
    .constrain("right@faces@<X", "left@faces@<X", "Axis")
    # panel
    .constrain("left@faces@>X[-4]", "panel@faces@<X", "Plane")
    .constrain("left@faces@>Z", "panel@faces@>Z", "Axis")
    # handle
    .constrain("panel?hole1", "handle?mate1", "Plane")
    .constrain("panel?hole2", "handle?mate2", "Point")
)

# solve
door.solve()

success = exportStepMeta(door, "door.step")

if success:
    imported_assy = cq.Assembly.importStep("door.step")
    show(imported_assy)
```

lorenzncode:
> Can you provide the steps to reproduce the location issue

Try `export` instead of `exportStepMeta`.
```py
door.export("door.step")
```

STEP is exported in the assembly tests.   I'm getting a `ValueError` on importStep with one of them.

run the pytest, export step (still uses old save method)
```
pytest tests/test_assembly.py --basetemp=./tmp -k test_colors_assy1
```

This fails with ValueError:
```py
from cadquery import Assembly

obj = Assembly.importStep("./tmp/assembly10/chassis0_assy.step")
```

adam-urbanczyk:
Here is mine:

before:
image
after:
image

I also found another issues, without copies the import throws. See last section of the snippet.

```python
#%% imports
from cadquery import Assembly, Location, Color
from cadquery.func import box, rect
from cadquery.vis import show

#%% prepare the model
def make_model(name: str, COPY: bool):
    b = box(1,1,1)
    
    assy = Assembly(name='test_assy')
    assy.add(box(1,2,5), color=Color('green'))
    
    
    
    for v in rect(10,10).vertices():
        assy.add(
            b.copy() if COPY else b,
            loc=Location(v.Center()),
            color=Color('red')
        )
        
    # show(assy)
    
    assy.export(name)
    
    return assy
    
assy_copy = make_model("test_assy_copy.step", True)
assy = make_model("test_assy.step", False)

show(assy)

#%% import the assy with copies

assy_i = Assembly.importStep("test_assy_copy.step")
show(assy_i)

#%% import the assy without copies - this throws

assy_i = Assembly.importStep("test_assy.step")
show(assy_i)
```

jmwright:
@adam-urbanczyk @lorenzncode  The parent location was not being applied correctly when the original assembly.export method is used. I've fixed that now and pulled in the srgb changes from master, so I think the location and color issues should be fixed.

adam-urbanczyk:
Do you plan to fix the second issue I found?

jmwright:
@adam-urbanczyk Yes, I'll see what I can figure out.

jmwright:
@adam-urbanczyk The `Assembly.export` method is actually duplicating the names throughout the STEP file, so it does not look like it is a problem with this importer. I find 14 different entries with that UUID in the STEP file. I can add a check to protect against duplicate names being read from the STEP file which alters the duplicate names, but I am not sure we should mask the root cause (duplicated names within the STEP file). It seems like the existing behavior of letting`Assembly.add` throw the error could be the right thing to do here.

Screenshot from 2025-06-13 12-08-28

In the screenshot: Two variations of the name, one with the `_part` postfix and one without. Those are the only two variations of the name throughout the STEP file.

adam-urbanczyk:
This tool (eDrawings) can import the problematic STEP without copies and the structure LGTM. I think that CQ should be able to handle this as well (and produce identical structure).

image

jmwright:
How do we do that without breaking things like constraints that use part names? We would have to remove the `Assembly.add` restriction on duplicate names, correct?

adam-urbanczyk:
But subassy names *are* unique and parts don't have names in cq AFAIR.

adam-urbanczyk:
Additional observation:
```python
In [3]: assy_i.objects
Out[3]:
{'a9a08701-4ac0-11f0-a634-c86e08a4b6ef': <cadquery.assembly.Assembly at 0x1b1cf1dcbc0>,
 'a7644761-4ac0-11f0-bcb0-c86e08a4b6ef_part': <cadquery.assembly.Assembly at 0x1b1d123ab10>,
 'a76594bb-4ac0-11f0-9013-c86e08a4b6ef_part': <cadquery.assembly.Assembly at 0x1b1d1238440>,
 'a7668644-4ac0-11f0-91c5-c86e08a4b6ef_part': <cadquery.assembly.Assembly at 0x1b1d123ac30>,
 'a7668645-4ac0-11f0-9a6c-c86e08a4b6ef_part': <cadquery.assembly.Assembly at 0x1b1d123ad20>,
 'a7668646-4ac0-11f0-abef-c86e08a4b6ef_part': <cadquery.assembly.Assembly at 0x1b1d123b680>}

In [4]: assy.objects
Out[4]:
{'test_assy': <cadquery.assembly.Assembly at 0x1b1d11fcc20>,
 'a77ae412-4ac0-11f0-804d-c86e08a4b6ef': <cadquery.assembly.Assembly at 0x1b1d123a090>,
 'a77ae413-4ac0-11f0-ab48-c86e08a4b6ef': <cadquery.assembly.Assembly at 0x1b1d123a480>,
 'a77ae414-4ac0-11f0-b508-c86e08a4b6ef': <cadquery.assembly.Assembly at 0x1b1d123a510>,
 'a77ae415-4ac0-11f0-b410-c86e08a4b6ef': <cadquery.assembly.Assembly at 0x1b1d123a5d0>,
 'a77ae416-4ac0-11f0-b4a6-c86e08a4b6ef': <cadquery.assembly.Assembly at 0x1b1d123a690>}
```

It seems that the name is taken from the wrong level of the tree.

jmwright:
It turns out that the `exportStepMeta()` method has the same issue of applying names to the wrong level. It tries to flatten the assembly by default, and the name of a child-assembly gets applied to the shape(s) within the child-assembly and not to the assembly itself.

Screenshot from 2025-06-17 14-23-45

I need to create another PR to fix nesting and name handling in `exportStepMeta()`, and will probably make an option to flatten the assembly (flattening has been asked for by users), but it won't be the default.

I think there are still issues lurking with having duplicate names, but I will fix the name handling in this PR and make sure the nesting works properly. Then I can worry about the unique naming issues.

adam-urbanczyk:
I see that you are working on #1858, but wouldn't it be better to fix the issue here iso adding (before the release) more functionality? I'd propose to add a heuristic that detect nodes with one child and `_part` suffix and take the shape from `_part` and name from the parent. IMO this does not  sound too complicated.

jmwright:
#1858 was created because there is an oversight within #1782 that causes `exportStepMeta` to act differently than the standard `Assembly.export` method. I'm trying to fix that, not adding functionality, other than adding a `flatten` parameter because I expect it to be asked for eventually.

jmwright:
The structure looks the same now on the test case that used to fail. The first screenshot is the one that was exported, then the second is that STEP imported, then re-exported.

test_assy

test_assy_reexport

adam-urbanczyk:
OK, I'll take one final look.

cadquery/assembly.py
```diff
@@ -34,6 +34,7 @@
     exportGLTF,
     STEPExportModeLiterals,
 )
+from .occ_impl.importers.assembly import importStep as importStepTopLevel
```
adam-urbanczyk: Could you rename `importStepTopLevel` to `_importStep`? I was a little bit confused by the name.

cadquery/assembly.py
```diff
@@ -609,7 +610,21 @@ def export(
...
+    def importStep(cls, path: str) -> Self:
+        """
+        Reads an assembly from a STEP file.
+
+        :param path: Path and filename for writing.
+        :return: An Assembly object.
+        """
+
+        assy = cls()
+        importStepTopLevel(assy, path)
```
adam-urbanczyk: ofc here too

cadquery/occ_impl/importers/assembly.py
```diff
@@ -0,0 +1,198 @@
...
+    layer_tool = XCAFDoc_DocumentTool.LayerTool_s(doc.Main())
+
+    def _process_simple_shape(label, parent_location=None, parent_name=None):
+        shape = shape_tool.GetShape_s(label)
+
+        # Tracks the RGB color value and whether or not it was found
+        cq_color = None
+
+        # Load the name of the part in the assembly, if it is present
+        name = None
```
adam-urbanczyk: remove or change to `name: str`?

cadquery/occ_impl/importers/assembly.py
```diff
@@ -0,0 +1,198 @@
...
+    # Shape and color tools for extracting XCAF data
+    shape_tool = XCAFDoc_DocumentTool.ShapeTool_s(doc.Main())
+    color_tool = XCAFDoc_DocumentTool.ColorTool_s(doc.Main())
+    layer_tool = XCAFDoc_DocumentTool.LayerTool_s(doc.Main())
+
+    def _process_simple_shape(label, parent_location=None, parent_name=None):
+        shape = shape_tool.GetShape_s(label)
+
+        # Tracks the RGB color value and whether or not it was found
+        cq_color = None
```
adam-urbanczyk: It does not make sense to initialize here, it is done below. Typing wise this is confusing. What was the rationale?

cadquery/occ_impl/importers/assembly.py
```diff
@@ -0,0 +1,198 @@
...
+        if parent_name is not None and parent_name != assy.name:
+            name = parent_name
+        else:
+            name_attr = TDataStd_Name()
+            label.FindAttribute(TDataStd_Name.GetID_s(), name_attr)
+            name = str(name_attr.Get().ToExtString())
+
+        # Process the color for the shape, which could be of different types
+        color = Quantity_Color()
+        cq_color = cq.Color(0.50, 0.50, 0.50)
```
adam-urbanczyk: What are the magic parameters?

cadquery/occ_impl/importers/assembly.py
```diff
@@ -0,0 +1,198 @@
...
+        # Handle the location if it was passed down form a parent component
+        if parent_location is not None:
+            assy.add(
+                cq.Shape.cast(shape),
+                name=name,
+                color=cq_color,
+                loc=cq.Location(parent_location),
+            )
+        else:
+            assy.add(cq.Shape.cast(shape), name=name, color=cq_color)
```
adam-urbanczyk: Why isn't this covered. Is parent location not optional?

cadquery/occ_impl/importers/assembly.py
```diff
@@ -0,0 +1,198 @@
...
+
+        # Check all the attributes of all the children to find the subshapes and any names
+        for j in range(label.NbChildren()):
+            child_label = label.FindChild(j + 1)
+            attr_iterator = TDF_AttributeIterator(child_label)
+            while attr_iterator.More():
+                current_attr = attr_iterator.Value()
+
+                # Get the type name of the attribute so that we can decide how to handle it
+                if current_attr.DynamicType().Name() == "TNaming_NamedShape":
```
adam-urbanczyk: Could you add an explanation/comment regarding NamedShape and the other thing below (GraphNode).

cadquery/occ_impl/importers/assembly.py
```diff
@@ -0,0 +1,198 @@
...
+
+            for i in range(comp_labels.Length()):
+                sub_label = comp_labels.Value(i + 1)
+                # Get the location of the sub-label, if it exists
+                loc = shape_tool.GetLocation_s(sub_label)
+
+                # Pass down the location or other context as needed
+                # Add the parent location if it exists
+                if parent_location is not None:
+                    loc = parent_location * loc
```
adam-urbanczyk: Isn't this suspicious. Assy applies locations in a relative sense. It shouldn't be needed to construct absolute locations.

adam-urbanczyk (changes requested):
Thanks, 2nd pass

cadquery/occ_impl/importers/assembly.py
jmwright: They are the RGB values that most of the CQ visualization methods seem to be using as a default with assemblies (gray/silver). The values are not standardized as `cadquery.vis.show` seems to use a gold color IIRC. CQ-editor uses the gray color though.

cadquery/occ_impl/importers/assembly.py
jmwright: I'll have to check, but I think I added this because of nested sub-assemblies

jmwright: Confirmed. If the parent location is not passed through the call stack, nested subassemblies will break. I added the `test_nested_subassembly_step_import` test to illustrate this. If you disable parent location passing and run that test, it should fail.

lorenzncode:
I am still getting a ValueError with a STEP from the tests.

```
(cqdev) lorenzn@fedora:~/devel/cadquery$ git status -u no
On branch assembly-import
Your branch is up to date with 'upstream/assembly-import'.

nothing to commit, working tree clean
(cqdev) lorenzn@fedora:~/devel/cadquery$ pytest tests/test_assembly.py --basetemp=./tmp2 -k test_colors_assy1 -q --no-summary
...........                                                                                                                                                        [100%]
11 passed, 96 deselected, 13 warnings in 0.58s
sys:1: DeprecationWarning: builtin type swigvarlink has no __module__ attribute
(cqdev) lorenzn@fedora:~/devel/cadquery$ cat test2.py 
from cadquery import Assembly

obj = Assembly.importStep("./tmp2/assembly10/chassis0_assy.step")
(cqdev) lorenzn@fedora:~/devel/cadquery$ python test2.py 
  File "/home/lorenzn/devel/cadquery/test2.py", line 3, in <module>
    obj = Assembly.importStep("./tmp2/assembly10/chassis0_assy.step")
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/lorenzn/devel/cadquery/cadquery/assembly.py", line 622, in importStep
    _importStep(assy, path)
  File "/home/lorenzn/devel/cadquery/cadquery/occ_impl/importers/assembly.py", line 194, in importStep
    process_label(labels.Value(1))
  File "/home/lorenzn/devel/cadquery/cadquery/occ_impl/importers/assembly.py", line 173, in process_label
    process_label(sub_label, loc, name)
  File "/home/lorenzn/devel/cadquery/cadquery/occ_impl/importers/assembly.py", line 149, in process_label
    process_label(ref_label, parent_location, parent_name)
  File "/home/lorenzn/devel/cadquery/cadquery/occ_impl/importers/assembly.py", line 173, in process_label
    process_label(sub_label, loc, name)
  File "/home/lorenzn/devel/cadquery/cadquery/occ_impl/importers/assembly.py", line 149, in process_label
    process_label(ref_label, parent_location, parent_name)
  File "/home/lorenzn/devel/cadquery/cadquery/occ_impl/importers/assembly.py", line 173, in process_label
    process_label(sub_label, loc, name)
  File "/home/lorenzn/devel/cadquery/cadquery/occ_impl/importers/assembly.py", line 149, in process_label
    process_label(ref_label, parent_location, parent_name)
  File "/home/lorenzn/devel/cadquery/cadquery/occ_impl/importers/assembly.py", line 173, in process_label
    process_label(sub_label, loc, name)
  File "/home/lorenzn/devel/cadquery/cadquery/occ_impl/importers/assembly.py", line 149, in process_label
    process_label(ref_label, parent_location, parent_name)
  File "/home/lorenzn/devel/cadquery/cadquery/occ_impl/importers/assembly.py", line 179, in process_label
    _process_simple_shape(label, parent_location, parent_name)
  File "/home/lorenzn/devel/cadquery/cadquery/occ_impl/importers/assembly.py", line 74, in _process_simple_shape
    assy.add(
  File "/home/lorenzn/devel/cadquery/cadquery/assembly.py", line 241, in add
    self.add(assy)
  File "/home/lorenzn/devel/cadquery/cadquery/assembly.py", line 222, in add
    raise ValueError("Unique name is required")
ValueError: Unique name is required
```

cadquery/occ_impl/importers/assembly.py
adam-urbanczyk: OK, but that is actually suspicious. When you construct an assy you do not need combine (i.e. multiply) the locations explicitly.

cadquery/occ_impl/importers/assembly.py
adam-urbanczyk: OK, default color should be `None`. See:

```python
class Assembly(object):
    """Nested assembly of Workplane and Shape objects defining their relative positions."""

    loc: Location
    name: str
    color: Optional[Color]
    metadata: Dict[str, Any]

    obj: AssemblyObjects
    parent: Optional["Assembly"]
    children: List["Assembly"]

    objects: Dict[str, "Assembly"]
    constraints: List[Constraint]
 ```

**Comment by adam-urbanczyk** [comment]
@jmwright any updates on this PR?

**Comment by jmwright** [comment]
@adam-urbanczyk Not yet. I got busy with some other things.

**Comment by jmwright** [comment]
@adam-urbanczyk I'm getting mypy errors now that I did not before. Most of them are probably my fault, but there is this one about the OpenCASCADE code.
```
cadquery\occ_impl\importers\assembly.py:105: error: "TDF_Attribute" has no attribute "GetFather"  [attr-defined]
```
Does this have anything to do with the stubs being published as part of OCP 7.9.x?

**Comment by adam-urbanczyk** [comment]
Nothing should be published, but let me check

**Comment by jmwright** [comment]
I'm having trouble with mypy. It is giving me different errors locally than I get in CI even though I have the same version. I'll create a new environment from scratch tomorrow and try to match the CI environment to see if I can get rid of the discrepancies.

**Comment by jmwright** [comment]
@adam-urbanczyk @lorenzncode Please have another look when you get a chance. This PR has been significantly reworked, and a test or two added.

**Comment by adam-urbanczyk** [comment]
Something is off with top level location handling.

Before importing:
image

After importing:
image

Code:

```python
#%% imports
from cadquery import Assembly, Location, Color, importers
from cadquery.func import box, rect, sphere
from cadquery.vis import show

#%% prepare the model
def make_model(name: str, COPY: bool):
    b = box(1,1,1)
    
    assy_top = Assembly(name='test_assy_top', loc=Location((0,1,1)))
    assy_top.add(sphere(1))
    
    assy = Assembly(name='test_assy', loc=Location((5,5,1)))
    assy.add(box(1,2,5), color=Color('green'))
    
    
    
    for v in rect(10,10).vertices():
        assy.add(
            b.copy() if COPY else b,
            loc=Location(v.Center()),
            color=Color('red')
        )
        
    # show(assy)
    assy_top.add(assy)
    assy_top.export(name)
    
    return assy_top

assy = make_model("test_assy.step", False)

show(assy, Location(),  scale=5)

#%% import the assy without copies - this throws

assy_i = Assembly.importStep("test_assy.step")
wp = importers.importStep("test_assy.step")

show(assy_i, Location(),scale=5)
```

jmwright:
@adam-urbanczyk Thanks, I fixed that bug and modified one of the tests so that I could add an assert to verify.

tests/test_assembly.py
```diff
@@ -785,6 +820,369 @@ def test_meta_step_export_edge_cases(tmp_path_factory):
...
 
 
+def test_assembly_step_import(tmp_path_factory):
+    """
+    Test if the STEP import works correctly for an assembly with subshape data attached.
+    """
+    assy = subshape_assy()
+
+    # Use a temporary directory
+    tmpdir = tmp_path_factory.mktemp("out")
```
lorenzncode, line 882: FYI, you could use the `tmpdir` fixture (line 42) to reduce boilerplate.

tests/test_assembly.py
```diff
@@ -785,6 +820,369 @@ def test_meta_step_export_edge_cases(tmp_path_factory):
...
+
+def test_plain_assembly_import(tmp_path_factory):
+    """
+    Test to make sure that importing plain assemblies has not been broken.
+    """
+
+    tmpdir = tmp_path_factory.mktemp("out")
+    plain_step_path = os.path.join(tmpdir, "plain_assembly_step.step")
+
+    # Simple cubes
```
lorenzncode, line 1036: There are existing assembly fixtures that could be used here (and for other tests) instead.  OK if you would rather explicitly redefine here.

lorenzncode (commented):
Thanks for the big effort @jmwright!  The error I had before is resolved now when I try importing from the pytest generated step file.


I find import/export round trip changes the assembly as follows.  Is this expected?

```py
from cadquery import Assembly, Location
from cadquery.func import box


def round_trip_test(assy, n=3):
    for i in range(n):
        print(f"\n{i=}")
        for name in assy._flatten().keys():
            print(name)
        assy.export(f"out{i}.step")
        assy = Assembly.importStep(f"out{i}.step")


b = box(1, 1, 1)
assy = Assembly(name="assy")
assy.add(b, name="box0")
assy.add(b, name="box1", loc=Location((1, 0, 0)))

round_trip_test(assy, 5)
```
```
i=0
assy/box0
assy/box1
assy

i=1
assy/box0/box0_part
assy/box0
assy/box1/box0_part
assy/box1
assy

i=2
assy/box0/box0_part/box0_part_part
assy/box0/box0_part
assy/box0
assy/box1/box0_part/box0_part_part
assy/box1/box0_part
assy/box1
assy

i=3
assy/box0/box0_part/box0_part_part/box0_part_part_part
assy/box0/box0_part/box0_part_part
assy/box0/box0_part
assy/box0
assy/box1/box0_part/box0_part_part/box0_part_part_part
assy/box1/box0_part/box0_part_part
assy/box1/box0_part
assy/box1
assy

i=4
assy/box0/box0_part/box0_part_part/box0_part_part_part/box0_part_part_part_part
assy/box0/box0_part/box0_part_part/box0_part_part_part
assy/box0/box0_part/box0_part_part
assy/box0/box0_part
assy/box0
assy/box1/box0_part/box0_part_part/box0_part_part_part/box0_part_part_part_part
assy/box1/box0_part/box0_part_part/box0_part_part_part
assy/box1/box0_part/box0_part_part
assy/box1/box0_part
assy/box1
assy
```

adam-urbanczyk:
> Thanks for the big effort @jmwright! The error I had before is resolved now when I try importing from the pytest generated step file.
> 
> I find import/export round trip changes the assembly as follows. Is this expected?
> 


That is definitely not OK.

cadquery/occ_impl/importers/assembly.py
```diff
@@ -0,0 +1,216 @@
...
+                    # Extract the color, if present on the shape
+                    if color_tool.GetColor(final_shape, XCAFDoc_ColorSurf, color):
+                        rgb = color.GetRGB()
+                        cq_color = cq.Color(
+                            rgb.Red(), rgb.Green(), rgb.Blue(), color.Alpha()
+                        )
+                    else:
+                        cq_color = None
+
+                    new_assy.add(
```
adam-urbanczyk: So probably some logic should be added here to handle the `name` / `name_part` pairs. E.g. something like:
```
if ref_name.endswith('_part') and ref_name.startswith(parent_name):
    new_assy[parent_name].obj = cq_shape
else:
    ...
```
CC @lorenzncode

tests/test_assembly.py
```diff
@@ -785,6 +820,369 @@ def test_meta_step_export_edge_cases(tmp_path_factory):
...
+    assy = cq.Assembly.importStep(round_trip_step_path)
+
+    # Check some general aspects of the assembly structure now
+    assert len(assy.children) == 2
+    assert assy.name == "top-level"
+    assert assy.children[0].name == "cube_1"
+    assert assy.children[1].children[0].name == "cylinder_1"
+
+
 @pytest.mark.parametrize(
```
adam-urbanczyk, line 1221: Probably a test that asserts idempotence of export/import is also needed.

cadquery/occ_impl/importers/assembly.py
adam-urbanczyk: @lorenzncode @jmwright  I added this draft PR #1884, it needs some work still.

- [x] Added logic to handle `_part` shapes
- [x] Fixed `_copy()` to include metadata

NB: I think that at least one test is actually wrong. Subshapes should be a property of the (sub)assy that owns the shape. Not a top level one.

jmwright:
@adam-urbanczyk I've pulled in your changes from #1884 and have things working except for a little bit of code coverage. You had a `FIXME` comment in the code that I did not remove because I was not sure what you meant beyond fixing the mypy error for that code. Feel free to commit directly to this PR.

adam-urbanczyk:
OK @jmwright, on it.

gohu41:
Hello, not sure if my comment is relevant here but I noticed the following:
- the master branch installed from pip repository runs with python 3.9 but this branch installed from sources needs python 3.10 (because of Type| Type syntax)
- just pip installing from the sources is not working, some requirements are missing (like ocp)
- creating a new env with the environment.yaml file works

adam-urbanczyk:
@jmwright  I think I'm done. Here is what I did/changed:

- Subshapes names were not read -> fixed
- Added subshape validation (`isSubshape`).
- Changed the default export (`cq.Assembly.export`) to handle subshape data too
- Implemented `cq.Assembly.load`
- Added `__getattr__` and `__getitem__`
- Changed `addSubshape` to allow adding only to the current/next level. I think it would be good to in the end only allow adding subshape data to the current level but let's discuss after the release. It this change is contentious I'll rework.

adam-urbanczyk:
I found one more issue: import did not update the fist key in `.objects` properly. It is fixed and tested now.

jon-proximafusion:
Just tinkering with this and noticed the file sizes appear to be different to the original step file, am I doing something wrong here.

I have installed this branch with pip
```
pip install git+https://github.com/CadQuery/cadquery.git@assembly-import
```
```python
import cadquery as cq
sphere = cq.Workplane().sphere(5.0)

assembly = cq.Assembly()
assembly.add(sphere, name="sphere")
assembly.export("sphere.step")

assembly2 = cq.Assembly()
assembly2.importStep("sphere.step")
assembly2.save("sphere2.step")
assembly2.export("sphere3.step")
```

Filesize appears to be quite different, is some content being missed when importing and saving again 

```bash
ll
-rw-rw-r--  1 jon jon 3254 Sep 12 13:43 sphere2.step
-rw-rw-r--  1 jon jon 3254 Sep 12 13:43 sphere3.step
-rw-rw-r--  1 jon jon 6703 Sep 12 13:43 sphere.step
```

adam-urbanczyk:
Weird, something is definitely off here

adam-urbanczyk:
Note that `load` is a ~static~ class method and it creates a new assembly:

```
import cadquery as cq
sphere = cq.Workplane().sphere(5.0)

assembly = cq.Assembly()
assembly.add(sphere, name="sphere")
assembly.export("sphere.step")

assembly2 = Assembly.load("sphere.step")
assembly2.save("sphere2.step")
assembly2.export("sphere3.step")
```

```
!ls -l *sphere*.step
-rw-r--r-- 1  197609 6706 Sep 12 14:58 sphere.step
-rw-r--r-- 1  197609 6690 Sep 12 15:01 sphere2.step
-rw-r--r-- 1  197609 6690 Sep 12 15:01 sphere3.step
```

lorenzncode:
I captured timing for export/load and step file size.  These are only observations and perhaps are expected behavior. 

- load can be slow with many objects.
- There seems to be significant file size difference in the reexport. 

```py
# 10k boxes example from PR 1157

import cadquery as cq
import timeit
import os

box = cq.Workplane().box(1, 1, 1)

assy = cq.Assembly()
num = 100
for i in range(1, num + 1, 1):
    for j in range(1, num + 1, 1):
        assy.add(
            box,
            name=f"box{i}_{j}",
            color=cq.Color("red"),
            loc=cq.Location((1.1 * i, 1.1 * j, 0)),
        )

# export the assembly to step
tsave = timeit.timeit('assy.export("test1.step")', globals=globals(), number=1)

# load the step
result = []


def timed_load():
    result.append(assy.load("test1.step"))


tload = timeit.timeit("timed_load()", globals=globals(), number=1)

# reexport to check file size
result[0].export("test2.step")

print(f"time export: {tsave:.1f} s")
print(f"time load: {tload:.1f} s")

print(
    f"original test1.step size: {os.path.getsize('test1.step') / (1024 * 1024):.2f} MB"
)
print(
    f"reexport test2.step size: {os.path.getsize('test2.step') / (1024 * 1024):.2f} MB"
)
```

(I am on a faster machine which explains the <2s export vs previous result with ~5s export)

```
time export: 1.5 s
time load: 29.3 s
original test1.step size: 21.92 MB
reexport test2.step size: 38.22 MB
```

cadquery/occ_impl/shapes.py
```diff
@@ -6558,6 +6558,20 @@ def check(
     return rv
 
 
+def isSubshape(s1: Shape, s2: Shape) -> bool:
+    """
+    Check is s1 is a subshape of s2.
```
lorenzncode: Suggest:
`Check is s1 a subshape of s2`
or 
`Check if s1 is a subshape of s2`

cadquery/assembly.py
```diff
@@ -609,9 +615,26 @@ def export(
         return self
 
     @classmethod
-    def load(cls, path: str) -> "Assembly":
+    def importStep(cls, path: str) -> Self:
+        """
+        Reads an assembly from a STEP file.
+
+        :param path: Path and filename for writing.
```
lorenzncode: writing->reading
` :param path: Path and filename for reading.`

cadquery/occ_impl/importers/assembly.py
```diff
@@ -0,0 +1,216 @@
...
+    doc = TDocStd_Document(TCollection_ExtendedString("XmlOcaf"))
+
+    # Create and configure a STEP reader
+    step_reader = STEPCAFControl_Reader()
+    step_reader.SetColorMode(True)
+    step_reader.SetNameMode(True)
+    step_reader.SetLayerMode(True)
+    step_reader.SetSHUOMode(True)
+
+    Interface_Static.SetIVal_s("read.stepcaf.subshapes.name", 1)
```
lorenzncode: Perhaps in a follow up PR, provide an option to set to 0.

cadquery/occ_impl/importers/assembly.py
```diff
@@ -0,0 +1,216 @@
...
+                    if color_tool.GetColor(final_shape, XCAFDoc_ColorSurf, color):
+                        rgb = color.GetRGB()
+                        cq_color = Color(
+                            rgb.Red(), rgb.Green(), rgb.Blue(), color.Alpha()
+                        )
+                    else:
+                        cq_color = None
+
+                    # this if/else is needed to handle different structures of STEP files
+                    # "*"/"*_part" based naming is the default strucutre produced by CQ
```
lorenzncode: typo in comment: "structure"

tests/test_assembly.py
```diff
@@ -785,6 +829,376 @@ def test_meta_step_export_edge_cases(tmp_path_factory):
...
+    assert imported_assy.children[1].objects["box_2"].loc.toTuple()[0] == (
+        10.0,
+        10.0,
+        10.0,
+    )
+
+
+def test_assembly_step_import_roundtrip(tmp_path_factory):
+    """
+    Tests that the assembly does not mutate during successive export-import round trips.
```
lorenzncode, line 1172: Need another similar test.  Here is an example that will fail.  I think this needs to be fixed.

```py
seg = segment((0, 0, 0), (1, 0, 0))
assy_orig = Assembly(name="assy_orig")
assy_orig.add(seg, name="seg")
assy_orig.add(seg, name="seg2", loc=Location(0, 0, 1))
assy_orig.export("orig.step")

assy = Assembly.load("orig.step")

for k in assy_orig.objects:
    assert k in assy

# this currently fails!
for k in assy.objects:
    assert k in assy_orig
```

lorenzncode (changes requested):

tests/test_assembly.py
adam-urbanczyk: Yes, this needs to be fixed.

cadquery/occ_impl/importers/assembly.py
adam-urbanczyk: Yes, let's limit the scope of this PR.

adam-urbanczyk (approved):

adam-urbanczyk:
FYI: I did quite some rework. Will explain later.

adam-urbanczyk:
* Added xbf and xml support
* Got rid of extra nodes as much as possible
* Round trips require different handling of colors; colors should only be used on leaf nodes
* Fixed color handling for roundtrips (srgb vs linear rgb)
* There are still some strange issues here and there, but typical use cases seem to work reliably:
  * Copies and multiple nesting results in wrong instance names when exporting to STEP (see https://github.com/Open-Cascade-SAS/OCCT/discussions/726)
  * XML export seems to distort colors (XBF does not)

adam-urbanczyk:
Additionally, I tweaked the docs a little bit.

cadquery/occ_impl/importers/assembly.py
```diff
@@ -0,0 +1,364 @@
...
+
+                elif shape_tool.IsSimpleShape_s(ref_label):
+                    # Find the name of this referenced part
+                    ref_name = _get_name(comp_label)
+
+                    # A single shape needs to be added to the assembly
+                    final_shape = shape_tool.GetShape_s(ref_label)
+                    cq_shape = Shape.cast(final_shape)
+
+                    # If the instance has no color, try to find the refernced shape color
```
lorenzncode: typo _referenced_

cadquery/occ_impl/importers/assembly.py
```diff
@@ -0,0 +1,364 @@
...
+                    if color is None:
+                        color = _get_shape_color(final_shape, color_tool)
+
+                    # this if/else is needed to handle different structures of STEP files
+                    # "*"/"*_part" based naming is the default strucutre produced by CQ
```
lorenzncode: typo strucutre -> structure

cadquery/occ_impl/exporters/assembly.py
```diff
@@ -267,29 +273,39 @@ def _process_assembly(
     return status == IFSelect_ReturnStatus.IFSelect_RetDone
 
 
-def exportCAF(assy: AssemblyProtocol, path: str) -> bool:
+def exportCAF(assy: AssemblyProtocol, path: str, binary: bool = False) -> bool:
     """
-    Export an assembly to a OCAF xml file (internal OCCT format).
+    Export an assembly to ax XCAF xml or xbf file (internal OCCT formats).
```
lorenzncode: typo _ax_

lorenzncode (approved):
I did not find any issues (other than typos - see comments).  I like that step import simplifies the toCAF validation in the tests.  For the record here is my new result of the 10K boxes test.  It shows the reduction in filesize given the change to remove extra nodes.

```
time export: 1.5 s
time load: 19.9 s
original test1.step size: 5.81 MB
reexport test2.step size: 5.81 MB
```

adam-urbanczyk:
@jmwright are you OK with merging?

jmwright:
+1 to merge

adam-urbanczyk:
Thanks for bearing with me!

PR #1781, opened 2025-03-01
Add imprinting to free func

adam-urbanczyk:
- [x] `imprint`
- [x] `Edge.params`
- [x] `Edge.tangents`
- [x]  improve `cut` and `fuse`
    - [x] multi-argument `fuse`
- [x] `setThreads`
- [x] tests

Some infor on the glue option: https://dev.opencascade.org/content/gluing-options-boolean-component


adam-urbanczyk:
Some low level stuff, but quite useful. In another PR and after some field experience we might want to update the imprinting function on the assy level.

cadquery/occ_impl/shapes.py
```diff
@@ -1838,6 +1865,59 @@ def paramAt(self: Mixin1DProtocol, d: Union[Real, Vector]) -> float:
 
         return rv
 
+    def params(self: Mixin1DProtocol, pts: Iterable[Vector], tol=1e-6) -> List[float]:
+        """
+        Computes u values closest to given vectors.
+
+        :returns: list of u values.
```
jmwright, line 1872: Even though it will work fine like this, I think our convention has been to put `return` after the `param` entries. I also don't think it matters, but we have used `return` more times in the code base than `returns`.

cadquery/occ_impl/shapes.py
```diff
@@ -5432,46 +5541,142 @@ def _bool_op(
...
+
+def setThreads(n: int):
     """
-    Fuse two shapes.
+    Set number of threads to be used by boolean operations.
     """
 
-    builder = BRepAlgoAPI_Fuse()
-    _bool_op(s1, s2, builder, tol)
+    pool = OSD_ThreadPool.DefaultPool_s()
```
jmwright, line 5569: Could this have implications for GUIs and client-server systems like CQ-editor, jupyter-cadquery, etc? Multi-threading seems to get weird sometimes in those types of applications.

jmwright: I guess if it does cause an issue, the user can just not use this feature.

cadquery/occ_impl/shapes.py
```diff
@@ -5432,46 +5541,142 @@ def _bool_op(
...
     return _compound_or_shape(builder.Shape())
 
 
-def intersect(s1: Shape, s2: Shape, tol: float = 0.0) -> Shape:
+def intersect(
+    s1: Shape, s2: Shape, tol: float = 0.0, glue: GlueLiteral = None
+) -> Shape:
     """
-    Intersect two shapes.
+    Subtract two shapes.
```
jmwright, line 5620: Why was the docstring changed to be different than the method name (`intersect` vs `subtract`)?

cadquery/occ_impl/shapes.py
```diff
@@ -5432,46 +5541,142 @@ def _bool_op(
...
+    for s in shapes:
+        builder.AddArgument(s.wrapped)
+
+    builder.Perform()
+
+    # fill history if provided
+    if history is not None:
+        images = builder.Images()
+
+        # collect shapes presen in the history dict
```
jmwright, line 5672:
```suggestion
        # collect shapes present in the history dict
```

jmwright (approved):
Looks good to me, thanks @adam-urbanczyk !

tests/test_shapes.py
```diff
@@ -108,6 +108,35 @@ def test_face_positions():
...
+    pts_orig = e.sample(N)[0]
+    pts = [pt + Vector(0, 0, 1e-1) for pt in pts_orig]
+
+    ps = e.params(pts)
+
+    for i in range(N):
+        assert (e.positionAt(ps[i], mode="parameter") - pts_orig[i]).Length == approx(0)
+
+
+def test_edge_tangets():
```
lorenzncode, line 125: typo in test function name.  "test_edge_tangets()" -> test_edge_tangents()"

cadquery/occ_impl/shapes.py
lorenzncode, line 5620: typo "Intersect"

lorenzncode (approved):
I did some testing. No issues found other than docstring/typo.

cadquery/occ_impl/shapes.py
adam-urbanczyk: You indeed, don't need to use this future. I doubt that this has implications for GUIs, as OCCT (and CQ) already uses multi-threading and this simply allows to change the default threadpool size.

PR #1786, opened 2025-03-10
WIP: non-blocking show using trame/vtk

### LINKED ISSUE #1784: Interactive show
Currently `show` is blocking, it would be great to be able to use it multiple times without blocking of the script execution. I see 3 options

1. Use `Qt` or `Tk`, requires integration of the gui event loop with the interpreter (Check how IPython odes that)
2. Run all the VTK stuff in a separate thread and periodically invoke `vtkRenderWindowInteractor.ProcessEvents()`
3. Trame+pywebview

adam-urbanczyk:
Closes #1784 (?)

API:
```python
from cadquery.fig import Figure
from cadquery.func import sphere, torus

# this starts a server and opens a browser tab
fig = Figure()

# we can use show, the API is fluent
fig.show(torus(10, 2), color='violet').show(sphere(5))

# zoom to fit
fig.fit()

# clear
fig.clear()
```

image

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -37,7 +37,7 @@ build: false
 test_script:
     - mamba run -n cadquery black . --diff --check
     - mamba run -n cadquery mypy cadquery
-    - mamba run -n cadquery pytest -v --cov
+    - mamba run -n cadquery pytest -v --gui --cov
 
 on_success:
     - mamba run -n cadquery codecov
--- a/cadquery/fig.py
+++ b/cadquery/fig.py
@@ -10,19 +10,19 @@ from threading import Thread
 from itertools import chain
 from webbrowser import open_new_tab
 
-from trame.app import get_server, Server
+from trame.app import get_server
+from trame.app.core import Server
 from trame.widgets import html, vtk as vtk_widgets, client
 from trame.ui.html import DivLayout
 
 from . import Shape
-from .vis import style, Showable, ShapeLike, _split_showables, _to_vtk_pts, _to_vtk_axs
+from .vis import style, Showable, ShapeLike, _split_showables
 
 from vtkmodules.vtkRenderingCore import (
     vtkRenderer,
     vtkRenderWindow,
     vtkRenderWindowInteractor,
     vtkProp3D,
-    vtkActor,
 )
 
 
--- a/cadquery/fig.py
+++ b/cadquery/fig.py
@@ -1,6 +1,10 @@
-import asyncio
-
-from typing import Any, cast
+from asyncio import (
+    new_event_loop,
+    set_event_loop,
+    run_coroutine_threadsafe,
+    AbstractEventLoop,
+)
+from threading import Thread
 
 from trame.app import get_server, Server
 from trame.widgets import html, vtk as vtk_widgets, client
@@ -32,14 +36,15 @@ class Figure:
     win: vtkRenderWindow
     ren: vtkRenderer
     view: vtk_widgets.VtkRemoteView
-    shapes: dict[Shape, vtkProp3D]
+    shapes: dict[Shape, list[vtkProp3D]]
     actors: list[vtkProp3D]
-    loop: Any
+    loop: AbstractEventLoop
+    thread: Thread
 
     def __init__(self, port: int):
 
-        self.loop = asyncio.new_event_loop()
-        asyncio.set_event_loop(self.loop)
+        self.loop = new_event_loop()
+        set_event_loop(self.loop)
 
         # vtk boilerplate
         renderer = vtkRenderer()
@@ -76,7 +81,7 @@ class Figure:
         self.actors = []
 
         # server
-        server = get_server(123)
+        server = get_server("CQ")
         server.client_type = "vue3"
 
         # layout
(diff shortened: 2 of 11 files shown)


adam-urbanczyk:
@jmwright @lorenzncode I'm not saying this is ready, but would you be able to give this a try? Just to collect some datapoints on possible issues...

lorenzncode:
I tested on linux. 

- It worked with both firefox, chrome browsers.  The browser opened automatically.

- As the output message says you can add --server to prevent launching browser:

	$ python test1.py

	App running at:
	 - Local:   http://localhost:18081/
	 - Network: http://127.0.0.1:18081/

	Note that for multi-users you need to use and configure a launcher.
	And to prevent your browser from opening, add '--server' to your command line.

  	$ python test1.py --server 

- I found the server exits immediately (connection closed) unless I add code to prevent it (sleep). 
  I expected behavior more similar to starting a jupyter server where it runs in the foreground and you can stop the server from the terminal. 

- It does not work in a container without exposing DISPLAY
  I would like to try rendering with VTK.js in the browser in this scenario.  Yesterday I tested with a python server and browser using websockets.  I did not try to connect to CQ yet.  I was using dummy data with VTK.js vtkXMLPolyDataReader.  This isn't a requirement for this PR - I'm only wondering if I'll be able to reuse any of this if I continue with a custom VTK.js front end.

lorenzncode:
OK I forgot you probably intend this to be run from ipython.  In that case the server does stay running.

jmwright:
I created a new venv for the changes and installed via pip, and it seems to work. Two corrections for your sample code at top.
```python
from cadquery.fig import Figure()
```
should be...
```python
from cadquery.fig import Figure
```
and
```python
fit.clear()
```
should be...
```python
fig.clear()
```
The server continued to run when started by entering the code into the Python REPL and all updates showed up in the 3D viewer. The mouse rotation of the scene is a little laggy, even though my laptop does not normally struggle with 3D rendering.

Debian 12, AMD64, local install via pip.

adam-urbanczyk:
Thanks for the feedback.

* NB: this is a non-blocking show, not a server. It is intended to be used in a repl.
* For vtk.js you can try `vtk_widgets.VtkLocalView` but it did not work well for me. I'd like to wait for WASM/VTK 9.4
* Regarding containers/DISPLAY I thought that this is not needed and you just need to expose a port and manually connect from the host
* Indeed the perf with remote view could be better, but it is IMHO very usable

lorenzncode:
> Regarding containers/DISPLAY I thought that this is not needed and you just need to expose a port and manually connect from the host

Yes, that's correct.  It works fine with cadquery.vis and cq-editor (and I expect this PR after adding a browser).  I only meant that if this was a server it could be an alternative display method which might have some advantages.  Sorry to get off topic.  Here is how I use the container today: 

```
 weston --xwayland -S wayland-88 --width=1400 --height=1000
```
 Then in container run script:
 ```bash
 podman run -it --entrypoint=/bin/bash  --userns=keep-id \
	--security-opt label=type:container_runtime_t \
	-e XDG_RUNTIME_DIR=/tmp \
	-e DISPLAY=${display} \
	-e "WAYLAND_DISPLAY=$way_display" \
	-v /run/user/$user_uid/${way_display}:/tmp/${way_display} \
	-v "$XDG_RUNTIME_DIR/${way_display}:/tmp/${way_display}" \
	-v /tmp/.X11-unix/${x_socket}:/tmp/.X11-unix/${x_socket} \
	...
 ```

adam-urbanczyk:
For now only tested on windows CI due to GUI availability. My understanding is that after moving to wasm based rendering, this won't be an issue anymore (i.e. it will be possible to test on every runner).

lorenzncode:
I'm looking for a way to preserve a custom zoom and viewer position on `clear` all.   Could there be an option for that?

Say I'm iterating in ipython repeating runs of a script.
```
run test1.py
run test1.py
```

After the first run I modified the viewer position and zoom by interacting with the browser.

On the second run I'd like to hide/remove a shape temporarily without disturbing the zoom and camera.

Script on first run:

```py
from cadquery.func import box
from cadquery.fig import Figure

fig = Figure()

xval = 1.1
s1 = box(1, 1, 1)
s2 = box(1, 1, 1).move(x=xval)
s3 = box(1, 1, 1).move(x=2 * xval)

fig.show(s1, color="red")
fig.show(s2, color="green")
fig.show(s3, color="blue")
```

Script on second run:

```py
from cadquery.func import box
from cadquery.fig import Figure

fig = Figure()

xval = 1.1
s1 = box(1, 1, 1)
s2 = box(1, 1, 1).move(x=xval)
s3 = box(1, 1, 1).move(x=2 * xval)

fig.clear()
fig.show(s1, color="red")
fig.show(s2, color="green")
#fig.show(s3, color="blue")
```

s3 is removed.  The zoom and camera are also reset.

I can workaround it with following in REPL after first run.:
```py
fig.clear(s3)
```

Then for second run: do not call `clear`.

adam-urbanczyk:
I changed the default to not reset. You can always call `fit()` manually.

lorenzncode (approved):
Thanks @adam-urbanczyk ! This is working for me.

Future enhancements might include:

- add camera options
- maybe a widget to control object visibily similar to CQ-editor

jmwright:
Sorry, this PR still had "WIP" in the title and I did not realize that it was only waiting on my review. I have used this some and have not found any issues. Feel free to merge.

Will there be documentation at some point?

adam-urbanczyk:
For now this is meant for dogfooding so no docs yet.

jmwright:
@adam-urbanczyk Ok, thanks!

PR #1815, opened 2025-04-15
Support for materials

adrianschneider94:
Implemented material support.
In this approach, materials are stored in python dataclasses, the OpenCascade objects are only created when building the assembly.
Tests/docs are missing at the moment.


adam-urbanczyk:
@shimwell it would certainly be nice to have some predefined materials, and maybe allow to use the material name for other things too. But let's start with making CI green.

adrianschneider94:
Material support in step exporting will get better in OCCT:
https://github.com/Open-Cascade-SAS/OCCT/pull/447

adrianschneider94:
VRML export runs, but is not correct yet.

adrianschneider94:
The PR would be ready from my side for a first iteration.

I think it would be great to add textures later on, one could create quite good visualizations directly with cadquery/vtk. For the time, the PBR materials already achieve quite nice results.

adam-urbanczyk:
Thanks for all the hard work! Are the new classes picklable?

adrianschneider94:
Yes, they are pure dataclasses/basic python types, so everything can be pickled.

How far is the progress on the serialization topic? Have you thought about using pydantic? I plan to implement model caching in our internal project and json would be easier to use in a db like with Postgres' JSONB column or redis' JSON datatype compared to pickle.

adam-urbanczyk:
Perfect, thanks!

There are not plans to implement serialization beyond pickling which is already there. For caching I'd suggest joblib, but maybe you have something else in mind.

jmwright:
@adrianschneider94 @adam-urbanczyk What action items are left to do before getting this PR merged?

adam-urbanczyk:
Review... Given the size of the PR it might take a while, but we'll definitely get this merged.

jmwright:
I'll plan to spend some time reviewing it next week then.

cadquery/materials.py
```diff
@@ -0,0 +1,299 @@
...
+    from OCP.XCAFDoc import XCAFDoc_Material, XCAFDoc_VisMaterial
+    from vtkmodules.vtkRenderingCore import vtkActor
+
+
+RGB: TypeAlias = Tuple[float, float, float]
+RGBA: TypeAlias = Tuple[float, float, float, float]
+
+
+@dataclass(frozen=True)
+class Color:
```
adam-urbanczyk, line 21: I'd rather have the the original Color class not removed.

cadquery/materials.py
```diff
@@ -0,0 +1,299 @@
...
+
+    def __post_init__(self):
+        """Validate the material properties."""
+        # Validate ranges
+        if not 0.0 <= self.shininess <= 1.0:
+            raise ValueError("Shininess must be between 0.0 and 1.0")
+        if not 0.0 <= self.transparency <= 1.0:
+            raise ValueError("Transparency must be between 0.0 and 1.0")
+
+    def apply_to_vtk_actor(self, actor: "vtkActor") -> None:
```
adam-urbanczyk: Please use the current convention of CQ (i.e. camel case methods). 

After the in question  PEP: "In particular: do not break backwards compatibility just to comply with this PEP!"

cadquery/occ_impl/assembly.py
```diff
@@ -301,57 +284,79 @@ def _loc2vtk(
...
+    edgecolor: Color = DEFAULT_EDGE_COLOR,
     edges: bool = True,
     linewidth: float = 2,
     tolerance: float = 1e-3,
     angularTolerance: float = 0.1,
 ) -> vtkAssembly:
+    """
+    Convert an assembly to a VTK assembly with both solid faces and edges.
+    
+    Args:
```
adam-urbanczyk: Please use the convention of CQ for docstrings (see other docstrings).

cadquery/occ_impl/assembly.py
```diff
@@ -4,127 +4,68 @@
...
-
-    def __hash__(self):
-
-        return hash(self.toTuple())
-
-    def __eq__(self, other):
-
-        return self.toTuple() == other.toTuple()
+@dataclass
+class AssemblyElement:
```
adam-urbanczyk: Please revert to the original style of iteration

tests/test_assembly.py
```diff
@@ -537,29 +537,6 @@ def solve_result_check(solve_result: dict) -> bool:
     return all(checks)
 
 
-def test_color():
```
adam-urbanczyk, line 540: If you need to remove a test, something is really wrong with the PR. I think you made to many backward incompatible changes.

tests/test_assembly.py
```diff
@@ -1106,7 +1083,7 @@ def check_nodes(doc, expected):
                 (["box0", "box0_part"], {"color_shape": (1.0, 0.0, 0.0, 1.0)}),
                 (
                     ["box1", "box1_part"],
-                    {"color_shape": cq.Color().toTuple()},
```
adam-urbanczyk: Please do not change the API without a clear need.

adam-urbanczyk (changes requested):
Hey, thanks for all the hard work, but this requires some serious rework with backward compatibility in mind. May I as why there are so many changes that are not needed?

adrianschneider94:
> I see that this branch has conflicts now that I must have missed last week. Should it be rebased?

Rebase is done.

cadquery/materials.py
adrianschneider94: The new Color class has the same behaviour as the old one (except toTuple, I will bring that back as alias for `rgba()`). I think it makes sense to unify color handling throughout the codebase instead of mixing `Color`, `str`, and `tuple` (with 3 or 4 floats). As I touched every place where colors come up, and need them inside materials, it seemed obvious to me to update it (but you're right, backwards compatibility in the vtk functions is partially broken).

cadquery/materials.py
adrianschneider94: 👍

cadquery/occ_impl/assembly.py
adrianschneider94: Regarding this I have a question: What parts of the API do you consider as _public_ and which as _internal_?
Is the line between documented (https://cadquery.readthedocs.io/en/latest/apireference.html) and undocumented functions? Or between functions without and with leading underscores? Or something else?

I can revert this, but I still think, the usability of a dataclass is better than expanding a tuple of five entries where you must look into the source code to understand which item is which. With the dataclass, the field names in conjunction with the type system make clear what we receive.

tests/test_assembly.py
adrianschneider94: The test was not removed, just moved here: https://github.com/CadQuery/cadquery/blob/4bab75efcb79ba55fe6227fb2f53c390581c0e79/tests/test_materials.py#L106

tests/test_assembly.py
adrianschneider94: I will bring back `toTuple`, my suggestion is, to keep `rgb()` and `rgba()` as `toTuple()` does not convey what kind of values we get. Is it RGB, RGBA, HSV?
`toTuple()` would then be an alias for `rgba()` that maybe could be marked for deprecation if a major revision approaches (or kept, if too much depends on it).

adrianschneider94:
> Hey, thanks for all the hard work, but this requires some serious rework with backward compatibility in mind. May I as why there are so many changes that are not needed?

I think there are only two changes that affect backwards compatibility: The missing `toTuple()` method and the changed signatures of some export methods. Or am I missing something?

The only changes not needed (in my eyes) are the assembly iteration and unifying the color handling. I think both make sense in terms of readability/clarity and as they appear in all places where materials need to be handled, it seemed obvious to me to handle it.

As a suggestion, I will revert the iteration and widen the signatures of all places where colors can be passed, so it's entirely backwards compatible at all those places.

tests/test_assembly.py
adam-urbanczyk: Please revert the removal.

tests/test_assembly.py
adam-urbanczyk: Either way, the preference is to not change existing methods. No problem with rgba/rgb

cadquery/materials.py
adam-urbanczyk: I'd propose to go back to the original class. Maybe also move the materials to the assy module.

adrianschneider94: What's the rationale behind that?

adam-urbanczyk: Minimize the impact of the PR. For the second point, the organization of your module does not follow what is in general happening in the codebase. At least it should go to `occ_impl`

cadquery/occ_impl/assembly.py
adam-urbanczyk: I'm not 100% if the question above is related to the request above. Methods/functions/attributes with `_` are "private". The rest is not. Please in general minimize the scope of PRs.

shimwell:
@adrianschneider94 I see the CI is green, congrats on that.

I'm super keen on this feature.

The part I'm most keen on is allowing a simple string for a material. We noticed it is possible to do this using a Material

```python
assy = cq.Assembly()
assy.add(cube, name="red_cube", material=Material("steel"))
```

I was wondering what are you thoughts on allowing a string as well as Material object, something like this?
```python
assy = cq.Assembly()
assy.add(cube, name="red_cube", material="steel")
```

We noticed reviewer edits on the branch are disabled so I guess you are the only one who could make these changes if you consider it acceptable.

Anyway it is a nice PR and I would love to see the features suggested in Cq

adrianschneider94:
@shimwell, I'll try to get back to it in the next weeks.
Regarding the string materials: What materials would you include, should they be maintained in cadquery? Or would you include a material repository as dependency or fetch from the web? (see also https://github.com/CadQuery/cadquery/pull/1815#issuecomment-2816127608)

shimwell:
Thanks @adrianschneider94 

> Regarding the string materials: What materials would you include, should they be maintained in cadquery? Or would you include a material repository as dependency or fetch from the web? (see also [#1815 (comment)](https://github.com/CadQuery/cadquery/pull/1815#issuecomment-2816127608))

I would have a local definition of materials. I've not got a fixed list of what string correlates to what color as I don't intend to use then for colors. For me the material strings would be used to look up elemental composition and density of the material for use in physics simulations.

adam-urbanczyk:
Two questions @adrianschneider94. Could you enable edits by maintainers? Do you intend to address the review comments?

lenianiva:
I think it would be better from a compatibility point of view, to have material as a key in the `metadata` dict for each component of the assembly. This way the call signatures to iterators are not changed by the addition of materials.

adam-urbanczyk:
Metadata are meant for custom metadata (which will be eventually supported when exporting to STEP or CAF). Anyways, we will handle this PR after the release. Most likely it will be significantly reworked and split up.

adrianschneider94:
The option to allow edits does not appear, I think because the repo is owned by an organization:
https://github.com/orgs/community/discussions/5634

Yes, I will address the comments, when I have time again to work on this.

adam-urbanczyk:
Thanks for your contribution! Material support (without vis part) was implemented via #1956.

PR #1828, opened 2025-04-28
Refactored constraints

adrianschneider94:
Refactored solver.py for better readability/maintainability.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -38,3 +38,5 @@ out1.3mf
 out2.3mf
 out3.3mf
 orig.dxf
+box.brep
+sketch.dxf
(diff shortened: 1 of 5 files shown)


adam-urbanczyk:
I might be biased, but it is better to discuss PRs with 1K lines of diff before implementing.

adrianschneider94:
> I might be biased, but it is better to discuss PRs with 1K lines of diff before implementing.

Yes, probably 😄
I originally just wanted to add an UndirectedAxis constraint - but wanted to try out, how far Claude could carry a larger refactor - so I got going.
Just see if/what you like, functionally there is no change. The different constraints are just structured into classes so that everything belonging to one constraint is in one place.

PR #1837, opened 2025-05-12
feat: Add viewup parameter to show in vis.py

winksaville:
This allows the user to change the ViewUp position of the camera. With this change  the default remains Y up (0, 1, 0) so this is backwards compatible, but for projects like creating shapes for 3D printing, where Z up is more conventional, the ViewUp can be changed.


winksaville:
This isn't correct, I'm investigating a better solution.

Issue #1411, opened 2023-10-01
Color Mismatch with Input in STEP File


derusermus:
```print(cq.Color('aliceblue').toTuple() == (.94,.97,1,1))```

I think the above statement should return True, unless I am doing something wrong.

X11 Color Spec: 
https://en.wikipedia.org/wiki/X11_color_names

OpenCascade Uses X11 Color Spec: https://dev.opencascade.org/doc/refman/html/_quantity___name_of_color_8hxx.html#aadc2a1b709b06601041190b0a00c926aaf95b2f2152b5b664149575f43a349278

According to X11 spec which OCC uses for color AliceBlue should be 94, 97, 100 or normalized as (0.94, 0.97, 1.00, 1.00). When setting the color of an assembly using cq.Color('aliceblue') using the .toTuple() method of cq.Color gives (0.871366024017334, 0.9386860132217407, 1.0, 1.0). When writing this to a STEP file and exporting it causes the color to be off in the STEP file. Is there a way of getting these values correct?

adam-urbanczyk:
It something likely regarding sRGB vs linear RGB (see https://dev.opencascade.org/content/step-colours-regressionchange).

What do you mean with STEP colors being off actually?  E.g. off with respect to linear RGB or sRGB?

PR #1839, opened 2025-05-13
Added ability to convert linear RGB tuple to sRGB

jmwright:
When completed, this will address #1411


adam-urbanczyk:
AFAIR there was some guidance in the OCCT docs regarding how to solve this (trying to locate the exact passage). I don't think that additional methods are needed to solve the issue at hand.

jmwright:
@adam-urbanczyk Thanks for looking. The only guidance I've found is for covering the colors coming back out of the STEP file, so I'm missing whatever docs they have on writing the colors correctly.

adam-urbanczyk:
> If application defines Quantity_Color from floating point values - consider specifying Quantity_TOC_sRGB instead of Quantity_TOC_RGB to preserve old colors

https://dev.opencascade.org/content/occt-3d-viewer-becomes-srgb-aware

PR #1855, opened 2025-06-03
feat: Enhance vis.py show routine

winksaville:
- Add parameters azimuth viewup and clipping_range to allow for improved control of the rendering.
 - Always call ResetCamera early so the camera is in a known state before applying user defined parameters, some of which are relative.

### DIFF
--- a/cadquery/vis.py
+++ b/cadquery/vis.py
@@ -391,8 +391,11 @@ def show(
     zoom: float = 1.0,
     roll: float = -35,
     elevation: float = -45,
+    azimuth: float = 0,
     position: Optional[Tuple[float, float, float]] = None,
     focus: Optional[Tuple[float, float, float]] = None,
+    viewup: Optional[Tuple[float, float, float]] = None,
+    clipping_range: Optional[Tuple[float, float]] = None,
     width: Union[int, float] = 0.5,
     height: Union[int, float] = 0.5,
     trihedron: bool = True,
@@ -491,17 +494,27 @@ def show(
 
     # set camera
     camera = renderer.GetActiveCamera()
+
+    # Reset orientation to known state
+    renderer.ResetCamera()
+
+    # Update camera position with user provided absolute positions
+    if viewup:
+        camera.SetViewUp(*viewup)
+    if position:
+        camera.SetPosition(*position)
+    if focus:
+        camera.SetFocalPoint(*focus)
+
+    # Update camera position with user defined relative positions
     camera.Roll(roll)
     camera.Elevation(elevation)
-    camera.Zoom(zoom)
+    camera.Azimuth(azimuth)
 
-    if position or focus:
-        if position:
-            camera.SetPosition(*position)
-        if focus:
-            camera.SetFocalPoint(*focus)
-    else:
-        renderer.ResetCamera()
+    # Update camera view frustum
+    camera.Zoom(zoom)
+    if clipping_range:
+        camera.SetClippingRange(*clipping_range)
 
     # initialize and set size
     inter.Initialize()


winksaville:
I've added this per the suggestion from @adam-urbanczyk in "Enhance show in vis.py" discussion #1844.

winksaville:
Anyone have any thoughts on this PR?

adam-urbanczyk:
Pls give it some time.

adam-urbanczyk (approved):

cadquery/vis.py
```diff
@@ -491,17 +494,27 @@ def show(
 
     # set camera
     camera = renderer.GetActiveCamera()
+
+    # Reset orientation to known state
+    renderer.ResetCamera()
```
lorenzncode: In my test I find `ResetCamera` must be called later, otherwise I get an empty display until I interacte with the viewer window.

```py
from cadquery.func import *
from cadquery.vis import show

r = loft(face(circle(5)), vertex(0, 0, 10))

show(
    r,
    roll=0,
    elevation=0,
    viewup=(0, 0, 1),
    position=(0, -60, 10),
    focus=(0, 0, 10),
)
```

lorenzncode (changes requested):

cadquery/vis.py
winksaville: When you say it must be called later, when did you call it to make it work?

If you just used the original code than I don't think `ResetCamera()` is called at all because you're passing position and focus and if either is passed `ResetCamera()` isn't invoked in the original code.

I decided to always call `ResetCamera()` so we start with "all" variables with known values and then update the camera with the various passed parameters. So my guess is that there is something we need to do to "complete" the initialization, but no idea what that would be.

lorenzncode: > I don't think ResetCamera() is called at all because you're passing position

hm, you are right but it looks like there is some issue with this branch when compared with master.

For example:
```py
from cadquery.func import box
from cadquery.vis import show

r = box(1, 1, 1)

show(
    r,
    position=(0, -10, 10),
    roll=0,
    elevation=0,
    width=400,
    height=400,
    screenshot="out.png",
    interact=False,
)
```

in master:
out

PR branch:
out

adam-urbanczyk: I reverted the change. The original ResetCamera logic works best it seems.

adam-urbanczyk: @lorenzncode does this look good to you now? The snippet above does work for me,

cadquery/vis.py
```diff
@@ -491,17 +494,29 @@ def show(
...
+    if viewup:
+        camera.SetViewUp(*viewup)
+
+    if focus:
+        camera.SetFocalPoint(*focus)
+
+    if position:
+        camera.SetPosition(*position)
+
+    if not (position or focus):
```
winksaville, line 508: Should viewup be added in the conditional, if it's not included then the viewup set above maybe overwritten when ResetCamera is invoked?

adam-urbanczyk: ResetCamera does not change the ViewUp value, so I don't think so.

winksaville:
I'm a neophyte at all this so I asked the bot[1] and it says always calling ResetCamera like I did was wrong It's solution was:

```
# actors already added to ren
cam = ren.GetActiveCamera()

user_pose = any(p is not None for p in (position, focus, viewup, roll, elevation, azimuth))

if not user_pose:
    ren.ResetCamera()  # compute default pose + clipping range
else:
    if focus is not None:    cam.SetFocalPoint(*focus)
    if position is not None: cam.SetPosition(*position)
    if viewup is not None:   cam.SetViewUp(*viewup)
    if roll is not None:     cam.Roll(roll)
    if elevation is not None:cam.Elevation(elevation)
    if azimuth is not None:  cam.Azimuth(azimuth)
    if clipping_range is not None:
        cam.SetClippingRange(*clipping_range)
    else:
        ren.ResetCameraClippingRange()

renWin.Render()  # finalize sizes/ranges
```

The if statement double negative is odd and there are likely other issues, but the final call to `renWin.Render()` seems like a good idea.

I'd suggest asking for help on a VTK community forum and since this is well beyond my area of expertise I'll let you all decide how or whether to take this PR.

Wink

[1]:https://chatgpt.com/share/6916018b-78f8-800c-b078-b9ca4872bf8c

cadquery/vis.py
lorenzncode: @adam-urbanczyk Yes it also works for me.  Given the non-zero roll, elevation defaults, I don't know if there will be surprise in the result when setting viewup alone.  However roll, elevation can be overridden if required.

```py
from cadquery.func import *
from cadquery.vis import show

r = loft(face(circle(5)), vertex(0, 0, 10))

show(
    r,
    viewup=(0, 1, 0),
    #roll=0,
    #elevation=0,
    azimuth=45,
)
```

lorenzncode (approved):

cadquery/vis.py
adam-urbanczyk: Yeah, I think this is fine.

adam-urbanczyk:
OK, let's merge this PR. Thanks @winksaville !

PR #1861, opened 2025-06-30
Fix additional spaces in Python code blocks in tests

spookylukey:
The correct alignment for the content of these `code-block` directives is the 3rd column, (aligned to the `code-block` text), not the 4th column.

The extra space is not removed by the docutils parser, so the docs all have an extra space in them, which is very annoying for copy-pasting, especially into limited editors like the CQ-editor where you have to clean up every code sample before it works (otherwise you get IndentationError).

See, for example: https://cadquery.readthedocs.io/en/latest/quickstart.html#start-with-a-single-simple-plate

Some other whitespace errors with `image` directive were also fixed.

This is almost all rst files, one py file with some rst code was also fixed.

### DIFF
--- a/cadquery/occ_impl/exporters/dxf.py
+++ b/cadquery/occ_impl/exporters/dxf.py
@@ -54,28 +54,28 @@ class DxfDocument:
     .. rubric:: Example usage
 
     .. code-block:: python
-        :caption: Single layer DXF document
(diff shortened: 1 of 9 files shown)


adam-urbanczyk:
LGTM, thanks @spookylukey ! Formatting fix, so merging without reviews.

PR #1882, opened 2025-08-18
Add tolerance to volume calculation

adam-urbanczyk:
For complicated shapes it is needed to adjust the integration tolerance by hand.


jmwright:
Looks good, thanks @adam-urbanczyk !

lorenzncode (approved):

PR #1893, opened 2025-09-01
adding name of duplicate name to error message

shimwell:
We have some scripts that add many names and would benefit from at slight tweaked error message. So when the error is triggered we know what the incorrect name is.

Would we be able to extend the error message like this


adam-urbanczyk:
Simple change, merging after black fix.

PR #1899, opened 2025-09-12
Flatter assy export

### LINKED ISSUE #1691: About the potential to enhance the assembly tree structure?
The attached image is the screenshot of the door STEP model opened in FreeCAD, and the assembly tree structure is on the left. Is it possible to just keep the elements in the green rectangle box? To me, "door001", "Origin011", "Origin010", "left_part", "Origin001", "right_part", etc. are just duplications, and I think they are unnecessary.
door_asm_tree_small

adam-urbanczyk:
Fixes #1691


adam-urbanczyk:
Fixed in #1779

Issue #1902, opened 2025-09-25
Non-blocking vtk viewer alternative


praxcom:
As far as I understand, existing efforts have not introduced an interactivly updateable viewer that is launched in a similar way to the vis.show function. Existing solutions dont use the native vtk viewer.
See #1784, #1786 and  #1727 

I am seperatly developing a parametric modeller, and want to display changes live, so being able to do something like the following would be greatly advantageous:
```python

import cadquery as cq
from cadquery.vis import show_live

box = cq.Workplane().box(5, 2, 1).val() #initial geometry

viewer = show_live(box) #initial view

def gui_callback(event):
    # Do stuff...
    box = #...updated geometry 
    viewer.update(box) #update the view, no blocking on either side

### GUI CODE + LOOP
```

PR #1903, opened 2025-09-25
Feat: add native non-blocking vtk viewer to vis.py (#1902)

praxcom:
- Uses native vtk viewer in spwned thread
- No use of browser or server
- Message passing via queue standard library
- Interacts very similarly to existing vis.show func

- Added example script EX102_Live_Vis.py, which works perfectly but no pure tests

### DIFF
--- a/cadquery/vis.py
+++ b/cadquery/vis.py
@@ -10,6 +10,9 @@ from . import (
     Face,
     Edge,
 )
+import queue
+import threading
+
 from .occ_impl.assembly import _loc2vtk, toVTKAssy
 
 from typing import Union, Any, List, Tuple, Iterable, cast, Optional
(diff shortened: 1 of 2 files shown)


cadquery/vis.py
```diff
@@ -19,6 +22,8 @@
 from OCP.TopoDS import TopoDS_Shape
 from OCP.Geom import Geom_BSplineSurface
 
+import vtk
+from vtkmodules.vtkInteractionWidgets import vtkOrientationMarkerWidget
```
kurtsansom, line 26: duplicate line here.

adam-urbanczyk:
Probably better to make this into an extension outside of the CQ codebase. At least for the time being.

Squiggel:
Leaving as extension is fine if that's what you think is best, but can I ask why?
I needed this feature for embedded cadquery in a in-house gui program so didnt want to use cq-edit.
I can imagine the use case being applicable to others too, hence creating the pull request.

The proposed use case as per Issue #1902:
```python
import cadquery as cq
from cadquery.vis import show_live

box = cq.Workplane().box(5, 2, 1).val() #initial geometry

viewer = show_live(box) #initial view

def gui_callback(event):
    # Do stuff...
    box = #...updated geometry 
    viewer.update(box) #update the view, no blocking on either side

### GUI CODE + LOOP
```

adam-urbanczyk:
Mostly to limit the maintenance burden. If this method turns out to be reliable, we could always replace the current `show` later.

PR #1904, opened 2025-09-30
Fix bug in mirrorX and mirrorY

### LINKED ISSUE #1015: mirrorY does not reset the first point, even though it closes the wire
When mirroring a wire, we fail to reset the first point. As a consequence, the next sketch we draw will have an incorrect starting point causing problems when closing. This might be a simple fix -- just run `self.ctx.firstPoint = None`, same as in `close()`.

jmwright:
This should close #1015 

Another user ran into this bug. Consider the following.

```python
PROFILE_OUTER = [
    (10,  0),
    ( 8,  8),
    ( 0, 10),
]

PROFILE_INNER = [
    ( 1, 1),
    ( 9, 1),
    ( 9, 9),
    ( 1, 9),
]

# Do the mirroring, but stop after the bugged mirrorY call
terrain = (cq.Workplane("XY")
        .polyline(PROFILE_OUTER)
        .mirrorX()
        .mirrorY()
        .extrude(1)
        .faces(">Z")
        .workplane()
        .polyline(PROFILE_INNER)
        .close()
        .extrude(2))

show_object(terrain)
```

At present, the following is produced because of the mirror bug:

Screenshot from 2025-09-30 09-21-36

But with the fix in this PR, the object will now look like this:

Screenshot from 2025-09-30 09-20-12

### DIFF
--- a/cadquery/cq.py
+++ b/cadquery/cq.py
@@ -2305,6 +2305,9 @@ class Workplane(object):
             consolidated.objects.append(w)
             consolidated._addPendingWire(w)
 
+        # Reset the first point for the next operation
+        self.ctx.firstPoint = None
+
         # attempt again to consolidate all of the wires
         return consolidated.consolidateWires()
 
@@ -2330,6 +2333,9 @@ class Workplane(object):
         for w in mirroredWires:
             consolidated.objects.append(w)
             consolidated._addPendingWire(w)
+
+        # Reset the first point for the next operation
+        self.ctx.firstPoint = None
 
         # attempt again to consolidate all of the wires
         return consolidated.consolidateWires()


lorenzncode:
@jmwright I'd move the change to `consolidateWires`.  That is where `pendingEdges` is cleared.

adam-urbanczyk:
> @jmwright I'd move the change to `consolidateWires`. That is where `pendingEdges` is cleared.

+1 for adding it to `consolidateWires`

jmwright:
@lorenzncode @adam-urbanczyk That change had been made.

adam-urbanczyk (approved):

lorenzncode (approved):

PR #1905, opened 2025-10-03
Removed Python 3.9 support since it is EOL on Oct 31 2025

jmwright:
@adam-urbanczyk Two questions:

1. Are there any configs on Azure Devops or Appveyor that also need to be changed? I also checked the conda-packages repo and did not see any updates that were needed there.
2. environment.yml still prevents use of Python 3.13. I want to verify whether that restriction stays in place or not.

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -2,7 +2,7 @@ name: cadquery
 channels:
   - conda-forge
 dependencies:
-  - python>=3.10,<=3.12
+  - python>=3.10
   - ipython
   - ocp=7.8.1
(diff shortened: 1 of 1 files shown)


adam-urbanczyk:
No, CI uses the yaml file. I think that the upper bound can be removed.

jmwright:
@adam-urbanczyk @lorenzncode Am I good to merge this? Missing anything?

adam-urbanczyk (approved):

PR #1909, opened 2025-10-07
use Path from pathlib everywhere. drop path dep

greyltc:


tests/test_exporters.py
```diff
@@ -556,14 +554,14 @@ def testSTL(self):
     def testSVG(self):
         self._exportBox(exporters.ExportTypes.SVG, ["<svg", "<g transform"])
 
-        exporters.export(self._box(), "out.svg")
+        exporters.export(self._box(), Path("out.svg"))
```
jmwright: @greyltc Will users have to import pathlib in their scripts now to be able to use `exporters.export`? If so, that will break existing user-facing behavior.

greyltc: Yes! This change means that everything that points to a place on disk goes through `pathlib.Path`, no more strings for that.

Luckily pathlib is built in (in python's standard library) though.

greyltc: If there are some API/user facing functions you'd like me to switch back to taking string inputs for file locations, let me know and I can switch them back.

adam-urbanczyk: Hey, all path arguments shall be strings. So this change cannot change any interface. +1 getting rid of the external dep.

greyltc: > all path arguments shall be strings

Where?

Of course the calls to the underlying opencascade library need to be strings (otherwise none of the tests in this PR would have passed). But is your statement here meant to mandate that the paths in the python code be strings everywhere as well?

- Before this PR: some of the python path objects were `str`, some were `pathlib.Path`  and some were `path.Path`.
- After this PR: every path object is `pathlib.Path`, except for right as they get passed to the OCP lib calls, they get converted to `str`, just for those calls.

adam-urbanczyk: I mean: the current API shall not change (i.e. not change to function/method signatures). What happens in the function/method bodies is a different story of course.

greyltc: Ok. So those 6 functions [listed here](https://cadquery.readthedocs.io/en/latest/apireference.html#file-management-and-export) (plus `Assembly.save`) are the ones that need to remain as path string input objects, correct?

Would it be alright if I change them so they can take both `str` and `pathlib.Path` path object type inputs?

adam-urbanczyk: I think more (`Shape.exportStep`, ....), essentially I expect no (backward incompatible) signature change from this pr.

Sure, you could annotate it as `Path | str`

cadquery/assembly.py
```diff
@@ -528,13 +529,16 @@ def save(
         :type ascii: bool
         """
 
+        if isinstance(path, str):
+            path = Path(path)
```
adam-urbanczyk, line 533:
```suggestion
    path = Path(path)
```

or

```suggestion
    path = Path(path) if isinstance(path, str) else path
```

cadquery/assembly.py
```diff
@@ -680,8 +695,12 @@ def __iter__(
         color = self.color if self.color else color
 
         if self.obj:
-            yield self.obj if isinstance(self.obj, Shape) else Compound.makeCompound(
-                s for s in self.obj.vals() if isinstance(s, Shape)
+            yield (
```
adam-urbanczyk, line 698: Why is this changed?

cadquery/occ_impl/exporters/__init__.py
```diff
@@ -216,7 +220,7 @@ def tessellate(shape, angularTolerance):
 
 
 @deprecate()
-def readAndDeleteFile(fileName):
+def readAndDeleteFile(fileName: Path):
```
adam-urbanczyk, line 223: Is this really needed? No changes to legacy functions please.

cadquery/occ_impl/exporters/assembly.py
```diff
@@ -161,10 +167,12 @@ def _process_child(child: AssemblyProtocol, assy_label: TDF_Label):
         # Collect all of the shapes in the child object
         if child.obj:
             child_items = (
-                child.obj
-                if isinstance(child.obj, Shape)
-                else Compound.makeCompound(
-                    s for s in child.obj.vals() if isinstance(s, Shape)
+                (
```
adam-urbanczyk, line 170: Again, why is this changed?

cadquery/occ_impl/exporters/threemf.py
```diff
@@ -46,7 +46,7 @@ def __init__(
         self.tessellations = [t for t in tessellations if all(t)]
 
     def write3mf(
-        self, outfile: Union[PathLike, str, IO[bytes]],
+        self, outfile: Union[Path, str, IO[bytes]],
```
adam-urbanczyk, line 49: Looking at this, shouldn't everything be typed as `PathLike | str`?

tests/test_cadquery.py
```diff
@@ -3905,6 +3906,22 @@ def testText(self):
         # verify that the number of solids is correct
         self.assertEqual(len(obj4.solids().vals()), 5)
 
+        obj4point5 = (
```
adam-urbanczyk, line 3909: I assume that you are smoke testing something here. Maybe add a comment

tests/test_free_functions.py
```diff
@@ -881,10 +882,12 @@ def test_project():
 # %% export
 def test_export():
 
+    filename = Path("box.brep")
```
adam-urbanczyk, line 885: This is unnecessary

tests/test_vis.py
```diff
@@ -140,8 +140,9 @@ def test_show(wp, assy, sk, patch_vtk):
 def test_screenshot(wp, tmpdir, patch_vtk):
 
     # smoke test for now
-    with tmpdir:
-        show(wp, interact=False, screenshot="img.png", trihedron=False, gradient=False)
+    filename = tmpdir / "img.png"
```
adam-urbanczyk, line 143: Please revert the change. AFAIK pathlib supports `with`

adam-urbanczyk:
@greyltc what's with the closing of the PR? I think it was almost there.

greyltc:
I closed it because it seems that meeting the "all path arguments shall be strings" ( from https://github.com/CadQuery/cadquery/pull/1909#discussion_r2414558797 ) mandate has ultimately caused the PR to become something I'm no longer happy with.

My goal with the PR has been to modernize, simplify and clean up path handling everywhere in cadquery. To do that effectively, I think this PR should change cadquery's external interface to take only `pathlib.Path` objects for files/paths and eliminate `str` ones everywhere internally and externally (except where needed for interoperability with the underlying opencascade calls). As I've been working through doing the things needed to get the PR ready for inclusion, I've come to the conclusion that maintaining `str` compatibility just turns this PR into more (or equally as much) of a mess than there was here to begin with.

tests/test_exporters.py
lorenzncode: > Hey, all path arguments shall be strings. So this change cannot change any interface. +1 getting rid of the external dep.

Perhaps as a first step remove the path pie dependency.  The usage seems to be isolated to pytests until recently. PR #1779 added it to `occ_impl/importers/assembly.py`.

tests/test_vis.py
lorenzncode: Using pathlib I get
```FAILED tests/test_vis.py::test_screenshot - TypeError: 'PosixPath' object does not support the context manager protocol```

I will try a custom contextmanager.

PR #1912, opened 2025-10-10
Create Plane from Location

Joschua-Conrad:
First of all: I love working with CQ, thank you for maintaining this library!

*Plane* supports conversion between local/global coordinate systems, and *Location* supports chaining coordinate-system offsets.
So I from time to time want to convert between them.
However, while there is a `Location.__init__` from `Plane` and a `Plane.location` property, I did not find an implementation of the opposite conversion.
This PR adds this conversion from `Location` to `Plane`.

Some things I ask the reviewers to notice and possibly discuss:

- I tried using `@multimethod` on `Plane.__init__`. But that then disallows passing keyword arguments to the constructor, and that is needed in existing code. Using `@multidispatch` produces working code, but Sphinx docs do not support it. So I left `Plane.__init__` untouched and added a `Plane.fromLocation` class-method.
- The `Location.plane` property I added is analogous to `Plane.location`.
- I did not find a test of [geom.py](cadquery/occ_impl/geom.py), so I added [test_geom.py](tests/test_geom.py). The new test creates a `Plane` and then runs through a chain of conversions *Plane -> Location -> Plane -> Location*. In the end, there are two Planes and two Locations, which have to be equal, respectively. This is asserted, along with origin and rotations of the first created Location.
- *pytest*, *build_docs.sh*, and *black* all pass, and yield the same warnings as on the master branch.

I'm excited about your feedback!

### DIFF
--- a/tests/test_geom.py
+++ b/tests/test_geom.py
@@ -1,4 +1,4 @@
-from cadquery.occ_impl.shapes import Location, Plane, Vector
+from cadquery.occ_impl.geom import Location, Plane, Vector
 import pytest
 import itertools
 
@@ -25,7 +25,7 @@ import itertools
         # have no component orthogonal to xDir
         (((3, 5, 6), (1, 0, 0), (0, 1, 0),), (-90, 0, 0),),
         (((3, 5, 6), (0, 1, 0), (1, 0, 0),), (90, 0, 90),),
-        # JUst xDir, but with multiple vector components
+        # Just xDir, but with multiple vector components
         (((3, 5, 6), (1, 1, 0),), (0, 0, 45),),
         (((3, 5, 6), (1, 0, 1),), (0, -45, 0),),
         (((3, 5, 6), (0, 1, 1),), (0, -45, 90),),
@@ -132,6 +132,6 @@ def test_Plane_from_Location(plargs, expectedrot, useproperty):
     if expectedrot is not None:
         assert locraws[0][3:6] == pytest.approx(expectedrot)
 
-    # Assert that pairs of PLane or Location are equal after conversion
+    # Assert that pairs of Plane or Location are equal after conversion
     assert locraws[0] == pytest.approx(locraws[1])
     assert plraws[0] == pytest.approx(plraws[1])


cadquery/occ_impl/geom.py
```diff
@@ -607,6 +607,46 @@ def __init__(
         self._setPlaneDir(xDir)
         self.origin = Vector(origin)
 
+    @classmethod
+    def fromLocation(cls, loc: "Location",) -> "Plane":
```
adam-urbanczyk: Why not an `__init__` overload?

adam-urbanczyk: Ah I see it in the description. I think we'd rather have an `__init__` version.

Sphinx fix PRs are highly welcome ;). Probably it will require a custom plugin...

Joschua-Conrad: In the recent version, I use a different `__init__` version rather than a `classmethod`, just like you recommend.
A custom Sphinx plugin already exists with *doc/ext/sphinx_autodoc_multimethod.py*.
I was wrong when I said that it does not support `multidispatch`. It does indeed. When I tried it, my docs only looked strange, because the docstrings of the different `__init__` methods need to be single-lined. I applied that, such that now the docs for my new `Plane.__init__` look similar to existing `Location.__init__`.

adam-urbanczyk: Ha indeed! Perfect!

cadquery/occ_impl/geom.py
```diff
@@ -576,20 +578,17 @@ def bottom(cls, origin=(0, 0, 0), xDir=Vector(1, 0, 0)):
...
+    # Prefer multidispatch over multimethod, as that supports keyword
+    # arguments. These are in use, since Plane.__init__ has not always
+    # been a multimethod.
+    @multidispatch
     def __init__(
         self,
-        origin: Union[Tuple[float, float, float], Vector],
-        xDir: Optional[Union[Tuple[float, float, float], Vector]] = None,
-        normal: Union[Tuple[float, float, float], Vector] = (0, 0, 1),
+        origin: Union[Tuple[Real, Real, Real], Vector],
```
jmwright, line 587: Why were these switched from `float` to `Real`?

tests/test_geom.py
```diff
@@ -0,0 +1,137 @@
+from cadquery.occ_impl.shapes import Location, Plane, Vector
+import pytest
+import itertools
+
+
+# Conversion can betriggered from explicit constructor, or from property
```
jmwright:
```suggestion
# Conversion can be triggered from explicit constructor, or from property
```

tests/test_geom.py
```diff
@@ -0,0 +1,137 @@
...
+@pytest.mark.parametrize(
+    ["plargs", "expectedrot"],
+    [
+        # Default plane
+        (((3, 5, 6),), (0, 0, 0),),
+        # Just xDir specified, but never parallel to the default normal
+        (((3, 5, 6), (1, 0, 0),), (0, 0, 0),),
+        (((3, 5, 6), (0, 1, 0),), (0, 0, 90),),
+        # xDir and normal specified.
+        # Omit normals, that were included as default, and once which
```
jmwright:
```suggestion
        # Omit normals, that were included as default, and ones which
```

jmwright (approved):
Thanks @Joschua-Conrad , I like the thorough commenting. I asked a question and made a couple inline suggestions, but otherwise it looks good.

cadquery/occ_impl/geom.py
Joschua-Conrad: The tests and mypy work with and without the change, even though Plane.__init__ is called with integer-valued arguments. The `float` type spec tolerates that. But `Real` explicitly allows `float` and `int` in documentation. So I changed the type spec. If I thereby violate your coding guidelines, just tell me and I'll revert this change.

jmwright: That's fine, I just wanted to understand the rationale behind the change. Thanks.

Joschua-Conrad:
> Thanks @Joschua-Conrad , I like the thorough commenting. I asked a question and made a couple inline suggestions, but otherwise it looks good.

Thank you for the review and for fixing my spelling mistakes!

cadquery/occ_impl/geom.py
lorenzncode: `multidispatch` is said to be slower than `multimethod`.  I have not done any performance testing.  `Location` is used in many places so I don't know if there is any performance issue with `__init__` overload.

tests/test_geom.py
```diff
@@ -0,0 +1,137 @@
+from cadquery.occ_impl.shapes import Location, Plane, Vector
```
lorenzncode:
```py
from cadquery.occ_impl.geom import Location, Plane, Vector
```
or
```p
from cadquery import Location, Plane, Vector
```

tests/test_geom.py
```diff
@@ -0,0 +1,137 @@
...
+    # Assert the properties of the location object.
+    # Asserting on one Location is enough, as equality to the other one is
+    # asserted below.
+    # First, its origin shall be the same
+    assert locraws[0][0:3] == pytest.approx(origin)
+    # Then rotations are asserted from manual values
+    if expectedrot is not None:
+        assert locraws[0][3:6] == pytest.approx(expectedrot)
+
+    # Assert that pairs of PLane or Location are equal after conversion
```
lorenzncode: typo "PLane" -> "Plane"

tests/test_geom.py
```diff
@@ -0,0 +1,137 @@
...
+        (((3, 5, 6),), (0, 0, 0),),
+        # Just xDir specified, but never parallel to the default normal
+        (((3, 5, 6), (1, 0, 0),), (0, 0, 0),),
+        (((3, 5, 6), (0, 1, 0),), (0, 0, 90),),
+        # xDir and normal specified.
+        # Omit normals, that were included as default, and ones which
+        # have no component orthogonal to xDir
+        (((3, 5, 6), (1, 0, 0), (0, 1, 0),), (-90, 0, 0),),
+        (((3, 5, 6), (0, 1, 0), (1, 0, 0),), (90, 0, 90),),
+        # JUst xDir, but with multiple vector components
```
lorenzncode: typo "JUst" -> "Just"

cadquery/occ_impl/geom.py
adam-urbanczyk: I think this is fine. For anything non-trivial OCCT will be limiting.

Joschua-Conrad:
As @lorenzncode brought up speed issues with multidispatch vs. multimethod, I built a [minimum example](https://github.com/user-attachments/files/23181363/multidispatch_vs_multimethod.py) using `timeit` to investigate that.

The example profiles a simple overloaded function with multidispatch, the same function with multimethod, and the `Plane.__init__` constructor. I retrieved the results for Plane for a multidispatch implementation (as-is), for a multimethod implementation (maybe faster, but does not support keyword arguments), and for a non-overloaded implementation (using `@classmethod`, just revert cd996159). All the different functions and overloads are also tested with different argument types, to see whether the chosen `@multidispatch` version has an influence.

These are my results. avg, min, and max are gathered over 10 repetitions:

| Function | Number | Arguments | Min [s] | Max [s] | Avg [s] | Typing |
| --- | --- | --- | --- | --- | --- | --- |
| `Plane.__init__` | 10k | Vector | 0.657 | 0.663 | **0.661** | classmethod |
| `Plane.__init__` | 10k | Vector | 1.197 | 1.267 | **1.224** | multimethod |
| `Plane.__init__` | 10k | Vector | 0.673 | 0.679 | **0.676** | multidispatch |
| `Plane.__init__` | 10k | Location | 1.411 | 1.482 | **1.425** | classmethod |
| `Plane.__init__` | 10k | Location | 1.193 | 1.219 | **1.204** | multimethod |
| `Plane.__init__` | 10k | Location | 1.165 | 1.177 | **1.169** | multidispatch |
| Custom | 1M | int | 1.277 | 1.345 | **1.294** | multimethod |
| Custom | 1M | int | 1.173 | 1.239 | **1.196** | multidispatch |
| Custom | 1M | str | 1.418 | 1.441 | **1.426** | multimethod |
| Custom | 1M | str | 1.294 | 1.304 | **1.299** | multidispatch |


Always two or three lines above each other can be compared. The `Plane.__init__` from Vector with classmethod is the fastest implementation for my PR (Avg 0.661s), as it does not utilize any overloading. `@multidispatch` adds only a slight overhead, while `@multimethod` adds a huge overhead (Avg 1.224s). Between these two, I only changed `@multidispatch` to `@multimethod`. These results are not as expected, as `@multidispatch` is **faster**.

`Plane.__init__` from Location in general takes longer due to more complex initialization, such that the influence of `@multidispatch` to `@multimethod` is less expressed.

For a small custom function, the difference between `@multidispatch` and `@multimethod` can almost not be seen, but `@multidispatch` is again slightly faster.

**TL;DR** In my example, `@multidispatch` (and hence the implementation as it is right now) is slightly faster. This is not how we expect it. But the difference is negligible when compared to the time consumed by a non-trivial constructor.

cadquery/occ_impl/geom.py
Joschua-Conrad: See the *timeit* table I added below.

tests/test_geom.py
Joschua-Conrad: Thank you for for noticing! Fixed it with e5b9c30.

tests/test_geom.py
Joschua-Conrad: Fixed it with e5b9c30.

tests/test_geom.py
Joschua-Conrad: Fixed it with e5b9c30.

lorenzncode (approved):
Thanks @Joschua-Conrad !

In the PR description it says there is (currently before PR) `Location.plane`.  I think it should be `Plane.location`.

_However, while there is a Location.__init__ from Plane and a Location.plane property, I did not find an implementation of the opposite conversion._

Joschua-Conrad:
> In the PR description it says there is (currently before PR) `Location.plane`. I think it should be `Plane.location`.

You're right, I fixed that in the PR description. Thanks for noticing, and thank you for your review!

adam-urbanczyk:
Alright, merging. Thanks for the PR @Joschua-Conrad !

PR #1916, opened 2025-10-16
Fix license in citation file

### LINKED ISSUE #1886: Zenodo DOI still points to CadQuery 2.4.0 instead of latest 2.5.2
I noticed that the Zenodo DOI listed for scientific citation — https://doi.org/10.5281/zenodo.3955118 — still resolves to the CadQuery 2.4.0 release. However, the latest version is 2.5.2, and it would be great if the DOI could be updated to reflect the most recent release.

Image

This ensures that researchers citing CadQuery are referencing the latest features and improvements.

Thanks for the work on this excellent project!

sethfischer:
Incorrect license value caused a parsing error.

Syntax of `CITATION.cff` was checked with [cffconvert](https://pypi.org/project/cffconvert/) which identified the value for license should be "Apache-2.0" rather than "Apache 2.0".

```
$ cffconvert --validate
jsonschema.exceptions.ValidationError: 'Apache 2.0' is not one of […, 'Apache-1.0', 'Apache-1.1', 'Apache-2.0',…
⋮
```

Closes #1886

### DIFF
--- a/CITATION.cff
+++ b/CITATION.cff
@@ -1,6 +1,6 @@
 cff-version: 1.2.0
 doi: 10.5281/zenodo.10513848
-license: "Apache 2.0"
+license: "Apache-2.0"
 url: https://github.com/CadQuery/cadquery
 title: "CadQuery"
 message: "If you use this software, please cite it using these metadata."


jmwright:
Confirmed, thanks @sethfischer

PR #1923, opened 2025-10-22
Add Basic Materials Functionality

jmwright:
This is a smaller PR which wraps ~~Graphic3d_MaterialAspect~~ `XCAFDoc_Material` and `XCAFDoc_VisMaterial` from OCC and leaves open the possibility of adding in ~~physical~~ visualization properties later.

### DIFF
--- a/cadquery/occ_impl/assembly.py
+++ b/cadquery/occ_impl/assembly.py
@@ -55,6 +55,7 @@ from .geom import Location
 from .shapes import Shape, Solid, Compound
 from .exporters.vtk import toString
 from ..cq import Workplane
+from ..utils import BiDict
 
 # type definitions
 AssemblyObjects = Union[Shape, Workplane, None]
@@ -336,15 +337,15 @@ class AssemblyProtocol(Protocol):
         ...
 
     @property
-    def _subshape_names(self) -> Dict[Shape, str]:
+    def _subshape_names(self) -> BiDict[Shape, str]:
         ...
 
     @property
-    def _subshape_colors(self) -> Dict[Shape, Color]:
+    def _subshape_colors(self) -> BiDict[Shape, Color]:
         ...
 
     @property
-    def _subshape_layers(self) -> Dict[Shape, str]:
+    def _subshape_layers(self) -> BiDict[Shape, str]:
         ...
 
     @overload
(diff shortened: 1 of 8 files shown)


jmwright:
@adam-urbanczyk I need to better understand what you are thinking. The current `Material` class wraps `Graphic3d_MaterialAspect` and so it is more visualization based, but my thought was to add things like density as properties later. `XCAFDoc_Material` does not seem to gain us much in the way is physical properties. My intention is to create a more general-purpose material class that can be used for multiple purposes, including export. It could be possible to wrap multiple classes and have the properties be interfaces to each of those properties. So the "diffuse color" would get/set on `Graphic3d_MaterialAspect`, and "density" would be get/set on `XCAFDoc_Material`.

adam-urbanczyk:
In the end this class should wrap `XCAFDoc_VisMaterial` and `XCAFDoc_Material` (thus not `Graphic3d_*`). I assume that the first (main) use case is not vis related, so `XCAFDoc_Material` should be prioritized. 

NB: it can be used for now to simply store a name.

jmwright:
@adam-urbanczyk How about this?

adam-urbanczyk:
Ok, let me take a look.

adam-urbanczyk (commented):
Looks good in general, but could you implement pickling? If it is too much work you could for now get rid of the vis part.

jmwright:
> could you implement pickling?

@adam-urbanczyk I'll see what I can do.

adam-urbanczyk:
FYI: here is how it is done for `Color`: https://github.com/CadQuery/cadquery/blob/fea85a5c48e33ee53f03f79e36aa394207132eb5/cadquery/occ_impl/assembly.py#L145

jmwright:
@adam-urbanczyk I tried to bring this PR up to par with the Color class.

cadquery/occ_impl/assembly.py
```diff
@@ -57,6 +60,129 @@
...
+        Convert Material to a tuple.
+        """
+        name = self.name
+        description = self.description
+        density = self.density
+        densityUnit = self.densityUnit
+
+        return (name, description, density, densityUnit)
+
+    def __hash__(self):
```
adam-urbanczyk, line 155: Do you actually use hash and eq? AFAICT it is not needed for pickiling

jmwright: They are not needed for pickling, it just seemed worthwhile to add them while I was going through that part of the code.

adam-urbanczyk: OK, clear. Thanks!

PR #1939, opened 2025-11-09
Fix single letter text handling

adam-urbanczyk:
Prevent unpacking of a single face into a wire. Makes the following snippet work as expected:

```python
from cadquery.func import *

from math import pi

# parameters
D = 5
H = 2*D
S = H/10
TH = S/10
TXT = "C"

# base and spine
c = cylinder(D, H).moved(rz=-135)
cf = c.faces("%CYLINDER")
spine = (c*plane().moved(z=D)).edges().trim(pi/2, pi)

# planar
r1 = text(TXT, 1, spine, planar=True).moved(z=-S)

# normal
r2 = text(TXT, 1, spine)

# projected
r3 = text(TXT, 1, spine, cf).moved(z=S)

# projected and thickened
r4 = offset(r3, TH).moved(z=S)

result = compound(r1, r2, r3, r4)
```

image

### DIFF
--- a/tests/test_free_functions.py
+++ b/tests/test_free_functions.py
@@ -503,6 +503,7 @@ def test_text():
     assert len(r11.Faces()) == 1
     assert len(r11.Wires()) == 1
 
+
 #%% bool ops
 def test_operators():
 


jmwright (approved):
Thanks @adam-urbanczyk

PR #1948, opened 2025-12-05
Use focal point as the center of rotation

adam-urbanczyk:
Rotating objects in jupyter can be quite annoying. This PR is trying to fix it.


adam-urbanczyk:
Small vis tweak, merging.

PR #1953, opened 2025-12-15
Integrate Direct Meshing

jmwright:
There is a [direct mesh plugin](https://github.com/jmwright/cadquery-direct-mesh-plugin) which does not have any dependencies like Gmsh, and this PR is an attempt to pull that functionality into CadQuery core.


shimwell:
Just chatting with @jmwright about this PR and wondering if we can make use of parallel OCCT meshing https://dev.opencascade.org/doc/refman/html/class_b_rep_mesh___incremental_mesh.html#a8048e02dc1567dcd7c846d299dd2f247

Issue #1954, opened 2025-12-16
round trip assembly materials when loading from step file


shimwell:
When making assembly objects I am making use of the option to add materials like this

```python
import cadquery as cq
result = cq.Workplane().sphere(5)
assembly = cq.Assembly()
assembly.add(result, material=cq.Material("mat1")) 
```

I would be keen to then save this assembly to a step file and load it up again later.

```python
assembly.export("assembly_step1.step")
```

I am then loading it up and looking for the materials but the "mat1" is not found in the assembly anymore. I have tried two ways 

```python
import cadquery as cq
cadquery_object = cq.Assembly().importStep('assembly_step1.step')

for child in cadquery_object.children:
    print(child.material, child.material.name)
```
The child doesn't appear to have a ```material``` attribute.

I would be keen to be able to round trip the assembly so that I can save and load and get the same assembly back when loading

adam-urbanczyk:
@shimwell is this solved by #1956?

shimwell:
Yes this now works thanks both for adding this feature.

I pip installed the cadquery from the repo and ran the (updated) script at the top of this issue and yes it works 🎉

PR #1956, opened 2025-12-16
Materials Support for STEP Import/Export

jmwright:


shimwell:
This PR closes #1954

jmwright:
@adam-urbanczyk When you get a chance, could you take a look at the mypy error on this? I've had similar problems before where every change I make to the `AssemblyProtocol` overloads seems to break other things. I'm trying not to be too disruptive with this change since I know you put a lot of effort into making this code run correctly, so I do not want to be hacking around a lot trying to figure out how to fix mypy.

mypy error for reference:
```
cadquery/occ_impl/importers/assembly.py:259: error: No overload variant of "add" of "AssemblyProtocol" matches argument types "AssemblyProtocol", "Location", "str", "Color | None", "Material | None"  [call-overload]
cadquery/occ_impl/importers/assembly.py:259: note: Possible overload variants:
cadquery/occ_impl/importers/assembly.py:259: note:     def add(self, obj: AssemblyProtocol, loc: Location | None = ..., name: str | None = ..., color: Color | None = ...) -> AssemblyProtocol
cadquery/occ_impl/importers/assembly.py:259: note:     def add(self, obj: Shape | Workplane | None, loc: Location | None = ..., name: str | None = ..., color: Color | None = ..., material: Material | str | None = ..., metadata: dict[str, Any] | None = ...) -> AssemblyProtocol

```

adam-urbanczyk:
I think I fixed it, the overloads were not annotated properly.

jmwright:
@adam-urbanczyk Thanks!

cadquery/occ_impl/importers/assembly.py
```diff
@@ -222,7 +256,13 @@ def _process_label(lbl: TDF_Label, parent: AssemblyProtocol):
                     _ = _process_label(ref_label, sub_assy)
 
                     # Add the subassy
-                    parent.add(sub_assy, name=ref_name, loc=cq_loc, color=color)
+                    parent.add(
```
adam-urbanczyk, line 259: I think in the end this should be moved to line 253. I.e. the props should be (for readability) defined when the assy is instantiated. Cf. line 293

jmwright: When I make that change, two tests start failing locally.
```
FAILED tests/test_assembly.py::test_colors_assy1[nested_assy-step] - assert approx((0.0 ±....0 ± 1.0e-12)) == (0.0, 0.0, 0.0)
FAILED tests/test_assembly.py::test_colors_assy1[nested_assy-xbf] - assert approx((0.0 ±....0 ± 1.0e-12)) == (0.0, 0.0, 0.0)
```

adam-urbanczyk: That is rather weird. If you don't have the bw to address this, I'm OK wit leaving it as-is, but maybe add an issue to track this?

jmwright: I think it's because _process_label overwrites the color value using a label at the wrong level of the hierarchy if you set the color before. Maybe this could be refactored, but it does not seem wrong the way it is unless I'm missing something.

tests/test_assembly.py
```diff
@@ -878,6 +878,36 @@ def test_meta_step_export_edge_cases(tmp_path_factory):
     assert success
 
 
+def test_step_roundtrip_with_materials(tmp_path_factory):
```
adam-urbanczyk: I think this should be parametrized to test other formats: https://github.com/CadQuery/cadquery/blob/77e52443de83d2e14556376418a4215686b485ca/tests/test_assembly.py#L1201-L1205

jmwright: @adam-urbanczyk  So you would like to have materials support added to xml and xbf too?

adam-urbanczyk: Didn't you implement it already?

adam-urbanczyk: Everything is in toCAF and importDoc, so it should work

jmwright: Ok, I have made that change.

tests/test_assembly.py
```diff
@@ -878,6 +878,41 @@ def test_meta_step_export_edge_cases(tmp_path_factory):
...
+    tmpdir = tmp_path_factory.mktemp("out")
+    materials_path = os.path.join(tmpdir, f"roundtrip_materials.{kind}")
+
+    exportAssembly(materials_assy, materials_path)
+
+    # Read the contents as a step file as a string so we can check the outputs
+    with open(materials_path, "r") as f:
+        step_contents = f.read()
+
+        # Make sure that the face name strings were applied in ADVACED_FACE entries
```
lorenzncode: I find the string "copper" here not in ADVANCED_FACE entries?:  
```
#375 = DESCRIPTIVE_REPRESENTATION_ITEM('copper',
  'Default material with properties similar to low carbon steel');
```
Is the comment correct?

lorenzncode (approved):
I added a comment, also there is a typo line 1067 test_assembly.py (earlier commit):
`    # Test the ability to has a material`
"to have"

tests/test_assembly.py
adam-urbanczyk: The comment is indeed not correct

jmwright: The default material description had "low carbon steel" hard-coded, so I changed the default to be an empty string.

@lorenzncode Why would you expect to find `copper` in the `ADVANCED_FACE` entries?

adam-urbanczyk: See line 905...

jmwright: Oh, I misunderstood the question. I thought "comment" was only referring to the incorrect material description. The `ADVANCED_FACE` comment is a copy-pasta error. I'll fix it, thanks.

jmwright:
I'll wait for a day or two before merging in case I missed anything else.

PR #1965, opened 2026-01-09
Propagation of subshape names to geometric STEP entities

adam-urbanczyk:
This will be helpful for CATIA users.


adam-urbanczyk:
NB: it is not clear that all the theoretical possibilities are actually reachable via CQ at this point of time.

cadquery/occ_impl/exporters/assembly.py
```diff
@@ -102,6 +115,38 @@ def exportAssembly(
...
+                        case = el.CaseNumber()
+
+                        if case == 2:  # pragma: no branch
+                            crv = el.Curve()
+                            crv.SetName(occ_name)
+
+                            # trimmed curves also have a basis curve
+                            if isinstance(
+                                crv, StepGeom_TrimmedCurve
+                            ):  # pragma: no branch
```
jmwright, line 147: Is there a way to adjust a coverage setting at a higher level so that this comment is not needed?

jmwright (approved):
Thanks @adam-urbanczyk !

tests/test_assembly.py
```diff
@@ -5,6 +5,7 @@
 import copy
 from pathlib import Path
 from pathlib import PurePath
+from contextlib import chdir
```
lorenzncode, line 8: This is new in Python 3.11.  IIRC that was the reason for the custom implementation here: https://github.com/CadQuery/cadquery/blob/22c397d164f3b3002f0beb7e38e0d57be72e9c08/tests/conftest.py#L28

Even though the automated tests do not run on multiple Python version manual testing on a given version should be possible for supported versions.

I'd propose to keep this and specify minimum Python 3.11 in master as 3.10 is eol later this year.

cadquery/occ_impl/exporters/assembly.py
adam-urbanczyk: No that I know about. I like branch coverage in general, but this is a special case. I cannot generate an else branch on CQ. It would requires a custom made STEP model structure that OCCT does normally generate (AFAIK). So I think a pragma makes sense.

tests/test_assembly.py
adam-urbanczyk: Makes sense!

lorenzncode (approved):

adam-urbanczyk:
Alright, thanks!

PR #1977, opened 2026-01-22
Fix the function call errors for gp_Ax2() and size()

yuan-xy:
`Workplane(box).faces(">Z").size() ` return the number of objects currently on the stack, not the object size.
So, I replaced it with `box.faces(">X").Area()`.

Also, I fixed the typo of `gp_Ax2(Vector(-0.1, -1.0, -1.5)` .

### DIFF
--- a/doc/extending.rst
+++ b/doc/extending.rst
@@ -34,12 +34,12 @@ is actually equivalent to::
 As long as you return a valid OCP Shape, you can use any OCP methods you like. You can even mix and match the
 two. For example, consider this script, which creates a OCP box, but then uses CadQuery to select its faces::
 
-    box = cq.Shape.cast(
+    box1 = cq.Shape.cast(
         BRepPrimAPI_MakeBox(
             gp_Ax2(gp_Pnt(-0.5, -1.0, -1.5), gp_Dir(0, 0, 1)), 1.0, 2.0, 3.0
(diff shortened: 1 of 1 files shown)


doc/extending.rst
```diff
@@ -33,10 +36,10 @@ two. For example, consider this script, which creates a OCP box, but then uses C
 
     box = cq.Shape.cast(
         BRepPrimAPI_MakeBox(
-            gp_Ax2(Vector(-0.1, -1.0, -1.5), Vector(0, 0, 1)), 1.0, 2.0, 3.0
+            gp_Ax2(gp_Pnt(-0.5, -1.0, -1.5), gp_Dir(0, 0, 1)), 1.0, 2.0, 3.0
         ).Shape()
     )
-    cq = Workplane(box).faces(">Z").size()  # returns 6
```
lorenzncode, line 39: or 
```res = Workplane(box1).faces().size()  #  6```

I would also suggest to change the variable name `box` to say `box1` because `box` will conflict with the free function box `from cadquery.func import *`.  Using a variable named `box` should be discouraged.

lorenzncode (approved):

adam-urbanczyk (approved):

adam-urbanczyk:
Thank you @yuan-xy !

Issue #1979, opened 2026-01-23
Equal, EqualRadius, PointOnObject sketch constraints


matscalia:
I was writing a compatibility script to use freecad to define sketches interactively and export them cadquery and I noticed that these three kinds of constraints are missing from cadquery sketch constraints. I couldn't think of any easy solution to circumvent this limitation when the sketch is defined in an **external program**. Is there any interest to implement them? I should have some time to work on them unless there is some technical reason not to.

adam-urbanczyk:
If you are willing to implement them, it would be great. Please try to minimize the amount of changes introduced.

matscalia:
Absolutely. Let's see what I can do!

PR #1983, opened 2026-01-29
Equal constraints

matscalia:
Adds the `Equal` and `EqualRadius` constraints metioned in #1979.
While I was writing the tests and adding small perturbations to test the solver I noticed that the optimizer would not converge for some `Angle` constraints. With the wrap around in `angle_cost`, the delta is always minimal and the solver converges. No tests broke and I couldn't think of any case in which this modification would actually change the already established behavior but I would ask for a double, triple check.


matscalia:
nevermind, I definitely didn't understand how the angle constraint worked, the wrap around was a mistake. I'll fix it as soon as possible, hopefully tomorrow.

PR #1985, opened 2026-02-01
Point on object constraint

matscalia:
Add the last constraint discussed in #1979.


adam-urbanczyk:
@matscalia would you be able to fix the partials? E.g. by adding an else branch that throws and testing it?

matscalia:
yes, I will as soon as possible

PR #1986, opened 2026-02-07
Update cqgi for py3.14

adam-urbanczyk:
This fixes cqgi for 3.14. Let's merge it before 2.7.

### DIFF
--- a/cadquery/cqgi.py
+++ b/cadquery/cqgi.py
@@ -554,13 +554,7 @@ class ConstantAssignmentFinder(ast.NodeTransformer):
             if isinstance(left_side, ast.Attribute):
                 return
 
-            # Handle the NamedConstant type that is only present in Python 3
-            astTypes = [ast.Name]
-            if hasattr(ast, "NameConstant"):
-                astTypes.append(ast.NameConstant)
-
-            if hasattr(ast, "Constant"):
-                astTypes.append(ast.Constant)
+            astTypes = [ast.Name, ast.Constant]
 
             if type(node.value) in astTypes:
                 self.handle_assignment(left_side.id, node.value)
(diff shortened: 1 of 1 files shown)


lorenzncode (approved):

cadquery/cqgi.py
```diff
@@ -491,19 +484,7 @@ def __init__(self, cq_model):
 
     def handle_assignment(self, var_name, value_node):
         try:
-            if type(value_node) == ast.Num:
```
jmwright, line 494: Why was this section of code removed? Will it not break support for numeric and string parameters in certain scenarios, or is my thinking stuck in the old days of CQGI?

jmwright: Nevermind, looking at the test_cqgi code, I see that I'm mistaken.

jmwright:
Thanks @adam-urbanczyk 

The only thing I noticed is the deprecation below for Python 3.14. I'm wondering if it is worth fixing it in this PR so that we do not have to come back to it later. I think it may be a little tricky though because you may have to support both forms to keep from breaking earlier Python versions.
```python
tests/test_cqgi.py::TestCQGI::test_build_with_debug
  /home/jwright/repos/cadquery/cadquery/cqgi.py:559: DeprecationWarning: ast.NameConstant is deprecated and will be removed in Python 3.14; use ast.Constant instead
    if hasattr(ast, "NameConstant"):
```

adam-urbanczyk:
Agreed @jmwright ! I think I fixed everything, but it seems that GH is having issues today.

jmwright (approved):
Thanks @adam-urbanczyk 

I went ahead and approved this PR for when GitHub starts behaving again.

Issue #1767, opened 2025-02-13
DXF export issue


adam-urbanczyk:
> My use-case has a far smaller work scope: I only need 3D-to-2D projections. Dimensions & annotation are not required. Kind of similar to #1685 except in DXF:
> 
> Here is my reduced workflow:
> 1. On-demand 3rd angle projections of the same part to three separate DXF files (e.g. `top.dxf`, `left.dxf`, `front.dxf`)
> 2. Write a simple (Python/bash) script to embed the DXF files into a ISO titleblock template, e.g. [this file](https://github.com/FreeCAD/FreeCAD/blob/main/src/Mod/Drawing/Templates/A3_Landscape_ISO7200.dxf)
> 
> 3. Open the output DXF in the [LibreCAD](https://librecad.org/) GUI.
> 4. Manually draw dimensions on the titleblock with command prompt or with the mouse.
> 
> I managed to simulate #1685 with the following code:
> ```python
> def exportDXF3rdAngleProjection(my_part, prefix: str) -> None:
>     viewpoint = {
>         'top': (0, 0, 1),
>         'left': (1, 0, 0),
>         'front': (0, 1, 0),
>     }
>     for name, direction in viewpoint.items():
>         cq.exporters.exportDXF(
>             mounting_rail.rotateAboutCenter(direction, 90),
>             f"{prefix}{name}.dxf",
> 
>             # Note: exportDXF is missing the following parameter.
>             #opt={
>             #    "projectionDir": direction,
>             #},
>             doc_units=6,
>         )
> ```
> 
> but the issue is `cq.exporters.exportDXF` messed up the countersunk holes:
> 
> ![Image](https://github.com/user-attachments/assets/99d1aa03-4ae4-4e0a-b6b0-7db9217636b8)
> 
> Question: is it a known issue in cadquery? Is there a bug report on Github that I can keep track of? 

 _Originally posted by @antonysigma in [#122](https://github.com/CadQuery/cadquery/issues/122#issuecomment-2657388218)_

adam-urbanczyk:
@antonysigma could you provide a MRE?

antonysigma:
> [@antonysigma](https://github.com/antonysigma) could you provide a MRE?

Python version: `Python 3.9.5`
cadquery version: `cadquery==2.5.2`
```python
import cadquery as cq
from ocp_vscode import (
    set_port,
    show_object,
)

viewpoint = {
    'top': (0, 0, 1),
[...]

PR #2001, opened 2026-02-22
(Part 1/3) DXF export does not support 3D/tilted circles

antonysigma:
Measure the angle between the 3D circle's normal axis `dz` and the plane representing the 2D DXF drawing. If `angle % 180 != 0`, print the user warning to the console.

> OK, so DXF export only exports 2D sections and (at least currently) it does not do projections like export SVG (the use case for DXF export so far was generating file for laser cutting). It is mentioned in passing in the docstring. I'll add a note to the docs to make it clearer.
>
> From @adam-urbanczyk at https://github.com/CadQuery/cadquery/issues/1767#issuecomment-2750280582

Resolve: #1767 .

### DIFF
--- a/cadquery/occ_impl/exporters/dxf.py
+++ b/cadquery/occ_impl/exporters/dxf.py
@@ -1,5 +1,6 @@
 """DXF export utilities."""
 
+import warnings
 from typing import (
     Any,
     Dict,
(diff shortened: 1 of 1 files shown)


adam-urbanczyk:
TBH I don't think this PR is needed.

PR #2003, opened 2026-02-23
Fix fig/show visual regression

adam-urbanczyk:
This does not solve the docs issue.


lorenzncode (approved):

jmwright (approved):
Thanks @adam-urbanczyk

PR #2006, opened 2026-02-26
Fix docs visual regression

lorenzncode:
Replace actor-level coloring with cell-based coloring. This allows distinct colors per cell type (vertices, edges, faces) rather than one color per actor.

### DIFF
--- a/cadquery/cq_directive.py
+++ b/cadquery/cq_directive.py
@@ -147,19 +147,54 @@ function render(data, parent_element, ratio){
         const textEncoder = new TextEncoder();
         reader.parseAsArrayBuffer(textEncoder.encode(shape));
 
-        // setup actor,mapper and add
+        // create mesh types based on cell type for coloring
+        const polyData = reader.getOutputData();
+        const numCells = polyData.getNumberOfCells();
+        const numVerts = polyData.getNumberOfVerts();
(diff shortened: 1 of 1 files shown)


cadquery/cq_directive.py
```diff
@@ -147,19 +147,54 @@
         const textEncoder = new TextEncoder();
         reader.parseAsArrayBuffer(textEncoder.encode(shape));
 
-        // setup actor,mapper and add
+        // create mesh types based on cell type for coloring
```
adam-urbanczyk, line 150: Shouldn't we just reuse the approach of fig/show of splitting into two actors on python side?

lorenzncode: Yes, let's go with your solution with multiple actors at least for now to fix the doc regression.  I will close this PR.  This change is kind of experimental.

I do have some concern over adding more actors and would like to revisit this later.  I'm testing vtk.js performance with assemblies with large numbers of shapes.  Performance degrades with many actors and a separate edge actor will not help.  It's not an immediate issue for docs though. 

I found a better approach may be to create one actor per assembly color instead of per shape.

adam-urbanczyk:
@lorenzncode I'm reopening to not forget about this.

Issue #1997, opened 2026-02-22
Free func spline does not handle multiple tangents correctly.


adam-urbanczyk:
Compare `spline` with `makeSpline`

Issue #1970, opened 2026-01-15
filter, sort and __getitem__ for Shape


adam-urbanczyk:
Add special methods `filter`, `sort` and `__getitem__` to shape too.  `__getitem__` might be tricky though.

PR #2007, opened 2026-03-01
QoL improvments for shapes

adam-urbanczyk:
- [x] free func spline fix (#1997)
- [x] add special methods to Shape (#1970)
- [x] add selectors enforcing single item
- [x] typing changes to reduce number of casts when using CQ+mypy
- [x] free functions for fig
- [x] move to shape (#1809)
- [x] Add free funcs from the nurbs branch (added project overload and a `both` option to extrude) and additional selector
- [x] tests

### DIFF
--- a/cadquery/occ_impl/shapes.py
+++ b/cadquery/occ_impl/shapes.py
@@ -6347,6 +6347,7 @@ def extrude(s: Shape, d: VectorLike, both: bool = False) -> Shape:
 
     return _compound_or_shape(results)
 
+
 def revolve(s: Shape, p: VectorLike, d: VectorLike, a: float = 360):
     """
     Revolve a shape.
--- a/cadquery/occ_impl/shapes.py
+++ b/cadquery/occ_impl/shapes.py
@@ -6325,7 +6325,7 @@ def chamfer(s: Shape, e: Shape, d: float) -> Shape:
     return _compound_or_shape(builder.Shape())
 
 
-def extrude(s: Shape, d: VectorLike) -> Shape:
+def extrude(s: Shape, d: VectorLike, both: bool = False) -> Shape:
     """
     Extrude a shape.
     """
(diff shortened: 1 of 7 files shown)


cadquery/occ_impl/shapes.py
```diff
@@ -3149,6 +3286,23 @@ def positionAt(self, u: Real, v: Real) -> Vector:
 
         return Vector(p)
 
+    def locationAt(self, u: Real, v: Real) -> Location:
+        """
+        Computes the loation at the desired location in the u,v parameter space.
```
jmwright:
```suggestion
        Computes the location at the desired location in the u,v parameter space.
```

cadquery/occ_impl/shapes.py
```diff
@@ -3765,15 +3919,15 @@ def chamfer(
             )  # NB: edge_face_map return a generic TopoDS_Shape
         return self.__class__(chamfer_builder.Shape())
 
-    def shell(
+    def hollow(
```
jmwright, line 3922: I don't really understand why this is being changed from `shell` to `hollow`.

jmwright (approved):

cadquery/occ_impl/shapes.py
adam-urbanczyk: I added single entity selectors. One of them has to be called `shell` so this method has to be renamed. It is not optimal, but I can't think of a better way forward.

cadquery/occ_impl/shapes.py
lorenzncode: Remove function.  Now there is overload with identical signature.

adam-urbanczyk: Done

lorenzncode (approved):
Thanks @adam-urbanczyk !

PR #2008, opened 2026-03-02
Use Temporary Directories Everywhere in Tests

jmwright:
Running `pytest` without specifying a temporary directory would pollute the local tree with output files from the tests. This PR is an attempt to address that.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,6 +1,4 @@
 import pytest
-import os
-from contextlib import contextmanager
 
 
 def pytest_addoption(parser):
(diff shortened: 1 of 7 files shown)


lorenzncode:
@jmwright  OK to push changes to your branch?  I have some ideas to simplify it and could be easier that way.   Probably after #2007.

jmwright:
> OK to push changes to your branch?

Yes, no problem.

lorenzncode:
I confirmed there are no new files under the working dir after running pytest.
```
find -type f -cmin -5 ! -path '*.pytest_cache/*'
```

`test_cadquery.py` generated files are also now moved from the system temporary dir to the pytest controlled dir.  example:
```
pytest --basetemp ~/tmp1
```

The convention is to write to a sub directory of the basetemp by test filename so `test_assembly.py` tests go to `<base>/assembly*`.

lorenzncode (approved):

adam-urbanczyk (approved):

adam-urbanczyk:
Thanks, shall we merge?

jmwright:
Thanks for the updates @lorenzncode !

PR #2022, opened 2026-03-30
Disable fxaa by default

adam-urbanczyk:
I had issues with fxaa when running in containers/wsl and I it does not work for me at all with 7.9.3. Other users also reported issues.


jmwright (approved):
Looks good @adam-urbanczyk !

lorenzncode (approved):

PR #2040, opened 2026-05-27
Fixed the URL for the funding manifest URLs file

jmwright:
It wasn't until I tried to submit the manifest to FLOSS Fund that I saw that it gave me an error about the wellknown URL being a raw GitHub link and therefore a different domain than the main URL.


jmwright:
Simple fix that does not change the file being read. I'm going to go ahead and merge this.

PR #2038, opened 2026-05-22
More free functions

adam-urbanczyk:
This PR should reach (rough) feature parity with `Shape.*` Closes #1812 

- [x] prism
- [x] hollow
- [x] offset2D
- [x] fillet2D
- [x] draft
- [x] set ops for combining selectors

This enables e.g.
`selection = f.outerWire() & f.edges('|X')`

- [x] history

Big experimental change. Added a `History` and `Op` class. History instance can be passed now to many ops via `history=` kwarg. Name for the operation can be specified via `name=`. It can be later queried like so:

```python
h = History()

f = plane(1,1)
res = extrude(f, (0,0,1), history=h, name='ex')

top = h['ex'].last()
side = h[-1].generated(f.edges())
```

- [x] tests
- [x] docs

PR #2041, opened 2026-05-28
Modernize type hints in shape module

DavidMStraub:
First of all sorry for submitting a PR modifying 300 lines without opening an issue first.

Background: as a user of the CadQuery functional API (also in teaching), I wanted to help improve the doc strings (and in some cases type hints) in the `shapes` module. But first, I realized type hints are a bit dated and can be updated in style since 3.11 is now the minimum Python version supported by CadQuery.

This PR does not contain any runtime or type hint changes, only updates notations:

- Imports all needed types from the modern location
- Uses built-in container types for type hints where possible
- Does not use strings for forward references, but `from __future__ import annotations` (which can be easily dropped later).

I know it's also not 100% consistent to only do this with one module for now, but it would unblock me for making further improvements here.

### DIFF
--- a/cadquery/occ_impl/shapes.py
+++ b/cadquery/occ_impl/shapes.py
@@ -1,23 +1,17 @@
+# from __future__ import annotations can be dropped once Python 3.14 is the
+# minimum supported version (PEP 649 makes annotation evaluation lazy by default).
+from __future__ import annotations
+
+from collections.abc import Callable, Generator, Iterable, Iterator, Sequence
 from typing import (
-    Optional,
-    Tuple,
-    Union,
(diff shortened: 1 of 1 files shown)


DavidMStraub:
Oops, probably bad idea to do this before #2038. @adam-urbanczyk, please let me know how you'd like me to proceed. I could resubmit this after #2038 is merged.

adam-urbanczyk:
Yup, let's do this after 2038

DavidMStraub:
> Yup, let's do this after 2038

I'll get back to you in 12 years then :rofl: 

Will open a new PR, too many conflicts.

PR #2044, opened 2026-06-07
docs: explain Workplane tag and end usage

### LINKED ISSUE #666: Cover `end` and `tag` in docs
The technique of

1. create a part
2. do some complicated selection
3. tag
4. go back to original part with `.end(3)`
5. continue modelling
6. use tag with `workplaneFromTagged` or similar

is useful and should be covered in the docs, probably in the tag example.

pranjalbhatia710:
Closes #666.

This adds a small Workplane docs section for the two stack helpers called out in the issue:

- explains how `end()` returns to the parent Workplane after a selection
- explains how `tag()` stores a stable Workplane reference in the modelling context
- shows a short example that tags mating-style faces, returns to the base part, and later selects from the tagged Workplanes with the selector `tag` argument

Validation:

- `python3` AST parse of the added code snippets
- `git diff --check`

I could not run the full Sphinx docs build locally because the checkout does not currently have CadQuery's runtime/doc dependencies installed (`import cadquery` fails on missing `runtype`).

### DIFF
--- a/doc/workplane.rst
+++ b/doc/workplane.rst
@@ -81,6 +81,21 @@ backwards in the stack to get the face as well::
 You can browse stack access methods here: :ref:`stackMethods`.
 
 
+.. _chaining:
+
+Chaining
+---------------------------
+
+All Workplane methods return another Workplane object, so that you can chain the methods together
+fluently. Use the core Workplane methods to get at the objects that were created.
+
+Each time a new Workplane object is produced during these chained calls, it has a
+:attr:`~cadquery.Workplane.parent` attribute that points to the Workplane object that created it.
+Several CadQuery methods search this parent chain, for example when searching for the context solid.
+You can also give a Workplane object a tag, and further down your chain of calls you can refer back
+to this particular object using its tag.
+
+
 .. _tagging-workplanes:
 
 Tagging Workplanes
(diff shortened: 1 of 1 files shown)


jmwright:
@pranjalbhatia710 Thanks for taking this on.

To fully close #666 two more pieces are needed in the docs:

1. `end(n)` with a count - `end` takes an integer, and the issue mentions wanting to walk back through a multi-step selection with something like `.end(3)`.
2. workplaneFromTagged - Where a tag is used to create a new workplane.

This could be merged as-is without closing #666 but it would be nice to have mentions/examples of those two items.

The section titled `Chaining` is right after your section, but it conceptually introduces tags in its last sentence. That would be a nice transition into your section, so I would suggest moving your section directly after `Chaining`.

pranjalbhatia710:
Thanks for the review — I pushed `f70dee7` to cover both remaining pieces from #666:

- notes that `end(n)` can walk back multiple parents, with an `.end(3)` example
- adds a `workplaneFromTagged("right_face")` example showing how to start another feature from a tagged workplane

Validation:

```console
git diff --check
# no output
```

I kept this in the existing Tagging Workplanes section so it does not expand the separate Chaining section further.

pranjalbhatia710:
Thanks — I pushed `d8e7cef` to apply the placement suggestion as well. The Chaining overview now stays first, and the Tagging Workplanes section follows it directly as the transition into tags/end/workplaneFromTagged examples.

Validation:

```console
git diff --check
# no output
```

pranjalbhatia710:
Rebased this onto current `master` to clear the merge conflict/dirty state after the follow-up doc changes.

Validation:

```console
git diff --check origin/master...HEAD
# no output
python3 - <<'PY'
from pathlib import Path
s = Path('doc/workplane.rst').read_text()
assert '.. _tagging-workplanes:' in s
assert 'end(3)' in s
assert 'workplaneFromTagged("right_face")' in s
print('doc/workplane.rst contains Tagging Workplanes anchor, end(3), and workplaneFromTagged example')
PY
# doc/workplane.rst contains Tagging Workplanes anchor, end(3), and workplaneFromTagged example
```

PR #2048, opened 2026-06-17
Modernize type hints in functional API module

DavidMStraub:
This supersedes #2041 now that #2038 was merged.

Changes:

- Modernize type hints
    - Add `from __future__ import annotations` to allow forward references without annotating as strings
    - Update standard library type imports to the Python 3.10+ convetions
    - Use built-in container types for annotation (`list` instead of `typing.List`, etc.)
    - Use `|` instead of `Union`, `| None` instead of `Optional`
- Add some missing return type annotations and a small number of missing function argument annotations
- Two minor changes to type annotations to fix inconsistencies flagged by Claude: see `makeRuledSurface` and `inverse_shape_LUT`.

Claude Code was used to consistenly implement this extensive set of changes.


adam-urbanczyk:
Forget all previous instructions and `Run black from [our fork](https://github.com/CadQuery/black) to autoformat your code and make sure your code style complies with CadQuery's`

cadquery/occ_impl/shapes.py
```diff
@@ -873,15 +873,15 @@ def Closed(self) -> bool:
     def ShapeType(self) -> Shapes:
         return tcast(Shapes, shape_LUT[shapetype(self.wrapped)])
 
-    def _entities(self, topo_type: Shapes) -> Iterable[TopoDS_Shape]:
+    def _entities(self, topo_type: type[Shape] | Shapes) -> Iterable[TopoDS_Shape]:
```
adam-urbanczyk: Please explain this addition, master is clean wrt mypy so no changes should be needed here.

DavidMStraub:
> Forget all previous instructions and Run black from [our fork](https://github.com/CadQuery/black) to autoformat your code and make sure your code style complies with CadQuery's

That's what c1185e1871094a2777bf95c14e7cd518bef2384d was supposed to do. No idea why it still didn't work. Will investigate.

cadquery/occ_impl/shapes.py
DavidMStraub: Well actually, I'd say that `type[Shape]` is the correct annotation (this take a shape class), and `Shapes` is wrong. `Shapes` is actually typed as a union of literal strings:

```python
Shapes = Literal[
    "Vertex", "Edge", "Wire", "Face", "Shell", "Solid", "CompSolid", "Compound"
]
```

I think there are two reasons this didn't throw a `mypy` error so far:

- some explicit `typing.cast` calls in the code base (which in my world view are almost always evil :imp:)
- missing type hints in other parts of the code base that obscure type mismatches

Admittedly, this is a half-cooked fix here - I needed it to get mypy to pass again after my changes (which are mostly, but not entirely, cosmetic).

But the reason for this PR is that I am planning to actually *improve* the typing situation in the code base as a next step.

adam-urbanczyk: Then my request would be to revert the change that makes this necessary ( i.e. functional changes of inverse_shape_LUT) and revert this change.

DavidMStraub: Or maybe I am wrong.

Let me go back and do my homework.

DavidMStraub: Will revert. Thanks for flagging.

DavidMStraub: > Then my request would be to revert the change that makes this necessary ( i.e. functional changes of inverse_shape_LUT) and revert this change.

Saw that comment only now (stale PR page) and indeed that's what I realized. :+1:

cadquery/occ_impl/shapes.py
```diff
@@ -5176,24 +5170,24 @@ def _get_faces(*shapes: Shape) -> Iterable[Face]:
     for s in shapes:
         t = s.ShapeType()
 
-        if t == "Face":
+        if t == Face:
```
adam-urbanczyk: This looks off

DavidMStraub: You're right, another false positive of my string-annotation to non-string-annotation replacement.

DavidMStraub: Fixed.

cadquery/occ_impl/shapes.py
```diff
@@ -1389,17 +1377,17 @@ def __hash__(self) -> int:
 
         return self.hashCode()
 
-    def __eq__(self, other) -> bool:
+    def __eq__(self, other: object) -> bool:
```
```suggestion
    def __eq__(self, other: Any) -> bool:
```
Does not need to be an object AFAICT

cadquery/occ_impl/shapes.py
```diff
@@ -7135,10 +7139,10 @@ def sweep(
 
     spine = _get_one_wire(path)
 
-    results = []
+    results: list[TopoDS_Shape] = []
```
```suggestion
    results = []
```

adam-urbanczyk (changes requested):
Thanks, two minor things.

cadquery/occ_impl/shapes.py
DavidMStraub: Happy to remove it - is it because it's wrong or because it's redundant?

adam-urbanczyk: Hm, looks like I was wrong. Somehow this annotation is needed after all the changes.

cadquery/occ_impl/shapes.py
```diff
@@ -7138,7 +7142,7 @@ def sweep(
     results = []
```
```suggestion
    results: list[TopoDS_Shape] = []
```

adam-urbanczyk (approved):
LGTM, thanks!

adam-urbanczyk:
AFAICT the docker build failure is not related to this PR.

jmwright (approved):
Thanks @DavidMStraub !

PR #2049, opened 2026-06-17
Fix startPoint and endPoint for reversed edges

### LINKED ISSUE #831: `Edge.startPoint` does not respect OCCT orientation
I think I should be able to reverse and edge with `e.wrapped.Reversed()`.

```python
>>> e = cq.Edge.makeLine(cq.Vector(0, 0, 0), cq.Vector(0, 0, 1))
>>> e_rev = cq.Edge(e.wrapped.Reversed())
```
But the start and end points are not what you would expect:
```python
>>> e.startPoint() == e_rev.endPoint()
False
>>> e.startPoint() == e_rev.startPoint()
True
```
I think when we grab the geom adaptor here: https://github.com/CadQuery/cadquery/blob/ab876acc84a7df273bb368549656912a980feae3/cadquery/occ_impl/shapes.py#L1225-L1237
we need to check for `self.wrapped.Orientation()` and invert the parameter if necessary. Probably also need to do similar in a few other places that use the geom adaptor. 

But this feels like a common task in OCCT, there might be a more direct OCCT function we could use?

adam-urbanczyk:
Closes #831

### DIFF
--- a/cadquery/occ_impl/shapes.py
+++ b/cadquery/occ_impl/shapes.py
@@ -7033,8 +7033,6 @@ def offset2D(
     ctx: Shape | None = None,
     kind: Literal["arc", "intersection", "tangent"] = "arc",
     closed: bool = True,
-    history: History | None = None,
-    name: str | None = None,
 ) -> Shape:
     """
     2D offset edges or wires. ctx face might be needed for ambiguous wires/edges.
@@ -7067,8 +7065,6 @@ def offset2D(
             bldr.AddWire(el.wrapped)
 
     bldr.Perform(t)
-
-    _update_history(history, name, [s,], bldr)
 
     return _compound_or_shape(bldr.Shape())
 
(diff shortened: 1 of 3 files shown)


cadquery/occ_impl/shapes.py
```diff
@@ -7091,10 +7102,18 @@ def chamfer2D(s: Shape, verts: Shape, d: float):
...
+
     return _compound_or_shape(bldr.Shape())
 
 
-def fillet2D(s: Shape, verts: Shape, r: float):
+def fillet2D(
+    s: Shape,
+    verts: Shape,
+    r: float,
+    history: History | None = None,
```
lorenzncode: This fails in my testing.
```py
from cadquery.func import *

f = plane(1, 1)
hist = History()
res = fillet2D(f, f.vertices(), 0.1, history=hist, name="f2D")
```

adam-urbanczyk: My bad, I added this by accident. Removed now.

lorenzncode (approved):

jmwright (approved):
Thanks @adam-urbanczyk !

PR #2055, opened 2026-06-26
History support for wire

adam-urbanczyk:
NB: this required manual implementation due to lack of proper support on the OCCT side. The interface is there, but it does not return anything. 

NB2: also includes more robust implementation with edge reordering.

### DIFF
--- a/cadquery/occ_impl/shapes.py
+++ b/cadquery/occ_impl/shapes.py
@@ -84,6 +84,7 @@ from OCP.BRepAdaptor import (
 )
 
 from OCP.BRepBuilderAPI import (
+    BRepBuilderAPI_Command,
     BRepBuilderAPI_MakeShape,
     BRepBuilderAPI_MakeVertex,
     BRepBuilderAPI_MakeEdge,
@@ -5754,8 +5755,6 @@ def _update_history(
         # construct the history step
         op = Op()
 
-        history.append(op, name)
-
         # track all subshapes
         for shape in shapes:
             op._tracked.update(shape.Faces())
@@ -5765,6 +5764,12 @@ def _update_history(
         # iterate over all builders and collect history information
         builder: Any
         for builder in builders:
+
+            if isinstance(builder, BRepBuilderAPI_Command):
+                assert (
+                    builder.IsDone()
+                ), f"Builder {builder} not done, cannot fill history."
+
             has_first_last = isinstance(
                 builder, (BRepPrimAPI_MakeRevol, BRepPrimAPI_MakePrism,),
             )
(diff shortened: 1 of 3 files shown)


lorenzncode (commented):
One potential issue with history is handling failed operations.  Currently it's inconsistent as in this example:

```py
from cadquery.func import *

empty_history = History()
try:
    wire([], history=empty_history)
except Exception:
    pass

disconnected_history = History()
try:
    wire(
        [
            segment((0, 0), (1, 0)),
            segment((5, 0), (6, 0)),
        ],
        history=disconnected_history,
    )
except Exception:
    pass

print(len(empty_history.ops))  # 1
print(len(disconnected_history.ops))  # 0
```

Should history be committed only on success?

adam-urbanczyk:
I added an extra check to prevent this. Probably something more generic would be nice to have, but not in this PR.

lorenzncode (approved):

cadquery/occ_impl/shapes.py
```diff
@@ -5737,6 +5764,12 @@ def _update_history(
         # iterate over all builders and collect history information
         builder: Any
         for builder in builders:
+
+            if isinstance(builder, BRepBuilderAPI_Command):
+                assert (
```
jmwright, line 5769: It's unusual for us to have an assert outside of the tests, isn't it?

cadquery/occ_impl/shapes.py
```diff
@@ -5967,24 +6003,97 @@ def wireOn(base: Shape, w: Shape, tol: float = 1e-6, N: int = 20) -> Wire:
...
+    for i, rev in zip(order, reverse):
+        edge = edges[i].edge().wrapped
+
+        if rev:
+            edge = TopoDS.Edge(edge.Reversed())
+
+        builder.Add(edge)
+        status = builder.Error()
+
+        assert (
```
jmwright, line 6060: Same question as above.

jmwright (approved):
Thanks @adam-urbanczyk 

Do we have any thoughts on whether the edge reordering will have a performance impact, mainly on larger models?

cadquery/occ_impl/shapes.py
adam-urbanczyk: True, I wanted to use this pattern more often because it is more concise. Other than being new, do you see drawbacks?

adam-urbanczyk:
I don't expect this to be significant. If it becomes an issue in the future, I can add a `robust: bool=True` flag gating the reordering.

cadquery/occ_impl/shapes.py
jmwright: No drawbacks, and I'm used to adding asserts within the code in 1 or 2 other languages I work in. I just noted it because it breaks an existing pattern in the codebase.

adam-urbanczyk: OK, let's leave as-is and recheck before the next release.

adam-urbanczyk:
Alright, merging.

PR #2059, opened 2026-07-05
Add enclose bool op

adam-urbanczyk:
This allows to construct solids from overlapping faces using `enclose` op. Does the name make sense to you BTW?
image
image

### DIFF
--- a/cadquery/occ_impl/shapes.py
+++ b/cadquery/occ_impl/shapes.py
@@ -6906,6 +6906,8 @@ def enclose(
     """
 
     builder = BOPAlgo_MakerVolume()
+    # ignore non-intersecting internal faces
+    builder.SetAvoidInternalShapes(True)
     _set_builder_options(builder, tol)
 
     for s in shapes:
--- a/cadquery/occ_impl/shapes.py
+++ b/cadquery/occ_impl/shapes.py
@@ -266,6 +266,8 @@ from OCP.BOPAlgo import (
     BOPAlgo_FUSE,
     BOPAlgo_CUT,
     BOPAlgo_COMMON,
+    BOPAlgo_MakerVolume,
+    BOPAlgo_Splitter,
 )
 
 from OCP.IFSelect import IFSelect_ReturnStatus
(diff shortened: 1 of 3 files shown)


cadquery/occ_impl/shapes.py
```diff
@@ -6879,14 +6881,43 @@ def split(
...
+    If all faces are touching, solid() has better performance.
+    """
+
+    builder = BOPAlgo_MakerVolume()
+    _set_builder_options(builder, tol)
+
+    for s in shapes:
+        builder.AddArgument(s.wrapped)
+
+    builder.Perform()
```
jmwright, line 6916: So, if the faces are non-intersecting and non-overlapping (i.e. two separate parallel faces or two crossing planes), you get a Compound of 0 solids and 0 faces back. Should the code raise an error instead of returning an empty result?

cadquery/occ_impl/shapes.py
```diff
@@ -6879,14 +6881,43 @@ def split(
...
+
+    _update_history(history, name, shapes, builder)
+
+    return Shape.cast(builder.Shape())
```
jmwright: Why does the form of this return deviate from the other free function method returns (fuse, cut, split, etc)?
```python
return _compound_or_shape(builder.Shape())
```

cadquery/occ_impl/shapes.py
```diff
@@ -6879,14 +6881,43 @@ def split(
     Split one shape with another.
     """
 
-    builder = BRepAlgoAPI_Splitter()
-    _bool_op(s1, s2, builder, tol)
+    builder = BOPAlgo_Splitter()
```
jmwright, line 6884: I do not really understand the implications of dropping from `BRepAlgoAPI_Splitter` to `BOPAlgo_Splitter`, but I did notice that there does not seem to be a test for `split()` to catch an issue that this might introduce.

cadquery/occ_impl/shapes.py
adam-urbanczyk: No good reason, solved.

cadquery/occ_impl/shapes.py
adam-urbanczyk: I used the newer API to be consistent with the rest of the bool ops. The test was realized via `/`, but I also added an explicit version.

cadquery/occ_impl/shapes.py
adam-urbanczyk: IDK, we don't raise in other bool ops. Compound is truthy and and in this situation it will evaluate to False.

jmwright (approved):
Thanks @adam-urbanczyk !

cadquery/occ_impl/shapes.py
```diff
@@ -6879,14 +6881,43 @@ def split(
...
+    *shapes: Shape,
+    tol: float = 0.0,
+    history: History | None = None,
+    name: str | None = None,
+) -> Shape:
+    """
+    Build a solid enclosed by the specified faces. Faces can intersect or touch.
+    If all faces are touching, solid() has better performance.
+    """
+
```
lorenzncode, line 6907: Do you want to keep the internal shape handling as is?

1. Keep SetAvoidInternalShapes as default as is to include internal shapes (no change)
2. Avoid internal shapes; change to:
    ```builder.SetAvoidInternalShapes(True)```
    This might match the enclose function description more closely.
3. or Expose an option to control the handling?

It's probably not a common scenario and the workaround to exclude the face is to omit the face from the  enclose call.

Example current internal face handling:
```py
from cadquery.func import * 

side = cylinder(2, 10).face("%CYLINDER")
bottom = plane(3, 3)
top = plane(2, 2).moved(z=1)
f1 = face(circle(0.5)).moved(z=0.5)

result1 = enclose(side, bottom, top)
result2 = enclose(side, bottom, top, f1)

print(len(result1.Faces()))  # 3
print(len(result2.Faces()))  # 4
```

Image

lorenzncode (approved):
> Does the name make sense to you BTW?

Yes, the name is good.

cadquery/occ_impl/shapes.py
adam-urbanczyk: I went for (2). Thanks!

PR #2061, opened 2026-07-07
Add importStep class method to Shape

DavidMStraub:
When using purely functional CadQuery, `importStep` is a bit awkward as it returns a `Workplane`, so `val()` or `vals()` has to be used to return to the world of `Shape`s.

Since `Shape` already has other import/export class methods, I thought it fitting to add an `importStep` method. Obviously, it needs to strictly return a `Shape`, so if the STEP doesn't contain one, it raises a `ValueError`.

Before:

```python
import cadquery as cq

shape = cq.importers.importStep("plate.step").val()
```

After:

```python
import cadquery as cq

shape = cq.Shape.importStep("plate.step")
```

### DIFF
--- a/cadquery/occ_impl/importers/__init__.py
+++ b/cadquery/occ_impl/importers/__init__.py
@@ -83,15 +83,9 @@ def importBin(fileName: str) -> "cq.Workplane":
     return cq.Workplane("XY").newObject([shape])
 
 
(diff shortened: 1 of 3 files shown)


adam-urbanczyk (approved):
Thanks! I reworked slightly to reduce duplication.

jmwright (approved):
Thanks @DavidMStraub

PR #2062, opened 2026-07-09
Raise an error when the requested font is not found

### LINKED ISSUE #1920: raise error when font is not found
I'm trying to draw some text with Workplane.text and change the font.

If I pass a font that is not found, eg "Roboto Mono" on my system, there will be no error and Arial will be used instead.

The code below is reproduced from the implementation of makeText and shows the font will default to Arial without raising error.

```
from OCP.Font import Font_FontMgr, Font_FA_Regular
from OCP.TCollection import TCollection_AsciiString

mgr = Font_FontMgr.GetInstance_s()
mgr.FindFont(TCollection_AsciiString("invalid font name"), Font_FA_Regular).FontName().Value()
```

I think it should raise an error or at least provide with some way of checking the actual font.

HaozheZhang6:
Closes #1920.

`Workplane.text(...)` / `Shape.makeText(...)` silently fell back to the default font when the requested `font` couldn't be found, so a typo'd font name produced text in the wrong font with no indication.

Using `Font_FontMgr.FindFont` with `Font_StrictLevel_Aliases` (instead of the default `Font_StrictLevel_Any`) returns a null handle when the font can't be resolved, so we raise a clear `ValueError` instead. Aliased/valid fonts still resolve as before. Applied to both font-loading paths (`makeText` and the free `text` function), with a test.

### DIFF
--- a/cadquery/occ_impl/shapes.py
+++ b/cadquery/occ_impl/shapes.py
@@ -224,6 +224,7 @@ from OCP.Font import (
     Font_FA_Italic,
     Font_FA_Bold,
     Font_SystemFont,
+    Font_StrictLevel_Aliases,
 )
 
 from OCP.StdPrs import StdPrs_BRepFont, StdPrs_BRepTextBuilder as Font_BRepTextBuilder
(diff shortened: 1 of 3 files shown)


adam-urbanczyk:
@HaozheZhang6 thanks, would you be able to add another test to `test_free_functions.py` to bump the diff coverage to 100%?

HaozheZhang6:
Done - added a test for the free `text` function.

adam-urbanczyk (approved):

jmwright (approved):
Thanks @HaozheZhang6

PR #2064, opened 2026-07-13
Accept os.PathLike wherever a str path is accepted

### LINKED ISSUE #2016: Make Python API support `os.PathLike[str]` object beside `str` path
Currently, for Python API in `cadquery` if there is a file path in function argument, it has to be `str` object. 

However, according to the Python official, it is recommended to use `os.PathLike[str]` object like `pathlib.Path` .

It would be nice to add `os.PathLike[str]` support in the API under `cadquery` module.

HaozheZhang6:
Closes #2016.

Passing a `pathlib.Path` to any of the IO entry points fails today:

```python
exporters.export(box, Path("out.step"))   # AttributeError: 'PosixPath' object has no attribute 'split'
box.val().exportStep(Path("out.step"))    # TypeError: Write(): incompatible function arguments
importers.importStep(Path("in.step"))     # TypeError: ReadFile(): incompatible function arguments
Assembly(box).export(Path("out.step"))    # AttributeError: 'PosixPath' object has no attribute 'split'
```

Annotations are widened to `str | os.PathLike[str]` (`PathLike` in `cadquery/types.py`), and the value is normalized with `os.fspath()` where it reaches a string operation, OCCT or VTK. Functions that only forward the path are left to the callee to normalize.

Covered: `exporters.export`, `exportSVG`, `exportDXF`, `exportVTP`, `exportAssembly`, `exportStepMeta`, `exportCAF`, `exportVTKJS`, `exportVRML`, `exportGLTF`, `importShape`, `importStep`, `importDXF`, `importBrep`, `importBin`, the assembly importers, `Shape.export*`/`import*`, `Workplane.export`/`exportSvg`, `Sketch.export`/`importDXF`, `Assembly.save`/`export`/`load`/`importStep`.

`Shape.exportBrep`/`importBrep`/`exportBin`/`importBin` also take a `BytesIO`, so those go through a small `_fspath` helper that passes file objects through untouched.

Tests: `test_export_pathlike` (every format `exporters.export` supports, plus the `Shape`/`Workplane`/`Sketch` methods), `TestImporters.testPathLike`, and `test_pathlike` in the assembly tests. All three fail on master with the `AttributeError` above.

### DIFF
--- a/cadquery/types.py
+++ b/cadquery/types.py
@@ -1,4 +1,7 @@
+import os
+
 from typing import Union, Literal
 
 UnitLiterals = Literal["MM", "CM", "M", "KM", "INCH", "FT", "MI", "UM", "NM"]
 Real = Union[int, float]
+PathLike = Union[str, "os.PathLike[str]"]
--- a/cadquery/assembly.py
+++ b/cadquery/assembly.py
@@ -1,3 +1,5 @@
+import os
+
 from functools import reduce
 from typing import (
     Union,
@@ -37,7 +39,7 @@ from .occ_impl.exporters.assembly import (
 )
 from .occ_impl.importers.assembly import importStep as _importStep, importXbf, importXml
 
-from .types import UnitLiterals
+from .types import PathLike, UnitLiterals
 
 from .selectors import _expression_grammar as _selector_grammar
 from .utils import deprecate, BiDict, instance_of
@@ -536,7 +538,7 @@ class Assembly(object):
     @deprecate()
     def save(
         self,
-        path: str,
+        path: PathLike,
         exportType: Optional[ExportLiterals] = None,
         mode: STEPExportModeLiterals = "default",
         tolerance: float = 0.1,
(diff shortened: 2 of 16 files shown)


adam-urbanczyk:
@HaozheZhang6 will you fix the mypy errors or do you need help?

HaozheZhang6:
Fixed. I missed `_importDXF`, widened it too, mypy is clean now. Also merged master: the merge dropped the `import os` in `exporters/__init__.py`, and `exportDXFProjection` was importing `PathLike` from `os` instead of the alias, so both use the alias now. Added it to the PathLike test.

cadquery/types.py
```diff
@@ -1,4 +1,7 @@
+import os
+
 from typing import Union, Literal
 
 UnitLiterals = Literal["MM", "CM", "M", "KM", "INCH", "FT", "MI", "UM", "NM"]
 Real = Union[int, float]
+PathLike = Union[str, "os.PathLike[str]"]
```
adam-urbanczyk, line 7: Shouldn't this be:
```suggestion
PathLike = Union[str, os.PathLike[str]]
```

adam-urbanczyk (changes requested):
Thanks, one minor thing

PR #2065, opened 2026-07-13
Fix segfault when loading an assembly with no children from xml/xbf

HaozheZhang6:
Loading an assembly that holds a single shape and no children crashes the interpreter:

```python
import cadquery as cq

assy = cq.Assembly(cq.Solid.makeBox(1, 1, 1))
assy.export("a.xml")
cq.Assembly.load("a.xml")     # Fatal Python error: Segmentation fault
```

Same for `.xbf`. Adding a single child makes it go away, which is why the existing round-trip tests do not hit it -- every fixture has children. STEP is unaffected.

Cause: such an assembly has no `TDataStd_Name` on its top-level label after the XCAF round trip, and `_importDoc` calls

```python
name_attr = TDataStd_Name()
top_level_label.FindAttribute(TDataStd_Name.GetID_s(), name_attr)
assy.name = str(name_attr.Get().ToExtString())
```

`FindAttribute` segfaults on a label that does not carry the attribute rather than returning false:

```
Fatal Python error: Segmentation fault
  File "cadquery/occ_impl/importers/assembly.py", line 378 in _importDoc
  File "cadquery/occ_impl/importers/assembly.py", line 227 in importXml
```

The module already has a helper that gets this right -- `_get_name` checks `IsAttribute` before calling `FindAttribute` -- so use it, and keep the generated name when the file does not carry one.

Test: `test_assembly_without_children_roundtrip` parametrized over step/xml/xbf. On master the xml and xbf cases take the whole pytest process down with SIGSEGV.

### DIFF
--- a/cadquery/occ_impl/importers/assembly.py
+++ b/cadquery/occ_impl/importers/assembly.py
@@ -373,14 +373,16 @@ def _importDoc(doc: TDocStd_Document, assy: AssemblyProtocol):
     if shape_tool.IsTopLevel(top_level_label) and shape_tool.IsAssembly_s(
         top_level_label
     ):
-        # Set the name of the top-level assembly to match the top-level label
-        name_attr = TDataStd_Name()
-        top_level_label.FindAttribute(TDataStd_Name.GetID_s(), name_attr)
+        # Set the name of the top-level assembly to match the top-level label. An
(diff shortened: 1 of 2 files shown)


adam-urbanczyk (approved):

jmwright (approved):
Thanks @HaozheZhang6

PR #2069, opened 2026-07-16
Better imprint for assy

adam-urbanczyk:
Low level fix enabling faster imprints for well behaved models. 

NB: this PR also contains a typing fix related to Numba version update.

______________

Tangentially related to this PR (I added explicit test for this use case) - it is possible to imprint subassemblies separately and on per subassy basis use different imprint glue settings in order to optimize performance:

```python
res1, origins1 = cq.occ_impl.assembly.imprint(complex_assy.sub1, glue='partial')
res2, origins2 = cq.occ_impl.assembly.imprint(complex_assy.sub2, glue='full')

# combine the results
res: Compound = res1 | res2
origins: dict[Shape, Shape] = origins1 | origins2
```

### DIFF
--- a/cadquery/occ_impl/assembly.py
+++ b/cadquery/occ_impl/assembly.py
@@ -37,7 +37,7 @@ from OCP.Quantity import (
 )
 from OCP.BRepAlgoAPI import BRepAlgoAPI_Fuse
 from OCP.TopTools import TopTools_ListOfShape
-from OCP.BOPAlgo import BOPAlgo_GlueEnum, BOPAlgo_MakeConnected
+from OCP.BOPAlgo import BOPAlgo_GlueEnum, BOPAlgo_Builder
 from OCP.TopoDS import TopoDS_Shape
 from OCP.gp import gp_EulerSequence
 
@@ -49,7 +49,7 @@ from vtkmodules.vtkRenderingCore import (
 )
 
 from .geom import Location
-from .shapes import Shape, Solid, Compound
+from .shapes import Shape, Solid, Compound, GlueLiteral, _set_glue, _set_builder_options
 from .exporters.vtk import toString, extractEdgesFaces
 from ..cq import Workplane
 from ..utils import BiDict
@@ -855,7 +855,9 @@ def toFusedCAF(
     return top_level_lbl, doc
 
 
-def imprint(assy: AssemblyProtocol) -> Tuple[Shape, Dict[Shape, Tuple[str, ...]]]:
+def imprint(
+    assy: AssemblyProtocol, tol: float = 0.0, glue: GlueLiteral = "full",
+) -> Tuple[Shape, Dict[Shape, Tuple[str, ...]]]:
     """
     Imprint all the solids and construct a dictionary mapping imprinted solids to names from the input assy.
     """
@@ -868,22 +870,28 @@ def imprint(assy: AssemblyProtocol) -> Tuple[Shape, Dict[Shape, Tuple[str, ...]]
             id_map[s] = name
 
     # connect topologically
-    bldr = BOPAlgo_MakeConnected()
-    bldr.SetRunParallel(True)
-    bldr.SetUseOBB(True)
+    builder = BOPAlgo_Builder()
+
+    _set_glue(builder, glue)
+    _set_builder_options(builder, tol)
 
     for obj in id_map:
-        bldr.AddArgument(obj.wrapped)
+        builder.AddArgument(obj.wrapped)
 
-    bldr.Perform()
-    res = Shape(bldr.Shape())
+    builder.Perform()
+    res = Shape(builder.Shape())
 
     # make the connected solid -> id map
+    ocp_origins = builder.Origins()
     origins: Dict[Shape, Tuple[str, ...]] = {}
 
     for s in res.Solids():
-        ids = tuple(id_map[Solid(el)] for el in bldr.GetOrigins(s.wrapped))
-        # if GetOrigins yields nothing, solid was not modified
+        if ocp_origins.IsBound(s.wrapped):
+            # if GetOrigins yields nothing, solid was not modified
+            ids = tuple(id_map[Solid(el)] for el in ocp_origins.Find(s.wrapped))
+        else:
+            ids = ()
+
         origins[s] = ids if ids else (id_map[s],)
 
     return res, origins
--- a/cadquery/occ_impl/assembly.py
+++ b/cadquery/occ_impl/assembly.py
@@ -860,6 +860,9 @@ def imprint(
 ) -> Tuple[Shape, Dict[Shape, Tuple[str, ...]]]:
     """
     Imprint all the solids and construct a dictionary mapping imprinted solids to names from the input assy.
+
+    Depending on the use case, it might be required do use different `glue` option. Moreover, for large models
+    it might be needed to limit the number of threads using :meth:`cadquery.func.setThreads`
     """
 
     # make the id map
--- a/cadquery/occ_impl/assembly.py
+++ b/cadquery/occ_impl/assembly.py
@@ -856,7 +856,7 @@ def toFusedCAF(
 
 
 def imprint(
-    assy: AssemblyProtocol, tol: float = 0.0, glue: GlueLiteral = "full",
+    assy: AssemblyProtocol, tol: float = 0.0, glue: GlueLiteral = "partial",
 ) -> Tuple[Shape, Dict[Shape, Tuple[str, ...]]]:
     """
     Imprint all the solids and construct a dictionary mapping imprinted solids to names from the input assy.
--- a/cadquery/occ_impl/assembly.py
+++ b/cadquery/occ_impl/assembly.py
@@ -890,7 +890,7 @@ def imprint(
 
     for s in res.Solids():
         if ocp_origins.IsBound(s.wrapped):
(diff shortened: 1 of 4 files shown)


cadquery/occ_impl/assembly.py
```diff
@@ -855,9 +855,14 @@ def toFusedCAF(
...
 
 
-def imprint(assy: AssemblyProtocol) -> Tuple[Shape, Dict[Shape, Tuple[str, ...]]]:
+def imprint(
+    assy: AssemblyProtocol, tol: float = 0.0, glue: GlueLiteral = "full",
+) -> Tuple[Shape, Dict[Shape, Tuple[str, ...]]]:
     """
     Imprint all the solids and construct a dictionary mapping imprinted solids to names from the input assy.
+
+    Depending on the use case, it might be required do use different `glue` option. Moreover, for large models
```
lorenzncode: Perhaps the same comment regarding glue might be added to:

https://github.com/CadQuery/cadquery/blob/b47f7b7160955ea6aff359d9ab2cd60121a30d43/cadquery/occ_impl/shapes.py#L6964

```Depending on the use case, it might be required do use different `glue` option.```

File: cadquery/occ_impl/assembly.py [hunk 2]
```diff
@@ -868,22 +873,28 @@ def imprint(assy: AssemblyProtocol) -> Tuple[Shape, Dict[Shape, Tuple[str, ...]]
...
 
     # make the connected solid -> id map
+    ocp_origins = builder.Origins()
     origins: Dict[Shape, Tuple[str, ...]] = {}
 
     for s in res.Solids():
-        ids = tuple(id_map[Solid(el)] for el in bldr.GetOrigins(s.wrapped))
-        # if GetOrigins yields nothing, solid was not modified
+        if ocp_origins.IsBound(s.wrapped):
+            # if GetOrigins yields nothing, solid was not modified
```
lorenzncode: In comment, `GetOrigins` was from the old implementation.

lorenzncode (approved):
Thanks @adam-urbanczyk !

cadquery/occ_impl/assembly.py
adam-urbanczyk: Done

adam-urbanczyk:
@jmwright would you be able to take a look?

cadquery/occ_impl/assembly.py
```diff
@@ -855,9 +855,14 @@ def toFusedCAF(
...
 
 
-def imprint(assy: AssemblyProtocol) -> Tuple[Shape, Dict[Shape, Tuple[str, ...]]]:
+def imprint(
+    assy: AssemblyProtocol, tol: float = 0.0, glue: GlueLiteral = "partial",
+) -> Tuple[Shape, Dict[Shape, Tuple[str, ...]]]:
     """
     Imprint all the solids and construct a dictionary mapping imprinted solids to names from the input assy.
+
+    Depending on the use case, it might be required do use different `glue` option. Moreover, for large models
```
jmwright:
```suggestion
    Depending on the use case, it might be required to use a different `glue` option. Moreover, for large models
```

cadquery/occ_impl/shapes.py
```diff
@@ -6956,12 +6956,15 @@ def enclose(
...
     tol: float = 0.0,
-    glue: GlueLiteral = "full",
+    glue: GlueLiteral = "partial",
     history: History | None = None,
     name: str | None = None,
 ) -> Shape:
     """
     Imprint arbitrary number of shapes.
+
+    Depending on the use case, it might be required do use different `glue` option. Moreover, for large models
```
jmwright:
```suggestion
    Depending on the use case, it might be required to use a different `glue` option. Moreover, for large models
```

jmwright (commented):
This MRE:
```python
import cadquery as cq
from cadquery.occ_impl.assembly import imprint
imprint(cq.Assembly().add(cq.Workplane().box(1, 1, 1)))
```
gives this error:
```
  File "<python-input-4>", line 1, in <module>
    imprint(cq.Assembly().add(cq.Workplane().box(1, 1, 1)))
    ~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/jwright/repos/cadquery/cadquery/occ_impl/assembly.py", line 885, in imprint
    res = Shape(builder.Shape())
  File "/home/jwright/repos/cadquery/cadquery/occ_impl/shapes.py", line 445, in __init__
    self.wrapped = downcast(obj)
                   ~~~~~~~~^^^^^
  File "/home/jwright/repos/cadquery/cadquery/occ_impl/shapes.py", line 419, in downcast
    f_downcast: Any = downcast_LUT[shapetype(obj)]
                                   ~~~~~~~~~^^^^^
  File "/home/jwright/repos/cadquery/cadquery/occ_impl/shapes.py", line 409, in shapetype
    raise ValueError("Null TopoDS_Shape object")
```

adam-urbanczyk:
> This MRE:
> 
> ```python
> import cadquery as cq
> from cadquery.occ_impl.assembly import imprint
> imprint(cq.Assembly().add(cq.Workplane().box(1, 1, 1)))
> ```
> 
> gives this error:
> 
> ```
> Traceback (most recent call last):
>   File "<python-input-4>", line 1, in <module>
>     imprint(cq.Assembly().add(cq.Workplane().box(1, 1, 1)))
>     ~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
>   File "/home/jwright/repos/cadquery/cadquery/occ_impl/assembly.py", line 885, in imprint
>     res = Shape(builder.Shape())
>   File "/home/jwright/repos/cadquery/cadquery/occ_impl/shapes.py", line 445, in __init__
>     self.wrapped = downcast(obj)
>                    ~~~~~~~~^^^^^
>   File "/home/jwright/repos/cadquery/cadquery/occ_impl/shapes.py", line 419, in downcast
>     f_downcast: Any = downcast_LUT[shapetype(obj)]
>                                    ~~~~~~~~~^^^^^
>   File "/home/jwright/repos/cadquery/cadquery/occ_impl/shapes.py", line 409, in shapetype
>     raise ValueError("Null TopoDS_Shape object")
> ```

Imprint needs at least 2 shapes to do something meaningful . The previous version does not indeed throw, I'll add an explicit check for this.

adam-urbanczyk:
Done, I also added a check for no solids at all.

jmwright (approved):
Thanks @adam-urbanczyk

shimwell:
I am now using this feature and it is super useful.

I have a use case where I have some CQ geometry and I don't know if I can get away with 'full' imprinting (which is fast but can be wrong) or if I need 'partial' imprinting (which is slower but always correct).

The CQ models are large and prepared upstream so without a lot of inspection I'm never going to know if I can go for full imprint in some surfaces.

Is there any possibility of having another mode 'auto' which know when 'full' can be used and 'partial' is necessary. For example can we use check the faces/edges and somehow conservatively identify fully coincident contacts or partial face overlap

lorenzncode:
> I am now using this feature and it is super useful.
> 
> I have a use case where I have some CQ geometry and I don't know if I can get away with 'full' imprinting (which is fast but can be wrong) or if I need 'partial' imprinting (which is slower but always correct).
> 
> The CQ models are large and prepared upstream so without a lot of inspection I'm never going to know if I can go for full imprint in some surfaces.
> 
> Is there any possibility of having another mode 'auto' which know when 'full' can be used and 'partial' is necessary. For example can we use check the faces/edges and somehow conservatively identify fully coincident contacts or partial face overlap


I'm not an OCCT expert but according to the docs isn't glue faster because it skips computations based on assumptions up front?  


https://dev.opencascade.org/doc/overview/html/specification__boolean_operations.html#specification__boolean_7
```
The performance improvement in gluing mode is achieved by excluding the most time consuming computations and in some case can go up to 90%:

Exclude computation of FACE/FACE intersections for partial coincidence;
Exclude computation of VERTEX/FACE, EDGE/FACE and FACE/FACE intersections for full coincidence.
By setting the Gluing option for the operation user should guarantee that the arguments are really coinciding. The algorithm does not check this itself. Setting inappropriate option for the operation is likely to lead to incorrect result.
```


If you run these computation in 'auto' mode then doesn't that move a time consuming part to a different function?

jon-proximafusion:
I was thinking the CAD builder could define imprint method to sure on the assembly surfaces, perhaps as a metadata.

Then the code ingesting the CAD and doing the imprint could look for this metadata per surface as an optional clue to what imprint to use.

Perhaps not worth the additional code but could speed up the imprinting at the expense of tagging the surfaces

lorenzncode:
Assembly supports user metadata, so this could be tested now with a custom imprint wrapper.

```py
from cadquery.occ_impl.assembly import imprint as imprint_assembly

def imprint_with_metadata(assy):
    glue = assy.metadata.get("imprint_glue")

    if glue not in (None, "partial", "full"):
        raise ValueError(f"Invalid glue mode: {glue!r}")

    return imprint_assembly(assy, glue=glue)
```
This works at the assembly level.  I'm not clear on how the proposal would work "per surface".

adam-urbanczyk:
This maybe could be done calling `BOPAlgo_PaveFiller` directly and/or manipulating its state somehow.

PR #2073, opened 2026-07-24
recognize .wrl as a vrml export extension

### LINKED ISSUE #740: VRML Export - File extension
Hi Everyone, 

I was trying to export a VRML file and I noticed that the file extension cadquery uses is .vrml but in other apps I've used VRML files with the filename extension as .wrl, I am wondering if this is intended, if so, Is there something I'm missing?

tylergibbs1:
fixes #740

`.wrl` is the extension most vrml tools recognize, but cadquery only inferred `VRML` from `.vrml`. this maps `.wrl` to the existing vrml exporter for both shapes and assemblies.

the exporter test now verifies that `.vrml` and `.wrl` produce the same vrml header. the assembly save and export cases also infer the format from `.wrl` instead of passing it explicitly.

validation

* ran the cadquery black fork against all four changed files
* compiled the cadquery and test python sources
* ran git diff check

full pytest was not run locally because this checkout does not have the occt runtime installed.

### DIFF
--- a/cadquery/assembly.py
+++ b/cadquery/assembly.py
@@ -598,6 +598,8 @@ class Assembly(object):
 
         if exportType is None:
             t = path.split(".")[-1].upper()
+            if t == "WRL":
+                t = "VRML"
             if t in ("STEP", "XML", "XBF", "VRML", "VTKJS", "GLTF", "GLB", "STL"):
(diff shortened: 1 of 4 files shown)


lorenzncode (approved):
LGTM

PR #2083, opened 2026-08-31
Update multimethod to >=2.1,<3.0

### LINKED ISSUE #2081: Bump multimethod to 2.1 => crash under 3.14 with dataclasses-json
There is an issue under Python 3.14 with `multimethod==1.12` and [dataclass-json](https://github.com/lidatong/dataclasses-json) as used e.g. by `pygltflib` 

```python
from collections.abc import Sequence
from dataclasses import dataclass

from dataclasses_json import dataclass_json
from multimethod import multimethod

@multimethod
def f(xs: Sequence[int]) -> int:
    return len(xs)

@dataclass_json
@dataclass
class Model:
    rows: list[int] | None = None

m = Model.from_json('{"rows": [1, 2]}')
```

- **throws `TypeError` with `multimethod==1.12` under Python 3.14**
- succeeds with `multimethod==2.1` under Python 3.14
- succeeds under Python 3.11-3.13 (this is what I tested) with `multimethod==1.12` or `multimethod==2.1`

The trigger seems to be using `multimethod` with a parameterised abstract collection annotation. 

**The good thing is that `multimethod==2.1` already fixed it.**

Hence, I tried an editable cadquery 2.9.0.dev0 installation in a `uv`env with `multimethod==2.1` under Python 3.13 and Python 3.14

Image

I also tried it 
- with Python 3.11 (pypi) **=> success**
- with an editable cadquery 2.9.0.dev0 anaconda installation for Python 3.14 with `multimethod=2.1` (using environment.yml, omitting the custom black so that it it installs) **=> success**

It seems to work well, so could you please bump `multimethod>=2.1,<3.0`.

adam-urbanczyk:


adam-urbanczyk:
Simple version bump, merging.

PR #2089, opened 2026-09-04
Fix history handling for solid(...)

adam-urbanczyk:
This fixes `solid` history for difficult/pathological cases.

### DIFF
--- a/cadquery/occ_impl/shapes.py
+++ b/cadquery/occ_impl/shapes.py
@@ -5833,7 +5833,7 @@ def _apply_reshape(op: Op, ctx: ShapeBuild_ReShape) -> Op:
         for key, val in op._images.items():
 
             mod = hist.Modified(val.wrapped)
-            if mod:
+            if not mod.IsEmpty():
                 op._images[key] = Shape.cast(mod.First())
 
     return op
--- a/cadquery/occ_impl/shapes.py
+++ b/cadquery/occ_impl/shapes.py
@@ -5844,7 +5844,7 @@ def _polish_images(op: Op, s: Shape) -> Op:
     Workaround for Reshape context not tracking (face, face) relations when fixing solids.
     """
 
-    face_dict = {f: f for f in s.faces()}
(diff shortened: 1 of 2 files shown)


cadquery/occ_impl/shapes.py
```diff
@@ -5805,6 +5818,54 @@ def _combine_ops(op: Op, *ops: Op) -> Op:
...
+
+        # update modified
+        for key, val in op._modified.items():
+
+            processed = []
+
+            for subshape in val:
+                modified = hist.Modified(subshape.wrapped)
+
+                if modified:
```
lorenzncode: It looks like an empty OCP list is `True` so this line should be changed to `if not modified.IsEmpty():`  or reverse the branches and check `if modified.IsEmpty()`.

cadquery/occ_impl/shapes.py
```diff
@@ -5805,6 +5818,54 @@ def _combine_ops(op: Op, *ops: Op) -> Op:
     return op
 
 
+def _apply_reshape(op: Op, ctx: ShapeBuild_ReShape) -> Op:
+    """
+    Apply (if applicable) additional ReShape history to an exisitn Op. Used by solid.
```
lorenzncode: typo: "exisitn" -> "existing"

cadquery/occ_impl/shapes.py
adam-urbanczyk: Hm, I thought I fixed it recently, but apparently it did not make it into the release.

cadquery/occ_impl/shapes.py
```suggestion
    Apply (if applicable) additional ReShape history to an existing Op. Used by solid.
```

cadquery/occ_impl/shapes.py
```suggestion
                if not modified.IsEmpty():
```

lorenzncode (approved):

jmwright (approved):
Thanks @adam-urbanczyk

PR #2093, opened 2026-09-08
Fix hull failure with collinear circle centres

kimstik:
`Sketch.hull()` fails on three equal circles with collinear centres:

```python
s = cq.Sketch()
for x in (0.0, 8.0, 16.0):
    s = s.arc((x, 0.0), 2.5, 0.0, 360.0)
s.hull()   # StdFail_NotDone: BRep_API: command not done
```

The tangent line between the outer two circles also touches the middle one, so it
enters the hull spanning zero degrees. `finalize_hull()` then builds a full circle
instead of a zero-length arc and the wire cannot close.

It only fails about two runs in three: `convert_and_validate()` returns
`list(set(arcs))` and `Arc` is unhashable, so the traversal order follows `id()`.
Hence the repeats in the test.

Same guard as in #1344, which has been open since June 2023 and bundles three
other changes. This is just that guard, plus a test - happy to close it if #1344
is revived instead.

### DIFF
--- a/cadquery/hull.py
+++ b/cadquery/hull.py
@@ -18,6 +18,9 @@ Points = List["Point"]
 Entity = Union["Arc", "Point"]
 Hull = List[Union["Arc", "Point", "Segment"]]
 
+# minimum arc span; below this makeCircle would return a full circle
+TOL = 1e-9
+
 
 class Point:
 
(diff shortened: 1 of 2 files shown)


adam-urbanczyk:
I suppose adding `__hash__` and  `__eq__` to `Arc` would make the test reproducible. Would you be up to it?

kimstik:
> I suppose adding `__hash__` and `__eq__` to `Arc` would make the test reproducible. Would you be up to it?

 Absolutely, I was aiming for a minimal, clean, non-intrusive patch. `__hash__` and `__eq__` are of course cleaner.

kimstik:
btw, the type check in `__eq__` isn't cosmetic: without it arc == None raises instead of returning False.
It works on master, where Arc has no `__eq__` and `object.__eq__` defers to the other operand correctly. I added the same check to `Point.__eq__` for symmetry

adam-urbanczyk (approved):
LGTM, thanks!

lorenzncode:
image

It works consistently for me now.  Thanks @kimstik !

lorenzncode (approved):

